PR #5, opened 2020-05-13
Create Sphinx documentation

victor-torres:


.gitignore
```diff
@@ -3,3 +3,8 @@
 .tox/
 dist/
 htmlcov/
+
+# Docs
+docs/_build
+docs/_static
+docs/_templates
```
kmike, line 10: _static and _templates are referred to in conf.py; I think they're supposed to be in version control, though we don't have any custom static files at the moment.

docs/intro/install.rst
```diff
@@ -0,0 +1,28 @@
...
+web-poet and its dependencies from PyPI with:
+
+::
+
+    pip install web-poet
+
+Things that are good to know
+============================
+
+web-poet is written in pure Python and depends on two key Python packages:
```
kmike, line 22: I'm not sure we should be listing dependencies in docs. Arguments against it:

* we would need to keep it in sync with setup.py
* users may start to depend on specific behavior, e.g. ResponseData being an attr.s class, while this may change
* I'm not sure how is this information actionable for a user.

Mentioning in docs that selectors are powered by parsel is good, but I'm not sure about duplicating all the dependencies here.

.gitignore
victor-torres: That's true.

- removed `_static` and `_templates` folders from `.gitignore`
- they won't be committed to the repository because they're empty dirs

docs/intro/install.rst
victor-torres: Yeah, that's was an idea I took from Scrapy's documentation when writing scrapy-poet's one. It's weird here since this project has such a small number of deps. I've just rewritten this section following your tips.

PR #8, opened 2020-07-07
SplashResponseData and AutoextractProductResponse

sibiryakov:


web_poet/page_inputs.py
```diff
@@ -15,3 +15,38 @@ class ResponseData:
...
+    not an URL of the request, if possible.
+
+    ``html`` should be content of the HTTP body, converted to unicode
+    using the detected encoding of the response, preferably according
+    to the web browser rules (respecting Content-Type header, etc.)
+    """
+    url: str
+    sourceUrl: str
+    html: str
+    jpeg: bytes
```
victor-torres, line 34: Are we always returning jpeg images? Should it be turned into `Optional[bytes]`?

web_poet/page_inputs.py
```diff
@@ -15,3 +15,38 @@ class ResponseData:
...
+@attr.s(auto_attribs=True)
+class SplashResponseData:
+    """A container for a response, rendered using Splash browser.
+
+    ``url`` should be an URL of the response (after all redirects),
+    not an URL of the request, if possible.
+
+    ``html`` should be content of the HTTP body, converted to unicode
+    using the detected encoding of the response, preferably according
+    to the web browser rules (respecting Content-Type header, etc.)
```
victor-torres, line 29: It would be awesome if you could document other arguments as well.

web_poet/page_inputs.py
```diff
@@ -15,3 +15,38 @@ class ResponseData:
...
+
+
+@attr.s(auto_attribs=True)
+class AutoextractProductResponse:
+    name: str
+    description: str
+    mainImage: str
+    images: List[str]
+    url: str
+    additionalProperty: List[Dict[str, str]]
```
victor-torres, line 45: What do you think about using snake case on these properties?

web_poet/page_inputs.py
```diff
@@ -15,3 +15,38 @@ class ResponseData:
...
+    """
+    url: str
+    sourceUrl: str
+    html: str
+    jpeg: bytes
+    httpStatus: int
+
+
+@attr.s(auto_attribs=True)
+class AutoextractProductResponse:
```
victor-torres, line 39: It would be awesome if you could mention something here in a docstring (maybe linking to AutoExtract docs).

web_poet/page_inputs.py
victor-torres: Also, I think @kmike had a simpler implementation in mind regarding AutoExtract responses. Like for example, returning the response data in a simple dictionary instead of having all the attributes...

It's interesting to see how we are providing well-structured information here. If we decide to follow this path, we just need to keep in mind that we need to keep this class up to date with the current response schemas.

kmike:
Closing it; this use case is handled in a different way now, via autoextract-poet library.

PR #13, opened 2021-05-29
base_url and urljoin shortcuts

kmike:
Hey! I've added base_url and urljoin shortcuts in this PR. urljoin implementation is based on Scrapy's.


ivanprado (approved):
@kmike thank you for adding that. It was clearly a missing piece.

ivanprado:
Thank you @kmike. Merging and preparing a release.

PR #14, opened 2021-06-02
Setting up Github actions

ivanprado:


tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py36,py37,py38,mypy,docs
+envlist = py36,py37,py38,py39,mypy,docsk
```
kmike, line 2: docsk looks like a typo

ivanprado: Ups! Thank you for the catch. Fixing it here: https://github.com/scrapinghub/web-poet/pull/15

PR #16, opened 2021-11-30
handle_urls decorator using a new PageObjectRegistry

ivanprado:
~**Do not merge:** It requires `url-matcher` to be released in advance.~

The new decorator `handle_urls` can be used to annotate overriding rules directly in the Page Object classes. The function `find_page_object_overrides` list all the declared overrides within a particular module/package recursively.

The `PageObjectRegistry` class can be used to create different sets of rules. 

A CMD has been introduced to explore packages. This is what you get if you run `python -m web_poet tests.po_lib` or simply `web_poet tests.po_lib`:

```
Use this                                                    instead of                                                           for the URL patterns            except for the patterns      with priority  meta
----------------------------------------------------------  -------------------------------------------------------------------  ------------------------------  -------------------------  ---------------  --------------------
tests.po_lib.POTopLevel1                                    tests.po_lib.POTopLevelOverriden1                                    ['example.com']                 ['/*.jpg|']                            300  {}
tests.po_lib.POTopLevel2                                    tests.po_lib.POTopLevelOverriden2                                    ['example.com']                 []                                     500  {}
tests.po_lib.a_module.POModule                              tests.po_lib.a_module.POModuleOverriden                              ['example.com']                 []                                     500  {'extra_arg': 'foo'}
tests.po_lib.nested_package.PONestedPkg                     tests.po_lib.nested_package.PONestedPkgOverriden                     ['example.com', 'example.org']  ['/*.jpg|']                            500  {}
tests.po_lib.nested_package.a_nested_module.PONestedModule  tests.po_lib.nested_package.a_nested_module.PONestedModuleOverriden  ['example.com', 'example.org']  ['/*.jpg|']                            500  {}
```

A pair or new dependencies were introduced: 
* `url-matcher` for the `Patterns` definition
* `tabulate` to format the output in the CMD

The function/classes are documented, but no general documentation was introduced. I leave it as a follow-up task. I'm not sure this documentation should be part of `web-poet` though.

Todo:
* [x] Add some documentation about `handle_urls` and `find_page_objects_overrides`, with links to `scrapy-poet` documentation where really main part of the documentation should be. Also add documentation about the `python -m web_poet` command

**Schema of possible documentation**

A section is needed for "overrides" metadata or something like that. 

1. What is `handle_urls` for. We can say it is used by some engines, e.g. `scrapy-poet`, to locate the extraction code for particular sites or sets of pages. Link to `scrapy-poet` tutorial.  
2. Explain the basic: handle_url just define the rule: for this URLs use this Page Object instead of this one. 
3. Patterns. What are they and link to `url-matcher` documentation. 
4. `find_page_objects_overrides` function. It can be used to explore the handle_urls metadata. Explanation and usage. Link to `scrapy-poet` place where shows how to configure overrides using the function. 
5. Command line tool to list PO `python -m web_poet ...`
6. Namespaces


web_poet/__main__.py
```diff
@@ -0,0 +1,34 @@
...
+import tabulate
+
+from web_poet.overrides import find_page_object_overrides
+
+
+def qualified_name(cls: Callable) -> str:
+    return f"{cls.__module__}.{cls.__name__}"
+
+
+def main(args=None):
```
BurnzZ, line 13: Very neat! 👏 Thanks for adding this!

We could also add this line to **setup.py**: `entry_points={'console_scripts': ['web_poet = web_poet.__main__:main']}`

So instead of:

```
$ python -m web_poet tests.po_lib
```

We could do something like this directly:

```
$ web_poet tests.po_lib
```

BurnzZ (commented):
Great work on this @ivanprado ! Added some comments for my initial review.

docs/requirements.txt
```diff
@@ -1,8 +1,8 @@
...
-sphinxcontrib-qthelp==1.0.3
-sphinxcontrib-serializinghtml==1.1.4
+Sphinx
+sphinx-rtd-theme
+sphinxcontrib-applehelp
+sphinxcontrib-devhelp
+sphinxcontrib-htmlhelp
+sphinxcontrib-jsmath
+sphinxcontrib-qthelp
+sphinxcontrib-serializinghtml
```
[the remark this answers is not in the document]
kmike: +1, I think for docs we should pin the dependencies, to avoid unintended breakages on readthedocs.

web_poet/overrides.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def handle_urls(include: Union[str, Iterable[str]],
+                overrides: Callable,
+                *,
+                exclude: Union[str, Iterable[str], None] = None,
+                priority: int = 500,
+                namespace: str = "",
+                **kwargs
+                ):
```
kmike, line 54: What do you think about using classes instead of string namespaces within a dictionary?
This is what I have in mind, it'd probably be easier to explain with some code:

```py
class PageObjectRegistry:
    def __init__(self):
        self.data = {}
    
    def handles_urls(self, include: ..., ...):
        # no need to do magic to get handle_urls_dict and namespace_dict,
        # use self.data instead of namespace_dict


default_registry = PageObjectRegistry()

# expose a shortcut for the default registry, so that the usage is the same:
handle_urls = default_registry.handle_urls
```

web_poet/overrides.py
```diff
@@ -0,0 +1,146 @@
...
+        return [value]
+    return list(value)
+
+
+def handle_urls(include: Union[str, Iterable[str]],
+                overrides: Callable,
+                *,
+                exclude: Union[str, Iterable[str], None] = None,
+                priority: int = 500,
+                namespace: str = "",
```
sortafreel, line 52: I would suggest adding `namespace` description to the docstring below. It looks like the only missing one.

tests/test_overrides.py
```diff
@@ -0,0 +1,69 @@
...
+    assert pones.instead_of == PONestedModuleOverridenSecondary
+
+
+def test_list_page_objects_from_empty_namespace():
+    assert find_page_object_overrides("tests.po_lib", namespace="foo") == []
+
+
+def test_cmd():
+    from  web_poet.__main__ import main
+    main(["tests.po_lib"])
```
sortafreel, line 69: There're multiple "no newline" and "multiple spaces after keyword" issues in the files that would be fixed automatically by adding `pre-commit` here (isort, flake8, mypy, black). Does it make sense or better to fix them manually?

web_poet/overrides.py
```diff
@@ -0,0 +1,146 @@
...
+
+    return wrapper
+
+
+def walk_modules(module: str) -> Iterable:
+    """
+    Return all modules from a module recursively. Note that this will import all the modules and submodules.
+    It returns the provided module as well.
+    """
+    def onerror(mod):
```
sortafreel, line 115: It mentions `mod` from the outer scope. Is there a possibility of an error here? Do we even need an argument here instead of `_`?

sortafreel (commented):
LGTM :+1:

docs/requirements.txt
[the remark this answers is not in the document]
BurnzZ: Fixed and updated the pinned dependencies

tests/test_overrides.py
BurnzZ: +1 on this. I was also thinking of adding all `pre-commit` linters in all of our repos soon, for standardization.

I'll handle this one on a succeeding PR.

CHANGELOG.rst
```diff
@@ -2,6 +2,11 @@
 Changelog
 =========
 
+TBR
+------------------
+
+* ``handle_urls`` decorator and ``find_page_object_overrides`` function added.
```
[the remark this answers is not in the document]
BurnzZ: Handled this on https://github.com/scrapinghub/web-poet/pull/16/commits/ba52ce09e983d7f1589fe84b7ab74d13a37e005b

web_poet/__main__.py
BurnzZ: Handled this on https://github.com/scrapinghub/web-poet/pull/16/commits/ba52ce09e983d7f1589fe84b7ab74d13a37e005b

tests/po_lib/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+
+class POTopLevelOverriden1:
+    ...
+
+
+class POTopLevelOverriden2:
+    ...
+
+
+# This first annotation is ignored. A single annotation per namespace per class is allowed
```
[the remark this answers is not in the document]
BurnzZ: Handled this on https://github.com/scrapinghub/web-poet/pull/16/commits/f5cffefd521e17c419ef2db004d0fbc717417aa8 where it now displays something like this:

```python
overrides.py:175: UserWarning: Multiple @handle_urls annotations with the same 'overrides' are ignored in the same Registry. Ignoring duplicate annotation on 'example.com' for <class 'tests.po_lib.POTopLevel1'>.
```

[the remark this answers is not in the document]
BurnzZ: Handled this on https://github.com/scrapinghub/web-poet/pull/16/commits/f5cffefd521e17c419ef2db004d0fbc717417aa8 where it now displays something like this:

```python
overrides.py:175: UserWarning: Multiple @handle_urls annotations with the same 'overrides' are ignored in the same Registry. Ignoring duplicate annotation on 'example.com' for <class 'tests.po_lib.POTopLevel1'>.
```

web_poet/overrides.py
BurnzZ: Good point! updated this on https://github.com/scrapinghub/web-poet/pull/16/commits/f5cffefd521e17c419ef2db004d0fbc717417aa8.

BurnzZ: Good point! updated this on https://github.com/scrapinghub/web-poet/pull/16/commits/f5cffefd521e17c419ef2db004d0fbc717417aa8.

web_poet/overrides.py
BurnzZ: The namespace approach was replaced by using Classes instead as suggested by Mikhail. Refactored in https://github.com/scrapinghub/web-poet/pull/16/commits/f5cffefd521e17c419ef2db004d0fbc717417aa8.

web_poet/overrides.py
BurnzZ: Good point! updated this on https://github.com/scrapinghub/web-poet/pull/16/commits/f5cffefd521e17c419ef2db004d0fbc717417aa8.

web_poet/overrides.py
BurnzZ: Good point! I like the idea! 💡 as having them as classes would mean they can be extended to handle even more features later on.

I refactored the namespaces into a `PageObjectRegistry` instead in https://github.com/scrapinghub/web-poet/pull/16/commits/f5cffefd521e17c419ef2db004d0fbc717417aa8.

setup.py
```diff
@@ -22,6 +22,8 @@
     install_requires=(
         'attrs',
         'parsel',
+        'url-matcher',
+        'tabulate',
```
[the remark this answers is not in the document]
BurnzZ: Tox is handling both the Python env + dependencies required on them.

I was also thinking of standardizing all of our repos (_alongside the notion of using pre-commit linters_) in such a way that local development should be convenient. 

I also find myself limited by Tox's features and would usually resort to my own virtual envs for development. Having a `requirements.txt` for that would definitely help.

In any case, we can address this in another PR alongside all of our _poet_ repos.

web_poet/overrides.py
BurnzZ: I was thinking of renaming `PageObjectRegistry` to something else, as it might confuse the users later on when they use `scrapy-poet`. However, the terminology seems apt at the moment.

web_poet/overrides.py
```diff
@@ -0,0 +1,266 @@
...
+            mod = importlib.import_module(info.name)
+            yield mod
+
+
+def find_page_object_overrides(
+    module: str, registry_name: str = "default"
+) -> List[OverrideRule]:
+    """
+    Find all the Page Objects overrides in the given module/package and its
+    submodules.
```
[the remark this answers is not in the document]
BurnzZ: @kmike It's primarily used by the upcoming revised Registry format from this `scrapy-poet`'s PR: https://github.com/scrapinghub/scrapy-poet/pull/56/files. It's main use case is to read all `@handle_urls` annotation to populate this setting:

```
from web_poet import find_page_object_overrides

SCRAPY_POET_OVERRIDES = find_page_object_overrides("my_scrapy_project.page_objects")
```

It's quite convenient since the user doesn't need to interact directly with `web-poet`'s own `PageObjectRegistry` class. Moreover, should we decide to update the implementation away from `PageObjectRegistry` class, we could prevent any backward incompatible changes using this interface. _(We could then deprecate `PageObjectRegistry` class slowly but still present for backward compatibility)_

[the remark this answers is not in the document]
kmike: I see, thanks! I wonder how it'd look like if we remove the indirection.

```
from web_poet import default_registry
walk_modules("my_scrapy_project.page_objects")
SCRAPY_POET_OVERRIDES = default_registry.get_overrides()

# we can expose it as a function as well, e.g. 
# SCRAPY_POET_OVERRIDES = web_poet.get_overrides()
```
If you do this change, it'd allow to remove:

* name property from PageObjectRegistry
* `_declare_registry_in_module` method, as well as magic REGISTRY_MODULE_ANCHOR cache
* get_data_from_module mehod
* find_page_object_overrides funcion
* probably, it'd be possible to drop HandleUrlsSpec, and just use OverrideRule, not sure

The way it could work:

1. We should make sure walk_modules is called.
2. All work is done on import time, in handle_urls decorators; all results are stored in the registry object.

For me it looks like a big simplification of the system, but I may be missing some features or use cases.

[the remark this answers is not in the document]
BurnzZ: I see your point. 👍 Yeah, I think it make sense to remove `find_page_object_overrides()` if we'd directly expose the registry here. It would seem that the new `PageObjectRegistry` completely removes the need for such function since we're not dealing with the _magic_ module attributes anymore from the previous refactor.

I'd propose to have the following interfaces as I think a separate call to `walk_modules()` outside of the registry would introduce some complexity.

```python
from web_poet import default_registry

# retrieves all overrides inside the current Scrapy project
SCRAPY_POET_OVERRIDES = default_registry.get_overrides()

# or, could retrieve them from specific modules 
SCRAPY_POET_OVERRIDES = default_registry.get_overrides_from_module("common_page_objects.ecommerce.lalazon")
```

I'll try to do a soft refactor and see what the new code would look like. But I really do think this would simplify it a lot. Thanks for bringing it up!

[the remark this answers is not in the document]
kmike: Walking the current Scrapy project makes sene.

I'm not sure about `default_registry.get_overrides_from_module("common_page_objects.ecommerce.lalazon")`.
walk_modules makes sure that all modules in a package are imported, and so decorators are executed.
But there can be modules outside, which are also imported. So, if get_overrides_from_module is implemented as "walk_modules; return self.overrides", it may return overrides which are not in `"common_page_objects.ecommerce.lalazon"`.

Then, the next question: do we actually want to have such feature, "return only POs from this package"? For me it looks we're fine without this feature; users can use separate registries if they want.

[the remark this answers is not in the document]
BurnzZ: > Then, the next question: do we actually want to have such feature, "return only POs from this package"? For me it looks we're fine without this feature; users can use separate registries if they want.

During the middle of the refactoring process on this, I realized that collecting the overrides rules using a module string works really well when users organize and sort their Page Objects into directories. We've tackled such PO organization before with something like:

```
my_scrapy_project
|
└── page_objects
    ├── lalazon
    |   ├── base
    |   |   ├── __init__.py
    |   |   ├── lalazon_products.py
    |   |   └── lalazon_product_listings.py
    |   ├── us
    |   |   ├── __init__.py
    |   |   ├── lalazon_us_products.py
    |   |   └── lalazon_us_product_listings.py
    |   └── fr
    |       ├── __init__.py
    |       ├── lalazon_fr_products.py
    |       └── lalazon_fr_product_listings.py
    └── rawrmart
        ├── __init__.py
        ├── rawrmart_products.py
        └── rawrmart_product_listings.py
```

Given the Scrapy project structure above, users could opt with just a single registry which is `default_registry`. They would then have the capability to filter the rules per module.

```python
from web_poet import default_registry

SCRAPY_POET_OVERRIDES = default_registry.get_overrides_from_module("my_scrapy_project.page_objects.lalazon")
# or 
SCRAPY_POET_OVERRIDES = default_registry.get_overrides_from_module("my_scrapy_project.page_objects.rawrmart")
```

Here's a WIP commit for the ongoing refactor in case your interested on how it turned out: https://github.com/scrapinghub/web-poet/pull/16/commits/1efb5068ed1642bc04c8dd0ba814377a910eb28d. I've yet to update the docs, implement some methods, and add some more tests.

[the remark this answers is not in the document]
BurnzZ: Updated the docs for https://github.com/scrapinghub/scrapy-poet/pull/56 in-line with this recent refactoring: https://github.com/scrapinghub/scrapy-poet/pull/56/commits/c865c60f1cea0fba17186fd230391be0b0bb1577

web_poet/overrides.py
BurnzZ: Handled this on https://github.com/scrapinghub/web-poet/pull/16/commits/f5cffefd521e17c419ef2db004d0fbc717417aa8

web_poet/overrides.py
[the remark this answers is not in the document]
BurnzZ: @kmike Our last offline conversation ended on the thought of simplifying how the Override Rules are declared and organized by choosing either the **package-based** or **Multiple Registry** approach. However, after giving some time to think about it, I still believe it's worth keeping both approaches present to give adequate amount of options to developers.

I wrote down a dedicated section on the topic in our docs using this commit: https://github.com/scrapinghub/web-poet/pull/16/commits/75593edb29378299093dc80ae97c9e3fa483ee13. I believe this could clarify the advantages of keeping both approaches available by exploring different scenarios in using them.

docs/intro/overrides.rst
```diff
@@ -0,0 +1,387 @@
...
+    class DualExampleProductPage(ItemWebPage):
+        def to_item(self):
+            ...  # more specific parsing
+
+The code above declares that:
+
+    - For sites that matches the ``example.com`` pattern, ``ExampleProductPage``
+      would be used instead of ``GenericProductPage``.
+    - The same is true for ``YetAnotherExampleProductPage`` where it is used
+      instead of ``GenericProductPage`` for two URLs: ``dualexample.com`` and
```
kmike, line 83:
```suggestion
      instead of ``GenericProductPage`` for two websites: ``dualexample.com`` and
```

docs/intro/overrides.rst
```diff
@@ -0,0 +1,387 @@
...
+A convenience function is available discover and retrieve all rules from your
+project. Make sure to check out :ref:`Overrides API section <api-overrides>`
+to see the other functionalities.
+
+.. code-block:: python
+
+    from web_poet import default_registry
+
+    # Retrieves all rules that were registered in the registry
+    rules = default_registry.get_overrides()
```
kmike, line 140: If I'm not mistaken, this example has a big caveat: all modules which use handle_urls decorator must be imported already, before calling `default_registry.get_overrides()`.

docs/intro/overrides.rst
```diff
@@ -0,0 +1,387 @@
...
+
+    # Or, we could also filter out the rules by the module they were defined in
+    rules = default_registry.get_overrides_from_module("my_project.page_objects")
```
kmike, line 143: This example doesn't have the same caveat as an example above, so it's actually quite different; it is not only about filtering, but also about making it actually work.

docs/intro/overrides.rst
kmike: I.e. if you don't call osme kind of walk_modules, or have some framework doing it at a right time, there is a risk ruless are going to be missing.

docs/intro/overrides.rst
```diff
@@ -0,0 +1,387 @@
...
+Page Object would be used. The URL matching rules is handled by another library
+called `url-matcher <https://url-matcher.readthedocs.io>`_.
+
+Using such matching rules establishes the core concept of Overrides wherein
+its able to use specific Page Objects in lieu of the original one.
+
+This enables ``web-poet`` to be used effectively by other frameworks like 
+`scrapy-poet <https://scrapy-poet.readthedocs.io>`_.
+
+Example Use Case
```
kmike, line 16: 👍 that's a very well written docs section!

docs/intro/overrides.rst
```diff
@@ -0,0 +1,387 @@
...
+
+Then we could easily retrieve all Page Objects per subpackage or module like this:
+
+.. code-block:: python
+
+    from web_poet import default_registry
+
+    # We can do it per website.
+    rules = default_registry.get_overrides_from_module("my_page_obj_project.cool_gadget_site")
+    rules = default_registry.get_overrides_from_module("my_page_obj_project.furniture_site")
```
kmike, line 225: This example suggests that the function should be named `default_registry.get_overrides_from_package`

docs/intro/overrides.rst
```diff
@@ -0,0 +1,387 @@
...
+    # It can also drill down to the country domains on a given site.
+    rules = default_registry.get_overrides_from_module("my_page_obj_project.cool_gadget_site.us")
+    rules = default_registry.get_overrides_from_module("my_page_obj_project.cool_gadget_site.fr")
+
+    # or even drill down further to the specific module.
+    rules = default_registry.get_overrides_from_module("my_page_obj_project.cool_gadget_site.us.products")
+    rules = default_registry.get_overrides_from_module("my_page_obj_project.cool_gadget_site.us.product_listings")
+
+    # Or simply all of Override rules ever declared.
+    rules = default_registry.get_overrides()
```
kmike, line 236: I think for this to work user would need to call walk_modules

docs/intro/overrides.rst
```diff
@@ -0,0 +1,387 @@
...
+
+.. code-block:: python
+
+    # my_page_obj_project/cool_gadget_site_us_products.py
+
+    from . import cool_gadget_registry, cool_gadget_us_registry
+    from web_poet.pages import ItemWebPage
+
+    @cool_gadget_registry.handle_urls("coolgadgetsite.com", overrides=GenericProductPage)
+    @cool_gadget_us_registry.handle_urls("coolgadgetsite.com", overrides=GenericProductPage)
```
kmike, line 283: I was having a different idea in mind when talking about multiple registries, and probably a different use case.

Let's say you have all of the following:

1. A large package, with many POs for different websites.
2. A few small packages, with POs for some specific websites.

These packages may be in separate git repos, packaged separately, etc.

There are a few use cases:

1. We want to use all of the POs defined in all these packages. This probably should be handled by web_poet.get_overrides().
2. We want to use all of the POs defined in all packages, but replace or exclude some of them (you created a better version and want to use it instead).
3. We want to use POs only from some of the packages.
4. We want to use some POs from some of the packages.

In all these cases modifying the source code of the packages may not be an option; we'd really want to select which POs to enable and which POs not to use in a flexible way.

It seems this is possible with the `get_overrides_from_module` API, but all the examples of multiple registries require making changes to the source code. I'm not sure it'd be a good practice to suggest people to create many registries and modify the source code to add more handle_urls decorators (usually with the same parameters - a lot of copy-paste).

I was thinking about a different approach for using multiple registries, something like this:

```py

registry = web_poet.PageObjectRegistry()

from some_package.pages import CoolGadgetSiteProductPage
registry.handle_urls(CoolGadgetSiteProductPage, "coolgadgetsite.com", overrides=GenericProductPage)
```

Of course, you can also have POs in your own project using the same registry, declared with a decorator.

What I don't like in the example above is that all handle_urls parameters are still duplicated

web_poet/overrides.py
```diff
@@ -0,0 +1,207 @@
...
+            # Dict ensures that no duplicates are collected and returned.
+            rules.update(self._filter_from_module(mod.__name__))
+
+        return list(rules.values())
+
+    def _filter_from_module(self, module: str) -> Dict[Callable, OverrideRule]:
+        return {
+            cls: rule
+            for cls, rule in self.data.items()
+            if cls.__module__.startswith(module)
```
kmike, line 177: Can string matching lead to issues here? E.g. cls.__module__ is foo.bar, and module is "fo"?

I also wonder if startswith is necessary; doesn't walk_modules import all modules recursively, so that you can always use exact match?

Caveat: I'm not confident I understood this part correctly :)

docs/intro/overrides.rst
BurnzZ: @kmike I think this could actually be _**URL patterns**_ to improve clarity. I've also updated the examples so that it clearly reflects that it's not simply a **URL** or a **website** link but rather a pattern which follows the syntax of [url-matcher](https://url-matcher.readthedocs.io/en/stable/). We can take this opportunity to showcase some of the power of `url-matcher`.

Updated this on: https://github.com/scrapinghub/web-poet/pull/16/commits/7035ab3ba9b89921e51cb401140737cd0cd7bec7

docs/intro/overrides.rst
BurnzZ: I was thinking about that too.

I believe this was originally named like this to follow `importlib`'s naming convention where it uses the **module** terminology which includes **packages** and **subpackages**. Renaming this into `get_overrides_from_package` also present some slight ambiguity since it also supports **modules**.

I think we can rename it to simply `get_overrides_from` instead. This new name also opens up other possibilities of supporting paths like `my_proj/path/something` if we detect a `/` char. I'm not sure if users would like to have this but it's now an option in the future.

Update this on: https://github.com/scrapinghub/web-poet/pull/16/commits/c000cbc9d36f1c3022c22d27d67d83e4d86b41c7 Let me know if you have some concerns on this.

web_poet/overrides.py
BurnzZ: > Can string matching lead to issues here? E.g. cls.__module__ is foo.bar, and module is "fo"?

Great catch! 🙏  It would indeed have some matching issues here. Fixed it on https://github.com/scrapinghub/web-poet/pull/16/commits/c000cbc9d36f1c3022c22d27d67d83e4d86b41c7.

> I also wonder if startswith is necessary; doesn't walk_modules import all modules recursively, so that you can always use exact match?

What we're filtering out here is the Registry's `data` values which contains the mapping of override rules. There might be cases where the registry was applied with `walk_modules` and could now contain `data` values on some other package/module. This filtration would prevent that.

docs/intro/overrides.rst
BurnzZ: Gotcha. I see what you mean now. I've created a new `consume_modules()` function that can be _optionally_ called before `default_registry.get_overrides()`, should the need arises. Also updated the docs to properly reflect this caveat.

Updated this on: https://github.com/scrapinghub/web-poet/pull/16/commits/daa3ff9bdfa754bacfd8a20862f68acb12a3fa61

docs/intro/overrides.rst
BurnzZ: Updated this on: https://github.com/scrapinghub/web-poet/pull/16/commits/daa3ff9bdfa754bacfd8a20862f68acb12a3fa61

docs/intro/overrides.rst
BurnzZ: Gotcha. 👍 

BTW, I updated the newly renamed `get_overrides_from()` _(formerly called `get_overrides_from_module()`)_ to accept an arbitrary number of strings for convenience. Updated this on https://github.com/scrapinghub/web-poet/pull/16/commits/3b05c075d1c49058139fb36ec7ec9982f3ab41a2

docs/intro/overrides.rst
BurnzZ: Got it. I do see your point. Cheers for laying it out! I'll be updating this section in the docs to better showcase such scenario.

BurnzZ: Thanks for raising this @kmike ! I created a new dedicated section for this to explore various ways of using external packages which contain Page Objects. I can see how this would be a common requirement/usecase for the users.

> It seems this is possible with the get_overrides_from_module API, but all the examples of multiple registries require making changes to the source code. I'm not sure it'd be a good practice to suggest people to create many registries and modify the source code to add more handle_urls decorators (usually with the same parameters - a lot of copy-paste).

This section was not taking into account the use of external packages but rather exploring how to organize the Override Rules using multiple Registries in a single local project. I believe the new dedicated section named **Using Overrides from External Packages** might remove the confusion in the docs: https://github.com/scrapinghub/web-poet/pull/16/commits/bd3a88eacc0cdf69f7e94d1d428a88872a9baf50

web_poet/overrides.py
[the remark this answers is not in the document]
BurnzZ: Resolving this thread as we've branched out into newer discussions in separate threads.

docs/intro/overrides.rst
BurnzZ: I've improved upon the previous commit linked above by creating a much more simpler interface for both `get_overrides()` + _**consuming/importing external packages/modules**_. Changes are in https://github.com/scrapinghub/web-poet/pull/16/commits/94fed0f1f8b1e2c8b62ec24d0b193b494ee8446f.

BurnzZ:
Hi @kmike , 

After working and polishing on the separate PR in https://github.com/scrapinghub/scrapy-poet/pull/57, I realized we needed a quick and easy way to access the rules via CLI. Thus I've introduced a concept of a **registry_pool** where it contains all instances of `PageObjectRegistry` that were ever created. This allows us to easily pin point any of the registry instances to be able to access and modify them.

Commit is at https://github.com/scrapinghub/web-poet/pull/16/commits/de5563ab967010ae03bab93b3cc276c9ab5f8ee4

web_poet/overrides.py
```diff
@@ -0,0 +1,383 @@
...
+        # TODO: Remove this when the following has been implemented:
+        #   - https://github.com/zytedata/url-matcher/issues/3
+        pattern_hash = hash(
+            (
+                tuple(self.for_patterns.include),
+                tuple(self.for_patterns.exclude),
+                self.for_patterns.priority,
+            )
+        )
+        return hash((pattern_hash, self.use, self.instead_of))
```
BurnzZ, line 52: @kmike following our conversation in https://github.com/zytedata/url-matcher/pull/4, `OverrideRule` should be set to **frozen=True** as well.

Although this changes the use case I've written down on the docs for scenarios wherein:

- users are importing an external library of Page Objects
- manipulates the existing Registry rules from it _(e.g. improving rules, replacing POs, etc)_

I'll update the guide in the doc for this one. However, making `OverrideRule` frozen would need a new API to make it more easy to manipulate the rules.

I'll push a commit on this one soon.

BurnzZ:
Hi @kmike , I've updated the Registry API so that users can now easily **combine, modify, delete, search**, etc different rules from different registries. Done in https://github.com/scrapinghub/web-poet/pull/16/commits/bf0b3e5e04b3bbb2a8e25f9118d271727fc4f3a4.

web_poet/overrides.py
BurnzZ: Fixed this in https://github.com/scrapinghub/web-poet/pull/16/commits/bf0b3e5e04b3bbb2a8e25f9118d271727fc4f3a4.

docs/intro/overrides.rst
BurnzZ: Resolving this thread as it's outdated. The latest commits has a much simplified API and now simply calls it as `get_overrides()`. A param named `filter` is now available to easily filter out packages/modules. Reference: https://github.com/scrapinghub/web-poet/pull/16/files#diff-409cb5fe79e7d7d5ede41109573e31bd847a147a966f477de80c99279bcdf479R249-R251

docs/intro/overrides.rst
BurnzZ: Resolving this thread since the latest commits have combined the functionality of `get_overrides_from_module()` into one simple API named `get_overrides()` which has a **consume** and **filter** params.

More info in https://github.com/scrapinghub/web-poet/pull/16/files#diff-409cb5fe79e7d7d5ede41109573e31bd847a147a966f477de80c99279bcdf479R247-R251

docs/intro/overrides.rst
```diff
@@ -0,0 +1,678 @@
...
+
+Overrides contains mapping rules to associate which URLs a particular
+Page Object would be used. The URL matching rules is handled by another library
+called `url-matcher <https://url-matcher.readthedocs.io>`_.
+
+Using such matching rules establishes the core concept of Overrides wherein
+its able to use specific Page Objects in lieu of the original one.
+
+This enables ``web-poet`` to be used effectively by other frameworks like 
+`scrapy-poet <https://scrapy-poet.readthedocs.io>`_.
```
kmike, line 14: todo: think about how can we simplify the description here. There is nothing wrong with the current description, but it was a bit hard for me to understand what it means. It becomes much more clear in the next section.

docs/intro/overrides.rst
```diff
@@ -0,0 +1,678 @@
...
+
+Overrides
+=========
+
+Overrides contains mapping rules to associate which URLs a particular
+Page Object would be used. The URL matching rules is handled by another library
+called `url-matcher <https://url-matcher.readthedocs.io>`_.
+
+Using such matching rules establishes the core concept of Overrides wherein
+its able to use specific Page Objects in lieu of the original one.
```
kmike, line 11:
```suggestion
it's able to use specific Page Objects in lieu of the original one.
```

kmike:
Closing it in favor of https://github.com/scrapinghub/web-poet/pull/27.

PR #22, opened 2022-02-08
implementation of additional requests

BurnzZ:


web_poet/requests.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+@attr.define
+class Request:
+    """Represents a generic HTTP request."""
+
+    url: str
+    method: str = "GET"
+    headers: Optional[mapping] = None
+    body: Optional[str] = None
```
kmike, line 76: I think it should be bytes, not str

web_poet/requests.py
```diff
@@ -0,0 +1,120 @@
...
+from typing import Optional, List, Dict, ByteString, Any, Union
+
+import attr
+
+from web_poet.page_inputs import ResponseData
+
+logger = logging.getLogger(__name__)
+
+
+mapping = Dict[Union[str, ByteString], Any]
```
kmike, line 58: Why is header value Any? Shouldn't it be str or bytes?

BurnzZ: Good catch! It was supposed to be `AnyStr`.

Let's discuss this further in the separate PR in https://github.com/scrapinghub/web-poet/pull/24.

web_poet/requests.py
BurnzZ: This could be split into `raw_data: bytes` and `html: str` into a separate `HttpResponseBody` as per our recent discussions.

I've opened up a separate PR in https://github.com/scrapinghub/web-poet/pull/24 to open up the deep discussion on this.

docs/advanced/additional_requests.rst
```diff
@@ -0,0 +1,270 @@
...
+    the **Crawling Logic** at all. They are simply a low-level way to interact
+    with today's websites which relies on a lot of page interactions to display
+    its contents.
+
+
+HttpClient
+==========
+
+The main interface for executing additional requests would be :class:`~.HttpClient`.
+It also has full support for :mod:`asyncio` enabling developers to perform
```
kmike, line 30: Does it (it - page object's to_item method?) support asyncio or coroutines? For example, if it supports asyncio, I'd expect asyncio libraries to work, e.g. asyncio.gather, or asyncio.wait.

docs/advanced/meta.rst
```diff
@@ -0,0 +1,130 @@
...
+    import attr
+    import web_poet
+
+    @attr.define
+    class SomePage(web_poet.ItemWebPage):
+        # The HttpResponse attribute is inherited from ItemWebPage
+        meta: web_poet.Meta
+
+    response = web_poet.HttpResponse(...)
+    meta = web_poet.Meta("arbitrary_value": 1234, "cool": True)
```
kmike, line 31: it is a correct syntax?

web_poet/requests.py
kmike: We addressed it for Response, not we need to fix it for the Request in some way (which could actually be different).

web_poet/requests.py
```diff
@@ -0,0 +1,142 @@
...
+    download the actual requests. However, it can easily be overridable by
+    providing an optional ``request_downloader`` callable.
+
+    Providing the request implementation by dependency injection would be a good
+    alternative solution when you want to avoid setting up :mod:`contextvars`
+    like ``web_poet.request_backend_var``.
+
+    In any case, this doesn't contain any implementation about how to execute
+    any requests fed into it. When setting that up, make sure that the downloader
+    implementation returns a :class:`~.HttpResponse` instance.
```
kmike, line 94: I just realized there is one big piece which is missing: we need to standardize the exception and error handling as well. Users should be able to write error handling code in POs, and it should work across backends. 

We should also document and test error handling.

```py
try:
    response = await self.http_client.get("...")
except ????:
    # error handling
```

So, the questions:

1. What are the exceptions raised?
2. When are they raised?

For example, there are connection errors, tls validation errors, read timeout errors, etc. There are also non-200 status codes. When are we raising an error, and when we're not?

web_poet/requests.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def __init__(self, request_downloader: Callable = None):
+        self.request_downloader = request_downloader or _perform_request
+
+    async def request(
+        self,
+        url: str,
+        method: str = "GET",
+        headers: Optional[mapping] = None,
+        body: Optional[str] = None,
```
kmike, line 105: body shouldn't be str, it should be bytes or a special object

web_poet/requests.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+
+    def __init__(self, request_downloader: Callable = None):
+        self.request_downloader = request_downloader or _perform_request
+
+    async def request(
+        self,
+        url: str,
+        method: str = "GET",
+        headers: Optional[mapping] = None,
```
kmike, line 104: we might need to define HttpRequestHeaders, similar to HttpResponseHeaders.

kmike (changes requested):
hey @BurnzZ! I think we need to solve these 2 problems before merging the PR:

1. Implement, test and document error handling 
2. Finalize the Request class, and request.body as str in particular.

docs/advanced/additional_requests.rst
```diff
@@ -0,0 +1,278 @@
...
+                "url": self.url,
+                "name": self.css("#main h3.name ::text").get(),
+                "product_id": self.css("#product ::attr(product-id)").get(),
+            }
+
+            # Simulates clicking on a button that says "View All Images"
+            response: web_poet.ResponseData = await self.http_client.get(
+                f"https://api.example.com/v2/images?id={item['product_id']}"
+            )
+            page = web_poet.WebPage(response)
```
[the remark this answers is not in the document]
BurnzZ: Removed the need to cast pages into `WebPage` in https://github.com/scrapinghub/web-poet/pull/22/commits/f769147c165fd3f4821f2119ee18877498e070a7 following the new `HttpResponse` in https://github.com/scrapinghub/web-poet/pull/31 which provides **selectors** right off the bat.

docs/index.rst
```diff
@@ -34,6 +34,12 @@ and the motivation behind ``web-poet``, start with :ref:`from-ground-up`.
    intro/tutorial
    intro/from-ground-up
 
+.. toctree::
+   :caption: Advanced
+   :maxdepth: 1
+
+   advanced/additional_requests
```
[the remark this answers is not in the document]
BurnzZ: Good catch! Updated this in https://github.com/scrapinghub/web-poet/pull/22/commits/83839cf2d67d515f628f2e904fa55604bd4ee9c5

web_poet/requests.py
BurnzZ: Addressed this in https://github.com/scrapinghub/web-poet/pull/22/commits/a103181060048bffa49a19108c1d881c4d9131b0.

web_poet/requests.py
BurnzZ: Addressed this in https://github.com/scrapinghub/web-poet/pull/22/commits/a103181060048bffa49a19108c1d881c4d9131b0.

web_poet/requests.py
BurnzZ: Addressed this in https://github.com/scrapinghub/web-poet/pull/22/commits/a103181060048bffa49a19108c1d881c4d9131b0.

docs/advanced/additional_requests.rst
```diff
@@ -0,0 +1,278 @@
...
+
+        async def to_item(self):
+            item = {
+                "url": self.url,
+                "name": self.css("#main h3.name ::text").get(),
+                "product_id": self.css("#product ::attr(product-id)").get(),
+                "related_product_ids": self.parse_related_product_ids(self),
+            }
+
+            requests: List[web_poet.Request] = [
```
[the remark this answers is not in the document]
BurnzZ: Thanks for reminding me! Added a dedicated section for `HttpRequest` in https://github.com/scrapinghub/web-poet/pull/22/commits/a853d164b8acde81ea135a44ef7c082920e05009.

docs/advanced/additional_requests.rst
```diff
@@ -0,0 +1,278 @@
...
+        @staticmethod
+        def parse_related_product_ids(page: web_poet.WebPage) -> List[str]:
+            return page.css("#main .related-products ::attr(product-id)").getall()
+
+The key takeaways for this example are:
+
+    * A :class:`~.Request` can be instantiated to represent a Generic HTTP Request.
+      It only contains the HTTP Request information for now and isn't executed yet.
+      This is useful for creating factory methods to help create them without any
+      download execution at all.
```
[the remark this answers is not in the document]
BurnzZ: Inside the `HttpClient`, yes.

The rest of the methods like `request`, `get`, and `post` builds the `HttpRequest` themselves. This changes shifts the focus from creating `HttpRequest`s into actually executing the requests.

docs/advanced/additional_requests.rst
BurnzZ: Yes. Added more explicit notes on this in https://github.com/scrapinghub/web-poet/pull/22/commits/f0a3419325508c4e8e4b5bfec87a1d19aaacc660#diff-1b9a5e2fee681a2590fc89977cef10683e746d3fe4b2d207c262a75e17ec9657R119-R122

docs/advanced/meta.rst
BurnzZ: Thanks for catching! I was missing the `{` and `}` characters here. Updated this on https://github.com/scrapinghub/web-poet/pull/22/commits/f0a3419325508c4e8e4b5bfec87a1d19aaacc660#diff-f010592a6c5d89e2fe40fc8b7a02d88d657351a09fc8f693e91a2cf6bf445a83R30-R34

docs/advanced/additional_requests.rst
```diff
@@ -0,0 +1,278 @@
...
+an :class:`~.HttpClient` instance:
+
+
+.. code-block:: python
+
+    def request_implementation(r: web_poet.Request) -> web_poet.ResponseData:
+        ...
+
+    from web_poet import HttpClient
+    http_client = HttpClient(request_downloader=request_implementation)
```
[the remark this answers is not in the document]
BurnzZ: Improved the docs in https://github.com/scrapinghub/web-poet/pull/22/commits/e74ca5038ca0d5f25ba077fb3ece5fc456e9d2e2.

Kindly let me know if this is in line with the boilerplate you're looking for.

docs/advanced/additional-requests.rst
```diff
@@ -0,0 +1,432 @@
...
+
+    import json
+    import web_poet
+
+    request = web_poet.HttpRequest(
+        url="https://www.api.example.com/product-pagination/",
+        method="POST",
+        headers={
+            "Host": "www.example.com",
+            "Content-Type": "application/json; charset=UTF-8",
```
BurnzZ, line 43: perhaps something like a `json` keyword in `post()` could set this up automatically similar to the requests lib.

web_poet/requests.py
```diff
@@ -0,0 +1,204 @@
...
+    **web-poet** like :class:`~.HttpClient`.
+    """
+
+    url: str = attrs.field()
+    method: str = attrs.field(default="GET")
+    headers: HttpRequestHeaders = attrs.field(
+        factory=HttpRequestHeaders, converter=HttpRequestHeaders
+    )
+    body: HttpRequestBody = attrs.field(
+        factory=HttpRequestBody, converter=HttpRequestBody.from_anystr  # type: ignore
```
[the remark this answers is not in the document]
BurnzZ: @kmike , as discussed offline, a new PR will be branched off containing this proposal to be discussed further: https://github.com/scrapinghub/web-poet/pull/36

web_poet/requests.py
BurnzZ: @kmike , as discussed offline, I've added the base exception that frameworks implementing **web-poet** should support. All other exceptions that we'll be adding and supporting in the future should subclass from it.

Let me know what you think! Here's the commit: https://github.com/scrapinghub/web-poet/pull/22/commits/eda3352139ea4ca62ad87835f73ff445185f33d3

web_poet/requests.py
```diff
@@ -0,0 +1,184 @@
...
+request_backend_var: ContextVar = ContextVar("request_backend")
+
+
+class HttpRequestBody(bytes):
+    """A container for holding the raw HTTP request body in bytes format."""
+
+    pass
+
+
+class HttpRequestHeaders(CIMultiDict):
```
kmike, line 38: Why is the implementation different from HttpResponseHeaders? I haven't checked it, but wouldn't it make sense to have HttpHeaders class, and then have request/response header classes inherit from it?

web_poet/__init__.py
```diff
@@ -1,2 +1,14 @@
 from .pages import WebPage, ItemPage, ItemWebPage, Injectable
-from .page_inputs import HttpResponse, HttpResponseBody, HttpResponseHeaders
+from .requests import (
+    request_backend_var,
+    HttpRequest,
+    HttpRequestHeaders,
```
kmike, line 5: I wonder if these objects can be page inputs as well. E.g. can a page object  for the request headers?

web_poet/requests.py
```diff
@@ -0,0 +1,184 @@
...
+
+        A :class:`~.HttpResponse` instance should then be returned.
+
+        .. warning::
+            By convention, the request implementation supplied optionally to
+            :class:`~.HttpClient` should return a :class:`~.HttpResponse` instance.
+            However, the underlying implementation supplied might change that,
+            depending on how the framework using **web-poet** implements it.
+        """
+        req = HttpRequest(url, method, headers or {}, body or b"")
```
kmike, line 156: let's use keyword arguments here

web_poet/requests.py
BurnzZ: Good point, I was thinking of the same thing. I've addressed this in https://github.com/scrapinghub/web-poet/pull/22/commits/65badca5d6746e0d96ee09045b33e0bbf4556c87.

web_poet/page_inputs.py
```diff
@@ -1,37 +1,35 @@
...
+class HttpResponseBody(_HttpBody):
     """A container for holding the raw HTTP response body in bytes format."""
 
     def declared_encoding(self) -> Optional[str]:
         """ Return the encoding specified in meta tags in the html body,
         or ``None`` if no suitable encoding was found """
         return html_body_declared_encoding(self)
 
-    def json(self):
+    def json(self) -> Dict[Any, Any]:
```
kmike, line 25: I think the result could be any JSON object (list, str, number), not necessarily a dict.

web_poet/_base.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+from typing import Type, TypeVar, List, Dict
+
+from multidict import CIMultiDict
+
+T_headers = TypeVar("T_headers", bound="_HttpHeaders")
+
+
+class _HttpBody(bytes):
```
kmike, line 14: Not sure about this :) What's the use case? Unlike headers, request body seems to be quite different from response body.

web_poet/__init__.py
BurnzZ: Yeah, that makes sense. I imagine there would be some cases wherein developers need both the **request-response** pairs inside the Page Object.

We'll need to add a definition for this in `pages.py`. For now, however, I simply moved them into `page_inputs.py` in https://github.com/scrapinghub/web-poet/pull/22/commits/99e9b41cce3494c159d7f33df27897eb61fff38b. This simplifies the tests as well since they have a lot in common.

web_poet/requests.py
BurnzZ: Addressed this in https://github.com/scrapinghub/web-poet/pull/22/commits/6a53b90eb8cfa34d39bc7556d87f103480fd40c4.

Moreover, to clarify, did you mean to also enforce keyword args in the classes via `@attr.define(..., kw_only=True)`?

web_poet/page_inputs.py
BurnzZ: You're right. Thanks for catching this! Reverted the mistake in https://github.com/scrapinghub/web-poet/pull/22/commits/c1f7675783cdb5ba970619f352f14417fe179655

web_poet/_base.py
BurnzZ: I was initially planning to house alternative constructors like in https://github.com/scrapinghub/web-poet/pull/36 inside it. But mulling over it now, it's not a big of a motivation to introduce `_HttpBody` that what I initially envisioned.

Thanks for raising this! Reverted in https://github.com/scrapinghub/web-poet/pull/22/commits/3310c1b5c04281765407d190f85c8cfe638ddeab.

web_poet/requests.py
kmike: I was thinking only about changing this example. As for making all attributes kw-only, url would be nice to keep positional.

BurnzZ: Got it. Nonetheless, I think it's worth changing the other non-URL init params to be kw-only to enforce correctness in case the args have unintentionally swapped positions. I've updated this in https://github.com/scrapinghub/web-poet/pull/22/commits/d88a0e5ceeb195d9adb8dbafb89ee338ad06b545

docs/advanced/additional-requests.rst
```diff
@@ -0,0 +1,503 @@
...
+
+    page = SomePage(response=response, http_client=create_http_client())
+    item = page.to_item()
+
+From the code sample above, we can see that every time an :class:`~.HttpClient`
+is created for Page Objects needing an ``http_client``, the specific **request
+implementation** from a given framework is injected to it.
+
+
+Exception Handling
```
kmike, line 432: I think we should describe error handling before describing backends - just give some examples, of how to handle errors for 

1. individual requests (e.g. get or post)
2. batch requests: what're the semantics, is it possible to get partial data, etc.

There is no need to move the whole section above. I think our docs could have 2 main parts, aimed to a different type of users:

1. People who write Page Objects, and need additional requests
2. People who're implementing web-poet intergations. Information about backends, etc. is mostly for this group.

I think it'd be good to make it clear what to do for (1) first, and then provide information for (2).

docs/advanced/additional-requests.rst
```diff
@@ -0,0 +1,503 @@
...
+==================
+
+All exceptions that the HTTP Downloader Implementation explicitly raises when
+implementing it for **web-poet** should be :class:`web_poet.exceptions.http.HttpRequestError`
+or a subclass from it.
+
+For frameworks that implement and use **web-poet**, exceptions that ocurred when
+handling the additional requests like `connection errors`, `time outs`, `TLS
+errors`, etc should be replaced by :class:`web_poet.exceptions.http.HttpRequestError`
+by raising it explicitly.
```
kmike, line 442: I think we should document explicitly (here and in HttpRequestError docstring) that this error shouldn't be raised when the response is received, but the status code is not 200. Or that it should be raised :) We should have the behavior clear.

tests/test_page_inputs.py
```diff
@@ -178,7 +262,7 @@ def test_http_response_json():
     response = HttpResponse(url, body=b'{"key": "value"}')
     assert response.json() == {"key": "value"}
 
-    response = HttpResponse(url, '{"ключ": "значение"}'.encode("utf8"))
+    response = HttpResponse(url, body='{"ключ": "значение"}'.encode("utf8"))
```
kmike, line 265: Do we have tests which check that positional body works?

docs/advanced/additional-requests.rst
BurnzZ: Gotcha, that makes sense. Thanks for raising it!

I moved the **Exception Handling** section right after discussing how to make additional requests so it's a smoother transition to users developing POs _(as you've pointed out)_. Also written down more examples regarding execption handling in https://github.com/scrapinghub/web-poet/pull/22/commits/7947d588c35edc9e6498480d36281e1d93b8ba2e

Regarding `batch_requests()`, thanks for pointing out about retrieving partial data as I haven't considered it before. The behavior is now updated in https://github.com/scrapinghub/web-poet/pull/22/commits/5d2a7519a0f5b4d0e23e62bd1f2eae099620b37e

docs/advanced/additional-requests.rst
BurnzZ: Addressed this in https://github.com/scrapinghub/web-poet/pull/22/commits/7947d588c35edc9e6498480d36281e1d93b8ba2e

tests/test_page_inputs.py
BurnzZ: Good call! Tweaked some of the tests to test out the positional args alongside the keyword args. https://github.com/scrapinghub/web-poet/pull/22/commits/e7b4f081de979ce35d258629781d4342a40d1ac4

web_poet/requests.py
```diff
@@ -0,0 +1,132 @@
...
+        :class:`~.HttpRequest` instances that would be batch executed.
+
+        If any of the :class:`~.HttpRequest` raises an exception upon execution,
+        they will be suppressed. Instead, the actual exception is returned
+        alongside any successful :class:`~.HttpResponse`. This enables salvaging
+        any usable responses despite any possible failures.
+        """
+
+        coroutines = [self.request_downloader(r) for r in requests]
+        responses = await asyncio.gather(*coroutines, return_exceptions=True)
```
kmike, line 131: What do you think about exposing return_exceptions option? It could make sense to keep the interface similar to asyncio.gather. I also wonder if we should keep False as a default; it may be more intuitive. Otherwise, user would always need to add some code to the PO itself to handle the exceptions; with return_exceptions=False an error would be just raised & propagated, which may be a right behavior in many cases.

web_poet/requests.py
```diff
@@ -0,0 +1,132 @@
...
+"""
+
+import asyncio
+import logging
+from contextvars import ContextVar
+from typing import Optional, List, Callable, Union
+
+import attrs
+
+from web_poet.page_inputs import HttpResponse, HttpRequest, Headers, Body
```
kmike, line 18: What do you think about separating types like Body and "real" page inputs like HttpResponseBody further? My worry is that user can actually add Body as a dependency, and it'd work (andi supports unions). This import reads as if Body is a page input. But it's unlikely that'd be an intention, e.g. a provider which provides bytes may break it.

How to separate them further? I can see a few options:

1) move them to a separate module. This doesn't seem to be common to move such type definitions to a separate module, but it may be ok.
2) Rename to something like BodyT or BodyType.
3) Drop it or make private, use `Union[bytes, HttpResponseBody]`
4) somethins else?

web_poet/requests.py
```diff
@@ -0,0 +1,132 @@
...
+        request.
+        """
+        return await self.request(url=url, method="GET", headers=headers)
+
+    async def post(
+        self,
+        url: str,
+        *,
+        headers: Optional[Headers] = None,
+        body: Optional[Body] = None,
```
kmike, line 111: If I'm not mistaken, technically, even "bytes" annotation could be ok, because HttpRequestBody is a subclass of bytes. But it'd won't allow to show the intent here.

web_poet/page_inputs.py
```diff
@@ -123,6 +145,27 @@ def declared_encoding(self) -> Optional[str]:
         return http_content_type_encoding(content_type)
 
 
+Mapping = Dict[str, str]
```
kmike, line 148: a more specific name might be better, as a reader might confuse it with https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping when reading function definitions

web_poet/__init__.py
```diff
@@ -1,6 +1,19 @@
 from .pages import WebPage, ItemPage, ItemWebPage, Injectable
-from .page_inputs import HttpResponse, HttpResponseBody, HttpResponseHeaders
+from .requests import (
+    request_backend_var,
+    HttpClient,
```
kmike, line 4: what's interesting is that HttpClient also can be a PO dependency :) I wonder if we should part from having page_inputs module, and just split it. E.g. web_poet.http, or something like this.

web_poet/requests.py
BurnzZ: Exposing `return_exceptions` would be a good idea, as it could prevent some additional effort from the the user to manually tweak the behavior. 👍 Updated this on https://github.com/scrapinghub/web-poet/pull/22/commits/bc4baea396de7227ff0dacb1db33c1f2b2b165a0

---

Regarding on having its _**default**_ value set to `False`: I do agree that it would be consistent with the underlying `asyncio.gather()` API default expectations. 

However, I'm not quite sure if we're encouraging the right behavior to the user. For our use case, executing HTTP requests incurs more cost than other coroutine operations that asyncio could have. As such, we'd be interested in encouraging the behavior of salvaging any usable HTTP Responses that the user can. Having the exceptions err up when `return_exceptions=False` would lose track of usable responses altogether.

Moreover, it doesn't promote future possibilities like retrying a specific failed request from a `batch_request()` since the developer doesn't know which request had the issue. 

Let me know what you think. :)

web_poet/requests.py
BurnzZ: Good catch! `Headers` and `Body` types are not being used anywhere inside the **page_inputs.py** module. They're only being used in the **requests.py** so we can move it there. In addition, we can also make them private to discourage users from importing them in their projects.

Updated this in https://github.com/scrapinghub/web-poet/pull/22/commits/9f43187c5944c0cf0ce3bc122f3c00580262dec1

web_poet/page_inputs.py
BurnzZ: Good point! Renamed this into `_StrMapping` in https://github.com/scrapinghub/web-poet/pull/22/commits/9f43187c5944c0cf0ce3bc122f3c00580262dec1.

web_poet/__init__.py
BurnzZ: This is a good thought to ponder. I think we have 3 main choices for this which are:

1. Move `HttpClient` from **requests.py** to **page_inputs.py**
2. Create a **page_inputs/** subpackage which houses **page_inputs/meta.py** and **page_inputs/http.py** (`HttpClient` is here)
3. Disband the **page_inputs.py** and simply have modules like **meta.py** and **http.py** (`HttpClient` is here)

I'm torn between 1 and 2 since having some sort of subpackage/module named **page_inputs** helps portray the idea that anything within it could be a PO dependency.

What do you think? Perhaps there might be other good options aside from these that I might've missed?

kmike: I'm also not sure :) Let's handle it separately.

docs/advanced/additional-requests.rst
```diff
@@ -0,0 +1,802 @@
...
+            return web_poet.HttpRequest(
+                url="https://www.api.example.com/product-pagination/",
+                method="POST",
+                headers={
+                    "Content-Type": "application/json;charset=UTF-8"
+                },
+                body=json.dumps(
+                    {
+                        "Page": page_num,
+                        "ProductID": item["product_id"],
```
kmike, line 440: item is not defined

BurnzZ: Nice catch! Updated in https://github.com/scrapinghub/web-poet/pull/22/commits/80f95a8f9a882e4fb3579f546e10bd54801z6e172.

web_poet/requests.py
BurnzZ: Hi @kmike , as discussed offline last week, we'll be having the _**default**_ value of `return_exceptions` set to `False`.

This leads to a more meaningful exception message that explicitly logs `HttpRequestError` as opposed to something like _"css method not found in ..."_ since the developer has not automatically handled the returned exception. Raising them by default makes the developer more aware on how to handle them.

Updated in https://github.com/scrapinghub/web-poet/pull/22/commits/80f95a8f9a882e4fb3579f546e10bd548016e172.

web_poet/__init__.py
BurnzZ: Sure. I've filed a draft in https://github.com/scrapinghub/web-poet/pull/37 which explores **option 2** to see how it may look like.

BurnzZ:
Hello @kmike, Thanks for reviewing this PR again last week. 🙌 

Aside from these large changes:
- update `return_exceptions` default from `True` to `False` https://github.com/scrapinghub/web-poet/pull/22/commits/80f95a8f9a882e4fb3579f546e10bd548016e172
- expose `execute()` for single requests and rename `batch_requests()` into `batch_execute()` https://github.com/scrapinghub/web-poet/pull/22/commits/5d82a37cc8816be0950b66cc1d7bee5ebd06e206

Small doc-related changes are included in https://github.com/scrapinghub/web-poet/pull/22/commits/753e6ad598f5925289fd0085d6900a279ddac69b which includes _(but not limited to)_:
- fixing some incorrect code samples _(they can't run due to syntax errors)_
- emphasize that order is preserved for `batch_execute()`
- mention that `batch_execute()` is merely a shortcut for `asyncio.gather()` and other asyncio functions like `asyncio.as_completed()` can be used depending on the need
- mention what are the expectations for redirections

Lastly, I've created another PR in https://github.com/scrapinghub/web-poet/pull/38 to explore the default behavior of raising exceptions when receiving 400-5xx responses.

kmike:
Thanks so much @BurnzZ! I've checked the PR again, it looks good 💪

We might think about re-organizing docs a bit, but the content is there; this can be done later.

BurnzZ:
Thanks for your patience in reviewing this one @kmike ! 🙏

PR #23, opened 2022-02-18
Introduce Meta as a way to pass information inside a PO

BurnzZ:
**NOTES:** This is built on top of the https://github.com/scrapinghub/web-poet/pull/22 branch.

### Progress

- [x] Docs
- [x] Tests
- [x] Changelog

### DIFF
--- a/web_poet/page_inputs.py
+++ b/web_poet/page_inputs.py
@@ -37,13 +37,10 @@ class Meta(dict):
     """Container class that could contain any arbitrary data to be passed into
     a Page Object.
 
-    This is basically a subclass of a ``dict`` but adds some additional
-    functionalities to ensure consistent and compatible Page Objects across
-    different use cases:
-
-    * Ensures that some params with data types that are difficult to
-      provide or pass like ``lambdas`` are checked. Otherwise, a ``ValueError``
-      is raised.
+    This is basically a subclass of a ``dict`` that adds the ability to check
+    if any of the assigned values are not allowed. This ensures that some input
+    parameters with data types that are difficult to provide or pass via CLI
+    like ``lambdas`` are checked. Otherwise, a ``ValueError`` is raised.
     """
 
     # Any "value" that returns True for the functions here are not allowed.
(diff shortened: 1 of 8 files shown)


web_poet/page_inputs.py
```diff
@@ -24,7 +26,60 @@ class ResponseData:
...
+    a Page Object.
+
+    This is basically a subclass of a ``dict`` that adds the ability to check
+    if any of the assigned values are not allowed. This ensures that some input
+    parameters with data types that are difficult to provide or pass via CLI
+    like ``lambdas`` are checked. Otherwise, a ``ValueError`` is raised.
+    """
+
+    # Any "value" that returns True for the functions here are not allowed.
+    restrictions: Dict = {
```
kmike, line 47: I wonder if we should remove restrictions support, and make Meta just a subclass of dict. 

The problem is that the current restrictions are not enough: there are many other objects which may be not serializable - instances of custom classes, lambdas within dictionaries / lists, etc. There is an issue with an implementation as well - only values are checked, not keys. At the same time, it's technically possible to all make objects listed here serializable, by using various hooks in libraries. 

It seems the only way to check it reliably is to try to serialize meta - but it'd assume a certain method of serializing, and a performance overhead. So, I'm worried that with the current approach the code may be growing over time (trying to support more cases), while still not solving the problem.

Maybe we can add something later, or maybe the current code is helpful as-is. But I think it may not be the right time to add this validation.

BurnzZ: It would seem that we have these options:

**1. Improve the current restrictions by adding more checks and checking the key as well.**

This would always have the risk of missing out on something since it requires an exhaustive barrage of checks to be made. It would also make the code more heavy as more checks are added later on.

**2. Simplify the restriction mechanism by checking if serializing it raises an error**

This would make the restriction checks easier as it'd simply be a `try-except` block in lieu of the current hardcoded checks.

This would need a default serializer to do the validation but it can also be overridden to something else if needed.

**3. Same as option 2 but not during `__init__` and `__setattr__`**.

This reduces the performance overheads on unneeded validations when users won't need it. If they do, they need to deliberately run a method to see if it's serializable or not.

**4. Remove it completely.**

Reduces the `web_poet.Meta` into a simple Python `dict` subclass.

----

I believe that we should at least have a simple validation check in place so that users are more deliberate on the keys and values that they try to pass. It could be the case that they don't attempt to seralize POs at all, and that's okay. 

On the other hand, some users might need to serialize POs at some point. It would be handy for them to realize that the mechanism of validating serializability is present which encourages conscious usage of `web_poet.Meta`.

What do you think?

BurnzZ: I'm not quite sure about **Option 5** since `scrapy-poet` doesn't need some serialization for Meta at its current state. Rather, other handlers of `web-poet`'s Meta like a platform deployment tool would need it, and thus, `web-poet` needs to have the validation in place.

For now, I think we can simplify this PR by going with **Option 6**. :) I'll split the validation logic of this PR into another one to be further discussed later on.

BurnzZ: Move the validation logic to this separate PR: https://github.com/scrapinghub/web-poet/pull/32/files

kmike:
Looks good, thanks @BurnzZ!

Issue #18, opened 2021-12-13
introduce more pages in addition to WebPage: JsonWebPage, CsvWebPage, etc


BurnzZ:
This issue aims to discuss adding more page handlers as proposed by @gatufo.

----

Currently, we only have `web_poet.pages.WebPage` which supports selectors like `css()` and `xpath()` on HTML pages.

However, there are also other types of pages we could support:

- `JsonWebPage` which could have `jmespath` queries
- `CsvWebPage` which could have a "unified" way to read them _(usually used as seed data inputs in spiders)_
- `PlainTextWebPage` which could also have a "unified" way to read newline-separated data _(also usually use for seed URL inputs)_

kmike:
If something like this is useful, it makes total sense to add it to web-poet. 
I'd start with an implementation in a project-specific repo, and then move commonly used code to web-poet.

kmike:
Do you think we can close it, for the same reason as https://github.com/scrapinghub/web-poet/pull/25?

BurnzZ:
@kmike , yes, we can re-open this one again later on should users ask for it.

PR #25, opened 2022-02-23
create ItemJsonPage, JsonPage, and JsonResponseShortcutsMixin

BurnzZ:
Addresses the need for JSON Shortcuts as described in https://github.com/scrapinghub/web-poet/issues/18.

I noticed that there's a growing need for it in our users. They typicall write this repetitively in their projects. Thus, it's worth adding support in this core library itself.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,8 @@ TBR
 
 * removed support for Python 3.6
 * added support for Python 3.10
+* added ``web_poet.ItemJsonPage``, ``web_poet.JsonPage``, and
+  ``web_poet.mixins.JsonResponseShortcutsMixin``.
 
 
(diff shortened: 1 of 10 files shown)


kmike:
This is partially addressed in https://github.com/scrapinghub/web-poet/pull/29, where HttpResponseBody got .json() method.

kmike:
Hey @BurnzZ! What do you think about closing it? Or do you see some use cases not covered by .json shortcut?

BurnzZ:
Hi @kmike , https://github.com/scrapinghub/web-poet/pull/30 covered a lot of the cases here. I think the only thing left is the ability to do jmespath queries. It's only good to have though and not really essential. We can tackle that in a separate PR. 👍 Closing this one out.

PR #27, opened 2022-02-28
introduce PageObjectRegistry with @hande_urls annotations

BurnzZ:
This is built as a light-weight version of the new features introduce by this PR: https://github.com/scrapinghub/web-poet/pull/16/files

In contrast, this only contains a minimal set of features that allows developers to:
- import and use the `default_registry`'s `@handle_urls` annotation for declaring `OverrideRule`s.
- use `search_overrides()` to find specific `OverrideRule`s amased from importing multiple Page Object Projects (POP).

This means that other non-essential features from the old PR aren't included:
- `registry_pool` which allows keeping track of other non-default `PageObjectRegistry` instances
- CLI-tool for displaying all of the `OverrideRule`s
- `copy_overrides_from()` which allows copying of `OverrideRule`s from multiple non-default `PageObjectRegistry` instances
- `remove_overrides()` since an **exclusion-list** strategy may be brittle if the underlying source changes. This results in some unintentional `OverrideRules` to be present in the output if the developer isn't aware of them.
- `replace_overrides()`  since a more explicit alternative would be creating a new `OverrideRule` which has the intended attribute replacements.
- the `filter` param of `get_overrides()` since it could break the developer's expectation if new `OverrideRule`s are added to the filter module which could lead to unintentional `OverrideRules` to be present.

### Progress:

* [x] Tests
* [x] rewrite docs to focus on more minimal usage

### To address in other PRs:

* POP entry point - https://github.com/scrapinghub/web-poet/pull/28
* `OverrideRule.instead_of` to support data types as well

### DIFF
--- a/docs/intro/overrides.rst
+++ b/docs/intro/overrides.rst
@@ -109,7 +109,7 @@ Retrieving all available Overrides
 ----------------------------------
 
 The :meth:`~.PageObjectRegistry.get_overrides` method from the ``web_poet.default_registry``
-allows discovery and retrieval of  all :class:`~.OverrideRule` from your project.
+allows retrieval of  all :class:`~.OverrideRule` in the given registry.
 Following from our example above, using it would be:
 
 .. code-block:: python
@@ -130,12 +130,12 @@ in the :class:`~.OverrideRule` to be written into ``web_poet.default_registry``.
 
     :meth:`~.PageObjectRegistry.get_overrides` relies on the fact that all essential
     packages/modules which contains the :func:`web_poet.handle_urls`
-    annotations are properly loaded.
+    annotations are properly loaded `(i.e imported)`.
 
-    Thus, for cases like importing Page Objects from another external package,
-    you'd need to properly load all ``@handle_urls`` annotations from the external
-    source. This ensures that the external Page Objects have their annotations
-    properly loaded.
+    Thus, for cases like importing and using Page Objects from other external packages,
+    the ``@handle_urls`` annotations from these external sources must be read and
+    processed properly. This ensures that the external Page Objects have all of their
+    :class:`~.OverrideRule` present.
 
     This can be done via the function named :func:`~.web_poet.overrides.consume_modules`.
     Here's an example:
@@ -146,9 +146,6 @@ in the :class:`~.OverrideRule` to be written into ``web_poet.default_registry``.
 
         consume_modules("external_package_A.po", "another_ext_package.lib")
         rules = default_registry.get_overrides()
-
-        # Fortunately, `get_overrides()` provides a shortcut for the lines above:
-        rules = default_registry.get_overrides(consume=["external_package_A.po", "another_ext_package.lib"])
 
     The next section explores this caveat further.
 
@@ -187,17 +184,14 @@ This can be done something like:
     from web_poet import default_registry, consume_modules
 
     # ❌ Remember that this wouldn't retrieve any rules at all since the
-    # annotations are NOT properly loaded.
+    # annotations are NOT properly imported.
     rules = default_registry.get_overrides()
     print(rules)  # []
 
     # ✅ Instead, you need to run the following so that all of the Page
-    # Objects in the external packages are recursively loaded.
+    # Objects in the external packages are recursively imported.
     consume_modules("ecommerce_page_objects", "gadget_sites_page_objects")
     rules = default_registry.get_overrides()
-
-    # Alternatively, this could be shortened using:
-    rules = default_registry.get_overrides(consume=["ecommerce_page_objects", "gadget_sites_page_objects"])
 
     # The collected rules would then be as follows:
     print(rules)
@@ -209,9 +203,9 @@ This can be done something like:
 .. note::
 
     Once :func:`~.web_poet.overrides.consume_modules` is called, then all
-    external Page Objects are loaded and available for the entire runtime
-    duration. Calling :func:`~.web_poet.overrides.consume_modules` again makes
-    no difference unless a new set of modules are provided.
+    external Page Objects are recursively imported and available for the entire
+    runtime duration. Calling :func:`~.web_poet.overrides.consume_modules` again
+    makes no difference unless a new set of modules are provided.
 
 .. _`intro-rule-subset`:
 
@@ -221,7 +215,7 @@ Using only a subset of the available OverrideRules
 Suppose that the use case from the previous section has changed wherein a
 subset of :class:`~.OverrideRule` would be used. This could be achieved by
 using the :meth:`~.PageObjectRegistry.search_overrides` method which allows for
-convenient selection of a subset of rules from the ``default_registry``.
+convenient selection of a subset of rules from a given registry.
 
 Here's an example of how you could manually select the rules using the
 :meth:`~.PageObjectRegistry.search_overrides` method instead:
@@ -274,7 +268,7 @@ You might've observed from the previous section that retrieving the list of all
 :class:`~.OverrideRule` from two different external packages may result in a
 conflict. 
 
-We can take a look at the rules for **#2** and **#3** when we were loading all
+We can take a look at the rules for **#2** and **#3** when we were importing all
 available rules:
 
 .. code-block:: python
--- a/docs/intro/overrides.rst
+++ b/docs/intro/overrides.rst
@@ -157,6 +157,8 @@ Developers have the option to import existing Page Objects alongside the
 :class:`~.OverrideRule` attached to them. This section aims to showcase different
 scenarios that come up when using multiple Page Object Projects.
 
+.. _`intro-rule-all`:
+
 Using all available OverrideRules from multiple Page Object Projects
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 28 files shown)


docs/intro/overrides.rst
```diff
@@ -0,0 +1,349 @@
...
+    The URL patterns declared in the ``@handle_urls`` annotation can still be
+    further customized. You can read some of the specific parameters in the
+    :ref:`API section <api-overrides>` of :func:`web_poet.handle_urls`.
+
+
+Retrieving all available Overrides
+----------------------------------
+
+The :meth:`~.PageObjectRegistry.get_overrides` method from the ``web_poet.default_registry``
+allows discovery and retrieval of  all :class:`~.OverrideRule` from your project.
```
kmike, line 112: I think get_overrides is not doing discovery, and also it doesn't limit rules to a certain project:

```suggestion
allows retrieval of all :class:`~.OverrideRule` in the registry.
```

docs/intro/overrides.rst
```diff
@@ -0,0 +1,349 @@
...
+
+Remember that using ``@handle_urls`` to annotate the Page Objects would result
+in the :class:`~.OverrideRule` to be written into ``web_poet.default_registry``.
+
+
+.. warning::
+
+    :meth:`~.PageObjectRegistry.get_overrides` relies on the fact that all essential
+    packages/modules which contains the :func:`web_poet.handle_urls`
+    annotations are properly loaded.
```
kmike, line 133:
```suggestion
    annotations are properly loaded (i.e. imported).
```

docs/intro/overrides.rst
```diff
@@ -0,0 +1,349 @@
...
+
+    .. code-block:: python
+
+        from web_poet import default_registry, consume_modules
+
+        consume_modules("external_package_A.po", "another_ext_package.lib")
+        rules = default_registry.get_overrides()
+
+        # Fortunately, `get_overrides()` provides a shortcut for the lines above:
+        rules = default_registry.get_overrides(consume=["external_package_A.po", "another_ext_package.lib"])
```
kmike, line 151: Because of the caveats I'm not sure it makes sense to provide `consume` keyword for get_overrides - the code above reads as it'd return only POs from the listed modules. With 

```py
consume_modules("external_package_A.po", "another_ext_package.lib")
rules = default_registry.get_overrides()
```

the caveat is more clear IMHO, and it's the same amount of code. Do you think we can remove this method?

docs/intro/overrides.rst
kmike: Overall, I think it may be a bit more clear if in some places further in the docs instead of "loaded" is replaced with "imported". consume_modules is not doing anything fancy, it just imports modules.

docs/intro/overrides.rst
```diff
@@ -0,0 +1,349 @@
...
+
+.. _`intro-rule-subset`:
+
+Using only a subset of the available OverrideRules
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Suppose that the use case from the previous section has changed wherein a
+subset of :class:`~.OverrideRule` would be used. This could be achieved by
+using the :meth:`~.PageObjectRegistry.search_overrides` method which allows for
+convenient selection of a subset of rules from the ``default_registry``.
```
kmike, line 224:
```suggestion
convenient selection of a subset of rules from the registry.
```

docs/intro/overrides.rst
```diff
@@ -0,0 +1,349 @@
...
+
+    # 2. OverrideRule(for_patterns=Patterns(include=['site_2.com'], exclude=[], priority=500), use=<class 'ecommerce_page_objects.site_2.EcomSite2'>, instead_of=<class 'common_items.ProductGenericPage'>, meta={})
+    # 3. OverrideRule(for_patterns=Patterns(include=['site_2.com'], exclude=[], priority=500), use=<class 'gadget_sites_page_objects.site_2.GadgetSite2'>, instead_of=<class 'common_items.ProductGenericPage'>, meta={})
+
+Notice that the ``instead_of`` param are the same and only the ``use`` param
+remained different.
+
+There are two main ways we recommend in solving this.
+
+**1. Priority Resolution**
```
kmike, line 310: I think the main use case here could be the following:

1. There is a library of page objects
2. In your project you want to have a different version for one of POs available in the library

In this case there is no need to modify any override rules, it's a matter of using handle_urls decorator in your own project, and setting a priority higher than the default.

docs/intro/overrides.rst
```diff
@@ -0,0 +1,349 @@
...
+When the last resort of ``priority``-resolution doesn't work, then you could always
+specifically select the list of :class:`~.OverrideRule` you want to use.
+
+We **recommend** in creating an **inclusion**-list rather than an **exclusion**-list
+since the latter is quite brittle. For instance, an external package you're using
+has updated its rules and the exlusion strategy misses out on a few rules that
+were recently added. This could lead to a `silent-error` of receiving a different
+set of rules than expected.
+
+For this approach, you can use the :meth:`~.PageObjectRegistry.search_overrides`
```
kmike, line 348: Hm, I was thinking about importing POs directly. How would you do inclusion in search_overrides?

web_poet/overrides.py
```diff
@@ -0,0 +1,305 @@
...
+        .. code-block:: python
+
+            rules = registry.search_overrides(use=ProductPO, instead_of=GenericPO)
+            print(len(rules))  # 1
+
+        """
+
+        # Short-circuit operation if "use" is the only search param used, since
+        # we know that it's being used as the dict key.
+        if set(["use"]) == kwargs.keys():
```
kmike, line 216:
```suggestion
        if {"use"} == kwargs.keys():
```

web_poet/overrides.py
```diff
@@ -0,0 +1,305 @@
...
+        from web_poet import default_registry, consume_modules
+
+        consume_modules("other_external_pkg.po", "another_pkg.lib")
+        rules = default_registry.get_overrides()
+
+    For this case, the :class:`~.OverrideRule` are coming from:
+
+        - ``my_page_obj_project`` `(since it's the same module as the file above)`
+        - ``other_external_pkg.po``
+        - ``another_pkg.lib``
```
kmike, line 281: + any other modules imported in the same process. I think the example only works if the code above is a whole script, not a module within a larger package.

web_poet/overrides.py
```diff
@@ -0,0 +1,305 @@
...
+
+    If the ``default_registry`` had other ``@handle_urls`` annotations outside
+    of the packages/modules listed above, then the corresponding
+    :class:`~.OverrideRule` won't be returned.
```
kmike, line 285: Again, this is only true if they're not imported in some other way. 

I think it may be fine to just explain that consume_module imports modules recursively.

kmike:
@BurnzZ the PR looks good to me. There are some ways to improve the docs, but it can be done later. 

+1 to merge after removing "consume" option from get_overrides method.

docs/intro/overrides.rst
BurnzZ: Good point. Updated this on https://github.com/scrapinghub/web-poet/pull/27/commits/37e5ba1a3696f629801478fe78ea1729164f39df

docs/intro/overrides.rst
BurnzZ: Updated this on https://github.com/scrapinghub/web-poet/pull/27/commits/37e5ba1a3696f629801478fe78ea1729164f39df

docs/intro/overrides.rst
BurnzZ: Yeah, we can make it more explicit by having to access `consume_modules()` externally.

Removed this on https://github.com/scrapinghub/web-poet/pull/27/commits/37e5ba1a3696f629801478fe78ea1729164f39df

docs/intro/overrides.rst
BurnzZ: Updated this on https://github.com/scrapinghub/web-poet/pull/27/commits/37e5ba1a3696f629801478fe78ea1729164f39df

web_poet/overrides.py
BurnzZ: Good one. Updated this on https://github.com/scrapinghub/web-poet/pull/27/commits/37e5ba1a3696f629801478fe78ea1729164f39df

web_poet/overrides.py
BurnzZ: Updated this in https://github.com/scrapinghub/web-poet/pull/27/commits/37e5ba1a3696f629801478fe78ea1729164f39df

web_poet/overrides.py
```diff
@@ -0,0 +1,305 @@
...
+    for_patterns: Patterns
+    use: Callable
+    instead_of: Callable
+    meta: Dict[str, Any] = field(default_factory=dict)
+
+    def __hash__(self):
+        return hash((self.for_patterns, self.use, self.instead_of))
+
+
+def _as_list(value: Optional[Strings]) -> List[str]:
```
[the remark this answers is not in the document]
BurnzZ: Moved it in https://github.com/scrapinghub/web-poet/pull/27/commits/50c0046daca9dae90ba0ed1223d2ea91e98768df

docs/intro/overrides.rst
BurnzZ: @kmike to clarify, are you suggesting something like the following?

If we have the conflict like:

```python

rules = default_registry.get_overrides()
print(rules)
# OverrideRule(for_patterns=Patterns(include=['site_2.com'], exclude=[], priority=500), use=<class 'ecommerce_page_objects.site_2.EcomSite2'>, instead_of=<class 'common_items.ProductGenericPage'>, meta={})
# OverrideRule(for_patterns=Patterns(include=['site_2.com'], exclude=[], priority=500), use=<class 'gadget_sites_page_objects.site_2.GadgetSite2'>, instead_of=<class 'common_items.ProductGenericPage'>, meta={})
```

It can be fixed in the project as something like:

```python
import handle_urls

handle_urls(ProductGenericPage, overrides=ProductGenericPage, priority=600)(EcomSite2)
```

At the moment, this won't work since the `PageObjectRegistry` stores the PO class as the key, and if a rule is already existing for a given PO, it ignores new rules attempted to be registered. [Reference](https://github.com/scrapinghub/web-poet/pull/27/files#diff-409cb5fe79e7d7d5ede41109573e31bd847a147a966f477de80c99279bcdf479R153-R161).

Do you think it's worth refactoring this from `Dict[<PO class>, OverrideRule]` into `Dict[<PO class>, List[OverrideRule]]` instead?

Though this would affect the other implementations like `search_overrides()` and how it's being read in **scrapy-poet** as well. Nonetheless, we're still quite early in the process and we can still accommodate this change.

BurnzZ: @kmike that makes sense! Thanks for explaining it as I initially had the wrong idea. This use case is indeed a common one and it deserves it's own section which I've added in https://github.com/scrapinghub/web-poet/pull/27/commits/7944956802e8c38d939c42b7866d2915f853b1fe.

docs/intro/overrides.rst
BurnzZ: Added another in-depth example for `search_overrides()` in https://github.com/scrapinghub/web-poet/pull/27/commits/74573319c7b495dbf16dc615be846af51721d508.

BurnzZ:
Thanks for the review @kmike ! I've addressed all of your comments alongside the specific commit for the change. Let me know if there's anything left to update here. 🙏

web_poet/utils.py
```diff
@@ -15,3 +16,22 @@ def new_method(self, *args, **kwargs):
...
+    >>> as_list("foo")
+    ['foo']
+    >>> as_list(123)
+    [123]
+    >>> as_list(["foo", "bar", 123])
+    ['foo', 'bar', 123]
+    """
+    if value is None:
+        return []
+    if not isinstance(value, list):
```
kmike, line 35: It'd be nice to add a test for tuple, e.g. what should `as_list(("foo", "bar", 123))` return? Should it be a single-item list, or a tuple converted to a list?

BurnzZ: Good point! After looking into this further, I've added more input types that `as_list()` should also be able to handle and convert. Updated in https://github.com/scrapinghub/web-poet/pull/27/commits/5eab374d5767fe2fed821351a5c10fd1e9c9631e.

kmike:
Thanks @BurnzZ!

PR #24, opened 2022-02-18
refactor ResponseData into their own subcomponents

BurnzZ:
Hi @kmike , creating this separate PR to dive deep into what we've discussed previously about refactoring `ResponseData` to its subcomponents.

**Progress:**

- [x] Proposed refactored changes
- [ ] Docs
- [x] Changelog
- [x] Tests

PR #29, opened 2022-03-16
HttpResponse object

kmike:
hey @BurnzZ! This is a follow-up to https://github.com/scrapinghub/web-poet/pull/24:

* useful utilities are ported from Scrapy's Response and TextResponse
* ResponseData is renamed to HttpResponse

I haven't checked if the code works :) Also, it's not a copy-paste port from Scrapy; the code is cleaned up and adjusted (which could mean new bugs).

Not handled here:

* there is overlapping code in ResponseShortcutsMixin, I haven't touched it.
* because of the above, response.urljoin is not implemented
* there are some FIXME's in the code
* tests and docs are broken
* we'd need to think about code organization when we add support for BrowserHtml type. There will be a lot of duplication.

### DIFF
--- a/web_poet/page_inputs.py
+++ b/web_poet/page_inputs.py
@@ -1,13 +1,34 @@
-from typing import Optional, Dict, List, Iterable, Tuple
+import json
+from typing import Optional, Dict, List
 
-import attr
+import attrs
 from multidict import CIMultiDict
+import parsel
+from w3lib.encoding import (
+    html_to_unicode,
+    html_body_declared_encoding,
+    resolve_encoding,
+    http_content_type_encoding
+)
+
+from .utils import memoizemethod_noargs
 
 
 class HttpResponseBody(bytes):
     """A container for holding the raw HTTP response body in bytes format."""
 
-    pass
+    @memoizemethod_noargs
+    def declared_encoding(self) -> Optional[str]:
+        """ Return the encoding specified in meta tags in the html body,
+        or ``None`` if no suitable encoding was found """
+        return html_body_declared_encoding(self)
+
+    @memoizemethod_noargs
+    def json(self):
+        """
+        Deserialize a JSON document to a Python object.
+        """
+        return json.loads(self)
 
 
 class HttpResponseHeaders(CIMultiDict):
@@ -53,18 +74,20 @@ class HttpResponseHeaders(CIMultiDict):
         """
         return cls([(pair["name"], pair["value"]) for pair in arg])
 
+    def declared_encoding(self) -> Optional[str]:
+        """ Return encoding detected from the Content-Type header, or None
+        if encoding is not found """
+        content_type = self.get('Content-Type', '')
+        return http_content_type_encoding(content_type)
 
-@attr.define
-class ResponseData:
+
+@attrs.define(auto_attribs=False)
+class HttpResponse:
     """A container for the contents of a response, downloaded directly using an
     HTTP client.
 
-    ``url`` should be an URL of the response (after all redirects),
-    not an URL of the request, if possible.
-
-    ``html`` should be content of the HTTP body, converted to unicode
-    using the detected encoding of the response, preferably according
-    to the web browser rules (respecting Content-Type header, etc.)
+    ``url`` should be a URL of the response (after all redirects),
+    not a URL of the request, if possible.
 
     ``body`` contains the raw HTTP response body.
 
@@ -76,10 +99,88 @@ class ResponseData:
     ``status`` should represent the int status code of the HTTP response.
 
     ``headers`` should contain the HTTP response headers.
+
+    ``encoding`` encoding of the response. If None (default), encoding
+    is auto-detected from headers and body content.
     """
 
-    url: str
-    html: str
-    body: Optional[HttpResponseBody] = None
-    status: Optional[int] = None
-    headers: Optional[HttpResponseHeaders] = None
+    url: str = attrs.field()
+    # FIXME: raise an error or handle the case when str is passed as body
+    body: HttpResponseBody = attrs.field(converter=HttpResponseBody)
+    status: Optional[int] = attrs.field(default=None)
+    headers: Optional[HttpResponseHeaders] = attrs.field(default=None)
+    _encoding: Optional[str] = attrs.field(default=None)
+
+    _DEFAULT_ENCODING = 'ascii'
+    _cached_text: Optional[str] = None
+
+    @property
+    def text(self) -> str:
+        """
+        Content of the HTTP body, converted to unicode
+        using the detected encoding of the response, according
+        to the web browser rules (respecting Content-Type header, etc.)
+        """
+        # Access self.encoding before self._cached_text, because
+        # there is a chance self._cached_text would be already populated
+        # while detecting the encoding
+        encoding = self.encoding
+        if self._cached_text is None:
+            fake_content_type_header = f'charset={encoding}'
+            encoding, text = html_to_unicode(fake_content_type_header, self.body)
+            self._cached_text = text
+        return self._cached_text
+
+    @property
+    def encoding(self):
+        """ Encoding of the response """
+        return (
+            self._encoding
+            or self._headers_declared_encoding()
+            or self._body_declared_encoding()
+            or self._body_inferred_encoding()
+        )
+
+    @property
+    @memoizemethod_noargs
+    def selector(self) -> parsel.Selector:
+        # XXX: should we pass base_url=self.url, as Scrapy does?
+        return parsel.Selector(text=self.text)
+
+    def xpath(self, query, **kwargs):
+        return self.selector.xpath(query, **kwargs)
+
+    def css(self, query):
+        return self.selector.css(query)
+
+    def json(self):
+        return self.body.json()
+
+    @memoizemethod_noargs
+    def _headers_declared_encoding(self):
+        if self.headers:
+            return self.headers.declared_encoding()
+
+    @memoizemethod_noargs
+    def _body_declared_encoding(self):
+        return self.body.declared_encoding()
+
+    @memoizemethod_noargs
+    def _body_inferred_encoding(self):
+        content_type = (self.headers or {}).get('Content-Type', '')
+        body_encoding, text = html_to_unicode(
+            content_type,
+            self.body,
+            auto_detect_fun=self._auto_detect_fun,
+            default_encoding=self._DEFAULT_ENCODING
+        )
+        self._cached_text = text
+        return body_encoding
+
+    def _auto_detect_fun(self, text):
+        for enc in (self._DEFAULT_ENCODING, 'utf-8', 'cp1252'):
+            try:
+                text.decode(enc)
+            except UnicodeError:
+                continue
+            return resolve_encoding(enc)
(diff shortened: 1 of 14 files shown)


web_poet/page_inputs.py
```diff
@@ -76,10 +99,88 @@ class ResponseData:
...
+    @memoizemethod_noargs
+    def selector(self) -> parsel.Selector:
+        # XXX: should we pass base_url=self.url, as Scrapy does?
+        return parsel.Selector(text=self.text)
+
+    def xpath(self, query, **kwargs):
+        return self.selector.xpath(query, **kwargs)
+
+    def css(self, query):
+        return self.selector.css(query)
```
BurnzZ, line 154: The `selector`, `xpath`, and `css` methods coincide with the `ResponseShortcutsMixin` like you've said. I don't believe they should be here as `ResponseData` is only concerned about representing the data compared to interacting with it. The `ResponseShortcutsMixin` should possess this ability.

web_poet/page_inputs.py
```diff
@@ -1,13 +1,34 @@
...
+        """ Return the encoding specified in meta tags in the html body,
+        or ``None`` if no suitable encoding was found """
+        return html_body_declared_encoding(self)
+
+    @memoizemethod_noargs
+    def json(self):
+        """
+        Deserialize a JSON document to a Python object.
+        """
+        return json.loads(self)
```
BurnzZ, line 31: Introducing a `json` helper here would mean that other types of HTTP Responses like `csv` files (which has various types of delimiters) have a strong position to be introduced here as well.

Instead, I think such functionality should be placed in the `pages.py` module, possible inheriting from the `WebPage` class as they offer more specialized features.

I would imagine instead of the current state which is `page.response.json` would become `page.json`, lifting it up one level higher when implemented in `pages.py`.

web_poet/page_inputs.py
```diff
@@ -53,18 +74,20 @@ def from_name_value_pairs(cls, arg: List[Dict]):
...
+        """ Return encoding detected from the Content-Type header, or None
+        if encoding is not found """
+        content_type = self.get('Content-Type', '')
+        return http_content_type_encoding(content_type)
+
 
-@attr.define
-class ResponseData:
+@attrs.define(auto_attribs=False)
+class HttpResponse:
```
BurnzZ, line 85: Renaming this from `ResponseData` into `HttpResponse` would warrant the `status` and `headers` to be **non-optional** for them to semantically represent a proper HTTP Response.

BurnzZ (commented):
Thanks creating this @kmike ! I now understand the body encoding mechanism in HTTP responses.

I've added some comments in some of the proposed aspects as I believe they could be trimmed down.

web_poet/page_inputs.py
kmike: I think the difference is that with these methods being in the response, you don't need to wrap responses in page objects anymore. Example form https://github.com/scrapinghub/web-poet/pull/22:

```py
    # Simulates clicking on a button that says "View All Images"
    response: web_poet.ResponseData = await self.http_client.get(
        f"https://api.example.com/v2/images?id={item['product_id']}"
    )
    page = web_poet.WebPage(response)
    item["images"] = page.css(".product-images img::attr(src)").getall()
```

can be rewritten as

```py
    # Simulates clicking on a button that says "View All Images"
    response = await self.http_client.get(
        f"https://api.example.com/v2/images?id={item['product_id']}"
    )
    item["images"] = response.css(".product-images img::attr(src)").getall()
```

It seems putting these methods to response itself may simplify many similar examples. 

In general, separation of concerns is good, but I'm not sure what are we separating here, by leaving ResponseData as a simple data container without any logic. Page objects still contain website-specific code. HttpResponse is still not tied to any particular framework, and doesn't imply any particular download method. It also makes the API more familiar, as it's very close to Scrapy's. 

ResponseShortcutsMixin is used as a mixin to pages, e.g. WebPage uses it. It makes certain implicit assumptions: there is a "response" attribute, named "response" and containing ResponseData. It enables the usage like the following:

```py
class MyPage(WebPage):
    def to_item(self):
        price = self.css(".price").get()
```

With methods available on responses, we don't need to make these implicit assumptions:
```py
class MyPage(ItemPage):
    def to_item(self):
        price = self.response.css(".price").get()
```

We can still provide some shortcut to make methods available on Page level, but it becomes less important.

web_poet/page_inputs.py
kmike: > Introducing a json helper here would mean that other types of HTTP Responses like csv files (which has various types of delimiters) have a strong position to be introduced here as well.

I was following Scrapy's TextResponse.json example here. Requests and aiohttp libraries also provide .json() method on response, it's really common. 

Initially I ported .json method to HttpResponse, but then realized that encoding detection is not needed (json.loads in stdlib could work on bytes, there is code in stdlib to handle this), so this method can be moved to HttpResponseBody. With the use case of making additional requests, emulating API calls, I think JSON is by far the most common data format; I don't see it expanding to csv. 

Other libraries (scrapy, requests, aiohttp) only provide shortcuts for json in their response classes. It's a bit different in Scrapy, which has an ability to change response class based on its detected type; we're not doing it here. Scrapy is also is not doing it for json specifically: it was an option, but in Scrapy there is no JsonResponse; instead, .json method is provided right on TextResponse.

> Instead, I think such functionality should be placed in the pages.py module, possible inheriting from the WebPage class as they offer more specialized features.

That's the original design of ResponseShortcutsMixin; I've written it, but I'm not sure I like it :) It's limiting, harder to figure out what's going on. I think what we shouldn't change is "all website-specific code must be in page objects". How to organize utility code is a different discussion. 

I think that's fine to have some of it in page classes, some as separate functions; now we're introducing more logic on the dependency classes as well. But this already started - e.g. HttpHeaders is not a simple data container, it has quite a lot of logic to handle multiple values for the same key, case sensitivity, etc.

> I would imagine instead of the current state which is page.response.json would become page.json, lifting it up one level higher when implemented in pages.py.

Yeah, that's a concern: there is no .json shortcut on the page itself. I'm not too worried about .json shortcut, as the most common case would be making additional request, where it actually simplifies things (no need to create a wrapper page). It's a downgrade for .css and .xpath methods though, but we can provide the same helpers as before.

web_poet/page_inputs.py
kmike: Getting back to the example:

```py
@attrs.define
class MyPage(ItemPage, ReponseShortcutsMixin):
    response: HttpResponse

    def to_item(self):
        price = self.css(".price").get()
```

I really don't like how concerns are separated here. MyPage class defines "response" attribute, but this attribute is actually managed by ReponseShortcutsMixin. The code is not "local": to understand how it works, you need to figure out that self.css comes from ReponseShortcutsMixin, and that ReponseShortcutsMixin uses MyPage.response, there is a strong bi-directional connection between MyPage and ReponseShortcutsMixin. 

You don't have any of these issues with

```py
@attrs.define
class MyPage(ItemPage):
    response: HttpResponse

    def to_item(self):
        price = self.response.css(".price").get()
```

It's very top-down in this example: MyPage defines response, and can change it whatever needed - rename the variable, use 2 different responses (if supported), etc. to_item method uses response to extract data from it.

web_poet/page_inputs.py
kmike: Yeah, that's a good call out. I left them optional, and default to None mostly because of an example you have in the docs (loading html files from disk). Status and headers are not needed for extraction in many cases.

I was also thinking about making them required, but providing defaults: 200 status code, empty headers. This has a large advantage: user code doesn't need to handle "response.headers is None" case, response.headers.get("content-type") can safely be called. At the same time, code to create responses remain simple. A downside is that you can't know if 200 is a real status code, or a status code is not available; the same for headers.

web_poet/page_inputs.py
BurnzZ: > you don't need to wrap responses in page objects anymore

That's a good point. Not having to do `page = web_poet.WebPage(response)` everytime is very convenient! I also see your point about `ReponseShortcutsMixin`.

Thanks for eloquently explaining this. Moving these selector methods here makes sense. 👍 

What do you think about structuring it like this?

``` python
class HttpResponse:
    url
    body
    status
    headers

class ResponseData(HttpResponse):

    def selector(self):
        ...

    def xpath(self):
        ...

    def css(self):
        ...
```

In this way, we still have access to the bare minimum attributes of `HttpResponse` that could represent simple stuff like API calls returning plaintext which wouldn't necessarily need the selectors functionality in this case.

Moreover, `HEAD` requests also have their responses without a **body**, so representing them with methods like `selector` wouldn't make sense.

Nonetheless, we'd still keep `ResponseData` as the default recommended container. `HttpResponse` is also available on its own to be open to obscure use cases like the ones mentioned above.

kmike: > What do you think about structuring it like this?

Not sure about this. What would be an advantage of having a bare data container, and a subclass with methods? User can always use a subclass with methods instead of a barebone class, so why a separate class is needed? 

I was thinking about a big backwards-incompatible change: removing ResponseData container completely. 

> Moreover, HEAD requests also have their responses without a body, so representing them with methods like selector wouldn't make sense.

Selectors won't work in this case, but they won't work in many other cases. For example, on json responses .json method would work, while .css won't. On text responses none of these methods would work, but response.text would be useful, as it'd decode the content properly. There can be GET requests which return a response without body (which is the same as empty body, right?).

web_poet/page_inputs.py
kmike: What do you think about option 4?

```py
status: Optional[int] = attrs.field(default=None)
headers: HttpResponseHeaders = attrs.field(factory=HttpResponseHeaders)
```

I.e. allow status to be None, but not headers - provide "fake" empty headers by default?

web_poet/page_inputs.py
BurnzZ: > but then realized that encoding detection is not needed (json.loads in stdlib could work on bytes, there is code in stdlib to handle this)

TIL about this. In that case, it makes sense to put this inside `HttpResponseBody`. 👍

web_poet/page_inputs.py
BurnzZ: Good idea. This is slightly better than option 1 as a header instance is already available. This prevents `response.headers.get("Content-Type")` from unintentionally breaking compared to being `None`.

web_poet/page_inputs.py
BurnzZ: > There can be GET requests which return a response without body (which is the same as empty body, right?).

You have a point there.

> What would be an advantage of having a bare data container, and a subclass with methods? User can always use a subclass with methods instead of a barebone class, so why a separate class is needed?

Aside from semantically separating the HTTP attributes against its selector/extractor counterparts there's not much strong advantages for them.

I was initially thinking of potentially being able to leverage bare HTTP data containers on its own for some other obscure use cases. But after re-reading our discussion, I realized that majority of the use cases that web-poet need to support is about extracting data from responses as opposed to storing them. 

With that, I think it makes sense to store the HTTP attributes and its selectosr in a single class as currently proposed. 👍

web_poet/page_inputs.py
kmike: Implemented it.

kmike:
Closing, since #30 is merged.

PR #30, opened 2022-03-18
HttpResponse instead of ResponseData

kmike:
This is https://github.com/scrapinghub/web-poet/pull/29 + https://github.com/scrapinghub/web-poet/pull/24, but sent against master branch, to ensure merging this doesn't depend on anything else.

TODO:

* [x] fix from-ground-up tutorial
* [x] double-check docs
* [x] tests for HttpResponse.css, xpath and selector
* [x] tests for HttpResponse.json
* [x] tests for HttpResponse.text and HttpResponse.encoding

Out o scope for this PR:
* implement HttpResponse.urljoin (can be done later)
* implement HttpResponse.base_url (can be done later)
* further clean-up of mixing (can be done later)
* clean-up of WebPage class?
* implement BrowserHtml (though we should do it before the release, as it might change code significantly)
* implement HttpRequestHeaders (though we should do it before the release, as it might change code significantly)

### DIFF
--- a/web_poet/page_inputs.py
+++ b/web_poet/page_inputs.py
@@ -1,17 +1,187 @@
-import attr
+import json
+from typing import Optional, Dict, List, TypeVar, Type
+
+import attrs
+from multidict import CIMultiDict
+import parsel
+from w3lib.encoding import (
+    html_to_unicode,
+    html_body_declared_encoding,
+    resolve_encoding,
+    http_content_type_encoding
+)
+
+from .utils import memoizemethod_noargs
+
+T_headers = TypeVar("T_headers", bound="HttpResponseHeaders")
 
 
-@attr.s(auto_attribs=True)
-class ResponseData:
-    """A container for URL and HTML content of a response, downloaded
-    directly using an HTTP client.
+class HttpResponseBody(bytes):
+    """A container for holding the raw HTTP response body in bytes format."""
 
-    ``url`` should be an URL of the response (after all redirects),
-    not an URL of the request, if possible.
+    def declared_encoding(self) -> Optional[str]:
+        """ Return the encoding specified in meta tags in the html body,
+        or ``None`` if no suitable encoding was found """
+        return html_body_declared_encoding(self)
 
-    ``html`` should be content of the HTTP body, converted to unicode
-    using the detected encoding of the response, preferably according
-    to the web browser rules (respecting Content-Type header, etc.)
+    def json(self):
+        """
+        Deserialize a JSON document to a Python object.
+        """
+        return json.loads(self)
+
+
+class HttpResponseHeaders(CIMultiDict):
+    """A container for holding the HTTP response headers.
+
+    It's able to accept instantiation via an Iterable of Tuples:
+
+    >>> pairs = [("Content-Encoding", "gzip"), ("content-length", "648")]
+    >>> HttpResponseHeaders(pairs)
+    <HttpResponseHeaders('Content-Encoding': 'gzip', 'content-length': '648')>
+
+    It's also accepts a mapping of key-value pairs as well:
+
+    >>> pairs = {"Content-Encoding": "gzip", "content-length": "648"}
+    >>> headers = HttpResponseHeaders(pairs)
+    >>> headers
+    <HttpResponseHeaders('Content-Encoding': 'gzip', 'content-length': '648')>
+
+    Note that this also supports case insensitive header-key lookups:
+
+    >>> headers.get("content-encoding")
+    'gzip'
+    >>> headers.get("Content-Length")
+    '648'
+
+    These are just a few of the functionalities it inherits from
+    :class:`multidict.CIMultiDict`. For more info on its other features, read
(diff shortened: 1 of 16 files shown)


web_poet/page_inputs.py
```diff
@@ -1,17 +1,187 @@
...
+            or self._headers_declared_encoding()
+            or self._body_declared_encoding()
+            or self._body_inferred_encoding()
+        )
+
+    # XXX: see https://github.com/python/mypy/issues/1362
+    @property   # type: ignore
+    @memoizemethod_noargs
+    def selector(self) -> parsel.Selector:
+        # XXX: should we pass base_url=self.url, as Scrapy does?
```
BurnzZ, line 148: I'm afraid I don't quite follow this comment. Does it mean for a separate method like `urljoin` as opposed to being inside the `selector`?

web_poet/page_inputs.py
```diff
@@ -1,17 +1,187 @@
...
+        body_encoding, text = html_to_unicode(
+            content_type,
+            self.body,
+            auto_detect_fun=self._auto_detect_fun,
+            default_encoding=self._DEFAULT_ENCODING
+        )
+        self._cached_text = text
+        return body_encoding
+
+    def _auto_detect_fun(self, text):
```
BurnzZ, line 181: We could rename `text` as `body` and annotate it as `body: bytes` to be more clear.

BurnzZ (commented):
Hi @kmike , aside from these initial comments, I also took the liberty of making some changes as per the list in the **TODO** section. You can find the PR here: https://github.com/scrapinghub/web-poet/pull/31

web_poet/page_inputs.py
```diff
@@ -1,17 +1,187 @@
...
+
+    def xpath(self, query, **kwargs):
+        return self.selector.xpath(query, **kwargs)
+
+    def css(self, query):
+        return self.selector.css(query)
+
+    @memoizemethod_noargs
+    def json(self):
+        return self.body.json()
```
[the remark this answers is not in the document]
kmike: Hey @BurnzZ! 

> I think we can move the memoization into HttpResponseBody itself.

This is what I did initially, but it doesn't want with the decorator we have. It uses weakrefs to store the cache, but bytes objects (and HttpResponseBody) are not weakref-able. We can implement the cache in a more manual way though.

> I think we can also make this a @property as well for ease of use.

requests library went the opposite direction: responses used to have ".json" property, but then it was changed to methods; it was a big backwards incompatible change. I think having a method is a bit safer, as it allows to add arguments later. That said, it makes it inconsistent e.g. with .text and .selector attributes.

web_poet/page_inputs.py
```diff
@@ -1,17 +1,187 @@
...
+        # while detecting the encoding
+        encoding = self.encoding
+        if self._cached_text is None:
+            fake_content_type_header = f'charset={encoding}'
+            encoding, text = html_to_unicode(fake_content_type_header, self.body)
+            self._cached_text = text
+        return self._cached_text
+
+    @property
+    def encoding(self):
```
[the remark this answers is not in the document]
kmike: All operations inside are memoized, so it I left the .encoding non-memoized

web_poet/page_inputs.py
kmike: Yeah, sorry, it is unclear indeed. Scrapy does this: https://github.com/scrapy/scrapy/blob/9a28eb0bad1acf986d997905a410058f77911b7c/scrapy/selector/unified.py#L79 - setting base_url to response.url in Selector. That's one of the differences between Scrapy's Selector and parsel's Selector. But I'm not sure if that's the right thing to do (maybe it is).

web_poet/page_inputs.py
kmike: nice catch!

web_poet/page_inputs.py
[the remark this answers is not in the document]
BurnzZ: > This is what I did initially, but it doesn't want with the decorator we have. It uses weakrefs to store the cache, but bytes objects (and HttpResponseBody) are not weakref-able. We can implement the cache in a more manual way though.

Got it. Yeah, I think we can address the current cache approach to accommodate this later on in a separate PR.

> I think having a method is a bit safer, as it allows to add arguments later. 

This is a good point, as some interest of having **jmespath** query expressions might come into play later on.

web_poet/page_inputs.py
BurnzZ: I see, thanks for explaining @kmike .

It might be worth adding the **base_url** with the notion of encouraging users that needs the relative lookups provided by the base_url. However, I'm not quite sure how common is this use case. Perhaps we can add this later on when user demand for it builds up.

BurnzZ (approved):
Looking good @kmike! 💯 I've also made another review in the docs and everything seems to be in order with respect to the changes on this PR. 👍

kmike:
Thanks for the review and fixes @BurnzZ! Merging it.

PR #33, opened 2022-03-28
add from_bytes_dict alternative constructor for HttpResponseHeaders

BurnzZ:
From the recent enhancement in https://github.com/scrapinghub/web-poet/pull/30, it would seem that we'll need to support insantiating the `HttpResponseHeaders` from raw bytes values. In particular, frameworks like Scrapy uses header values in bytes format.

By default, https://github.com/aio-libs/multidict only supports strings and not bytes. Thus, an alternative constructor is proposed for ease of use.

### DIFF
--- a/web_poet/page_inputs.py
+++ b/web_poet/page_inputs.py
@@ -76,7 +76,7 @@ class HttpResponseHeaders(CIMultiDict):
         return cls([(pair["name"], pair["value"]) for pair in arg])
 
     @classmethod
-    def from_bytes(
+    def from_bytes_dict(
         cls: Type[T_headers], arg: BytesDict, encoding: str = "utf-8"
     ) -> T_headers:
         """An alternative constructor for instantiation where the header-value
@@ -93,7 +93,7 @@ class HttpResponseHeaders(CIMultiDict):
         ...     b"Content-Type": [b"text/html"],
         ...     b"content-length": b"648",
         ... }
-        >>> headers = HttpResponseHeaders.from_bytes(raw_values)
+        >>> headers = HttpResponseHeaders.from_bytes_dict(raw_values)
         >>> headers
         <HttpResponseHeaders('Content-Encoding': 'gzip', 'Content-Encoding': 'br', 'Content-Type': 'text/html', 'content-length': '648')>
         """
--- a/web_poet/page_inputs.py
+++ b/web_poet/page_inputs.py
@@ -1,5 +1,5 @@
 import json
-from typing import Optional, Dict, List, TypeVar, Type
+from typing import Optional, Dict, List, TypeVar, Type, Union
 
 import attrs
 from multidict import CIMultiDict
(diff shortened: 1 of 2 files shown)


web_poet/page_inputs.py
```diff
@@ -74,6 +75,45 @@ def from_name_value_pairs(cls: Type[T_headers], arg: List[Dict]) -> T_headers:
...
+        cls: Type[T_headers], arg: BytesDict, encoding: str = "utf-8"
+    ) -> T_headers:
+        """An alternative constructor for instantiation where the header-value
+        pairs are in raw bytes form.
+
+        This supports multiple header values in the form of ``List[bytes]``
+        alongside a plain ``bytes`` value.
+
+        By default, it converts the ``bytes`` value using "utf-8". However, this
+        can easily be overridden using the ``encoding`` parameter.
```
BurnzZ, line 89: This `utf-8` default value was based on how Scrapy was using it as the default for its Headers.

web_poet/page_inputs.py
```diff
@@ -74,6 +75,45 @@ def from_name_value_pairs(cls: Type[T_headers], arg: List[Dict]) -> T_headers:
         """
         return cls([(pair["name"], pair["value"]) for pair in arg])
 
+    @classmethod
+    def from_bytes(
```
kmike, line 79: The name may be a bit misleading, one may think this creates Headers from HTTP response bytes which represent the headers. Maybe something like `from_bytes_dict`? Or maybe even more specific, `from_scrapy_headers`?

BurnzZ: Good point! Renaming it to `from_scrapy_headers` seems to be too specific. Other non-scrapy libs could have the same setup. Having it as `from_bytes_dict` makes it more clear and semantically more meaningful at the same time. Updated this on https://github.com/scrapinghub/web-poet/pull/33/commits/eb184270ff8db4bccbb7e3f7317c0192a0fb6a29

web_poet/page_inputs.py
```diff
@@ -74,6 +75,45 @@ def from_name_value_pairs(cls: Type[T_headers], arg: List[Dict]) -> T_headers:
...
+        >>> headers = HttpResponseHeaders.from_bytes_dict(raw_values)
+        >>> headers
+        <HttpResponseHeaders('Content-Encoding': 'gzip', 'Content-Encoding': 'br', 'Content-Type': 'text/html', 'content-length': '648')>
+        """
+
+        def _norm(data):
+            if isinstance(data, str):
+                return data
+            elif isinstance(data, bytes):
+                return data.decode(encoding)
```
kmike, line 105: I think it should raise an exception if data is not bytes

BurnzZ: Good catch! Adressed this in https://github.com/scrapinghub/web-poet/pull/33/commits/ddb7d203e76d2399fa108fcf43f08890ed731991 alongside the ability to handle `Tuple` alongside `List`.

kmike:
Thanks @BurnzZ!

PR #37, opened 2022-04-25
reorganize page_inputs.py as a submodule; move HttpClient to it

BurnzZ:
Stemming off from https://github.com/scrapinghub/web-poet/pull/22#discussion_r852036035.


web_poet/page_inputs/client.py
```diff
@@ -0,0 +1,153 @@
...
+)
+
+logger = logging.getLogger(__name__)
+
+_StrMapping = Dict[str, str]
+_Headers = Union[_StrMapping, HttpRequestHeaders]
+_Body = Union[bytes, HttpRequestBody]
+
+
+async def _perform_request(request: HttpRequest) -> HttpResponse:
```
BurnzZ, line 31: This needs to be inside this module, as moving it outside would cause a _"circular import problem"_.

It's because `_perform_request()` needs `HttpRequest` and `HttpResponse` from **page_inputs** for its annotations while `HttpClient` _(which is now inside the **page_inputs** subpackage)_ needs `_perform_request()`.

kmike: I think that can be possible to work around: 

1) add `from __future__ import annotations` to the module which contains _perform_request
2) add `if if typing.TYPE_CHECKING: from web_poet... import HttpRequest, HttpResponse`

See https://mypy.readthedocs.io/en/stable/runtime_troubles.html#import-cycles and https://peps.python.org/pep-0563/#runtime-annotation-resolution-and-type-checking

On a first sight, the having this function in client.py looks fine though :) Do you see a better location for this function?

BurnzZ: Hi @kmike , thanks for the suggestion! I was about to use your workaround but then simply importing `HttpRequest` and `HttpResponse` in requests.py is now working without the _"circular import problem"._ I think the problem before was a bad `./tox` cache in my local machine. 😅 

> On a first sight, having this function in client.py looks fine though :) Do you see a better location for this function?

It would seem that the `_perform_request()` function would be housed perfectly in `web_poet/requests.py`. This makes `web_poet/page_inputs/client.py` cleaner, containing code which emphasizes only on what users could use as page inputs.

Updated this in https://github.com/scrapinghub/web-poet/pull/37/commits/a4f1dcc019809c3da949944ec59d1c19e6a4de32.

kmike:
Thanks @BurnzZ!

PR #38, opened 2022-04-25
Raise exceptions when receiving 400-5xx responses

BurnzZ:
Built on top of https://github.com/scrapinghub/web-poet/pull/22.

**TODO:**
- [x] Tutorial Docs
- [x] Include `execute()` and `batch_execute()`
- [x] have the ability to allow all status codes
- [x] Explore both inclusion and exclusion list for status codes
  - opting for the default errors on `400-5xx` and having an exclusion list using `allow_status` param


docs/advanced/additional-requests.rst
```diff
@@ -750,9 +750,12 @@ doc section) explicitly raises when implementing it for **web-poet** should be
...
 errors`, etc should be replaced by :class:`web_poet.exceptions.http.HttpRequestError`
-by raising it explicitly. For responses that are successful but don't have a ``200``
-**status code**, this exception shouldn't be raised at all. Instead, the
-:class:`~.HttpResponse` should simply reflect the response contents as is.
+by raising it explicitly.
+
+For responses that are not really errors like in the 100-3xx status code range,
+this exception shouldn't be raised at all. However, for responses with status
+codes in the 400-5xx range, the implementing framework should properly raise
+the exception.
```
kmike, line 758: Hey! What do you think about a slightly different approach?

1. State that frameworks must not raise exceptions for 400-5xx status code. They should return an HttpResponse, with a proper status code set. I'm actually not sure how allow_status works in this PR, if a framework raises an exception for 400-5xx codes.
2. Frameworks must raise HttpRequestError for connections errors, etc. I'm not sure about timeouts though; it's fine to make frameworks handle it for now, but we'd need to think about it further.
3. We should define framework behavior for 3xx status codes, for redirects. For example, we can say that frameworks must follow redirects.
4. Probably it's a right time to add another exception class, to be able to distinguish errors without a response, and errors where response is present. We can even have a response as an attribute of this exception: because of (1), this information would be available.

BurnzZ: Hi @kmike , let me know what do you think about the points below?

1. Sounds good. It's currently the expectation on the `master` branch. So not much changes here aside from perhaps improving the docs to reflect on it more clearly.

    - I was also thinking of perhaps having a similar interface to [aiohttp](https://docs.aiohttp.org/en/stable/client_reference.html?highlight=errors#aiohttp.ClientResponse.raise_for_status) and [request](https://2.python-requests.org/en/master/api/#requests.Response.raise_for_status) `raise_for_status()` method could cover 400-5xx errors. This shifts the focus of having the errors being the default into an intentional one dictated by the developer.

2. We'll need to improve the docs on this regard regarding raising `HttpRequestError` for connection errors. For timeouts, we can **create** a new exception subclassing from `HttpRequestError`.

3. Got it, we'll need to improve the docs for this one. Perhaps **create** an entirely new section for a framework's checklist.

4. I think we can use the current `HttpRequestError` for errors without a response and **create** a new `HttpReponseError` for errors with responses. In addition, we could also **create** a parent class for both of them, perhaps calling it `HttpError`.

We can create a new PR for this as it has a different approach due to **(1)**.

kmike: Hey @BurnzZ! Regarding (1) - I was thinking that frameworks would not be raising these exceptions, but web-poet would.

BurnzZ: I see, I got that wrong. Thanks for the clarification! Will edit my points above.

kmike: Regarding exceptions: +1 to have 3 exceptions - a base exception, and 2 exceptions for these conditions.

kmike: As for timeouts, the main issue is who sets the timeout value - is it a framework, or is it something which can be controlled in PO? I think so far we've been side-stepping it, and letting framework to manage it. 

But it means that  using a specific timeout for a specific request, e.g. 
* decreasing it for requests which should be fast, or 
* increasing for requests which are expected to take long, or 
* setting a timeout if a library doesn't have a timeout

can't be done. I don't think this is something to solve in this PR, but it's something to think about.

BurnzZ: Hi @kmike ! I've reorganized the docs to better document all of these things. The main change is to properly distinguish the docs for Page Object developers vs developers that would implement a framework for **web-poet**. https://github.com/scrapinghub/web-poet/pull/38/commits/3ebb2973aaaea79bff7b9f4f033628233b9d8b87.

The new base class exception named `HttpError` was also introduced. https://github.com/scrapinghub/web-poet/pull/38/commits/470bc6f417431168d0138762dbf76d12801163c4.

Kindly let me know if there are other specific things we need to explicitly document here that were missed out.

----

Regarding the timeouts, yeah, that's indeed tricky. It also adds some additional complexity to Page Objects, since aside from knowing how to construct the requests, it would also need to consider _(and establish an expectation)_ how long it would take the response to arrive.

There would be some factors that the Page Objects might not have access to, like the number of requests made to the same domain in the past minute _(which could timeout due to rate limiting)_ or the shifting tide of non-`200` responses for a given domain _(where the site could throttle down requests as well)_. Letting a Page Object decide on timeouts based on limited knowledge is difficult.

I think at least for the short-term, we can let the framework handle it.

tests/test_requests.py
```diff
@@ -81,6 +81,62 @@ async def test_http_client_single_requests(async_mock):
...
+    assert response.status == 500
+
+    # As well as strings
+    await method(url_or_request, allow_status="500")
+    await method(url_or_request, allow_status=["500", "503"])
+
+    with pytest.raises(HttpResponseError) as excinfo:
+        await method(url_or_request)
+    assert isinstance(excinfo.value.request, HttpRequest)
+    assert isinstance(excinfo.value.response, HttpResponse)
```
BurnzZ, line 122: @kmike could you refresh my memory on why the `HttpRequest` and `HttpResponse` classes were set using `@attr.define(..., eq=False)`? [link to code](https://github.com/scrapinghub/web-poet/blob/master/web_poet/page_inputs.py#L148)

Setting `eq=True` could allow easy comparisons here in the test.

docs/advanced/additional-requests.rst
```diff
@@ -695,30 +715,39 @@ For this example, let's improve the code snippet from the previous subsection na
 
 Handling exceptions using :meth:`~.HttpClient.batch_execute` remains largely the same.
 However, the main difference is that you might be wasting perfectly good responses just
-because a single request from the batch ruined it.
+because a single request from the batch ruined it. Notice that we're using the base
+exception class of :class:`web_poet.exceptions.http.HttpError` to account for any
```
kmike, line 719: we can also add a subclass like BatchHttpError for this case

docs/advanced/additional-requests.rst
```diff
@@ -919,3 +889,118 @@ an :class:`~.HttpClient` instance:
...
+in order to handle and represent generic HTTP operations. The only time that
+it won't be returning :class:`~.HttpResponse` would be when it's raising exceptions
+(see :ref:`framework-exception-handling`).
+
+The Downloader should be able to properly resolve **redirections** except when
+the method is ``HEAD``. This means that the :class:`~.HttpResponse` that it'll
+be rendering is already the end of the redirection trail.
+
+Laslty, the Downloader should also be able to fully support :mod:`asyncio` to
+enable developers to perform additional requests asynchronously.
```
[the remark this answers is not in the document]
BurnzZ: Good catch! I was meaning to write `async/await` here but written down `asyncio` for some reason. Updated in https://github.com/scrapinghub/web-poet/pull/38/commits/38a53c9f823478f223515527fc47e5f99efd5827

docs/advanced/additional-requests.rst
kmike: > or would BatchHttpError only be raised when return_exceptions=False.

I was thinking it should work like this - only raise it when return_exceptions is False.

But actually I think I misread the code, and BatchHttpError may not be a good idea :) 

So, currently we're raising HttpRequestError or HttpResponseError, and catching HttpError, to account for both situations. This sounds like a right behavior to be, better than raising BatchHttpError and catching BatchHttpError.

docs/advanced/additional-requests.rst
```diff
@@ -919,3 +889,118 @@ an :class:`~.HttpClient` instance:
...
+-------------------
+
+The Downloader should be able to accept an instance of :class:`~.HttpRequest`
+as the input and return an instance of :class:`~.HttpResponse`. This is important
+in order to handle and represent generic HTTP operations. The only time that
+it won't be returning :class:`~.HttpResponse` would be when it's raising exceptions
+(see :ref:`framework-exception-handling`).
+
+The Downloader should be able to properly resolve **redirections** except when
+the method is ``HEAD``. This means that the :class:`~.HttpResponse` that it'll
```
kmike, line 903: > except when the method is ``HEAD``

That's an interesting behavior. It seems right to me. I wonder how easy would it be to implement it though. For example, Scrapy seems to redirect in case of HEAD: https://github.com/scrapy/scrapy/blob/b2afcbfe2bf090827540d072866bef0d1ab3a3e8/scrapy/downloadermiddlewares/redirect.py#L102

docs/advanced/additional-requests.rst
```diff
@@ -919,3 +889,118 @@ an :class:`~.HttpClient` instance:
...
+For responses that are not really errors like in the ``100-3xx`` status code range,
+no exception should be raised at all. For responses with status codes in 
+the ``400-5xx`` range, **web-poet** raises the :class:`web_poet.exceptions.http.HttpResponseError`
+exception.
+
+From this distinction, the framework shouldn't raise :class:`web_poet.exceptions.http.HttpResponseError`
+on its own at all, since the :class:`~.HttpClient` already handles that. However,
+the implementing framework could modify which status codes to allow (*meaning no
+exceptions raised*) using the ``allow_status`` parameter in the
+:class:`~.HttpClient` methods.
```
kmike, line 1006: I'm not sure I understand this part. Shouldn't it be like this:

1. Framework never raises HttpResponseError. Framework also has nothing to do with allow_status, it doesn't need to know about it.
2. web-poet takes a response, and can raise HttpResponseError, depending on allow_status value.

web_poet/exceptions/http.py
```diff
@@ -6,14 +6,59 @@
...
+
+    For more specific errors, it would be better to use :class:`.HttpRequestError`
+    and :class:`.HttpResponseError`.
+
+    :param request: The :class:`~.HttpRequest` instance that was used.
+    :type request: HttpRequest
+    :param response: The :class:`~.HttpResponse` instance that was received, if
+        any. Note that this wouldn't exist if the problem ocurred when executing
+        the :class:`~.HttpRequest`.
+    :type response: HttpResponse
```
kmike, line 26: I wonder if response should be an attribute of HttpResponseError, not an attribute of a base class. What's a use case for having it here, and in HttpRequestError?

web_poet/exceptions/http.py
```diff
@@ -6,14 +6,59 @@
...
+
+    For responses that are in the status code ``100-3xx range``, this exception
+    shouldn't be raised at all. However, for responses in the ``400-5xx``, this
+    will be raised by **web-poet**.
+
+    .. note::
+
+        Frameworks implementing **web-poet** should **NOT** raise this themselves
+        but rather, rely on the ``allow_status`` parameter found in the methods
+        of :class:`~.HttpClient`.
```
kmike, line 61:
```suggestion
        Frameworks implementing **web-poet** should **NOT** raise this  exception.
        
        This exception is raised by web-poet itself, based on ``allow_status`` 
        parameter found in the methods of :class:`~.HttpClient`.
```

docs/advanced/additional-requests.rst
```diff
@@ -919,3 +889,118 @@ an :class:`~.HttpClient` instance:
 From the code sample above, we can see that every time an :class:`~.HttpClient`
 is created for Page Objects needing an ``http_client``, the specific **request
 implementation** from a given framework is injected to it.
+
+Downloader Behavior
+-------------------
+
+The Downloader should be able to accept an instance of :class:`~.HttpRequest`
```
kmike, line 896: It's a general comment about framework docs. I wonder if we should use stronger language, similar to how it's used in RFCs: https://www.ietf.org/rfc/rfc2119.txt. 

So, Downloader MUST accept an instance of :class:`~.HttpRequest`, not "should accept" it, and it MUST either return an instance of :class:`~.HttpResponse` or raise  a subclass of HttpRequestError.

What do you think about it @BurnzZ ?

web_poet/requests.py
```diff
@@ -82,18 +84,56 @@ class HttpClient:
...
+            or response.status is None  # allows serialized responses from tests
+            or response.status < 400
+            or str(response.status) in allow_status_normalized
+        ):
+            return
+
+        if 400 <= response.status < 500:
+            kind = "Client"
+        elif 500 <= response.status < 600:
+            kind = "Server"
```
kmike, line 110: If that's desirable, you can get error description from https://docs.python.org/3/library/http.html, instead of having a logic here

web_poet/requests.py
```diff
@@ -82,18 +84,56 @@ class HttpClient:
...
     ) -> HttpResponse:
-        """This is a shortcut for creating a :class:`~.HttpRequest` instance and executing
-        that request.
+        """This is a shortcut for creating a :class:`~.HttpRequest` instance and
+        executing that request.
 
-        A :class:`~.HttpResponse` instance should then be returned.
+        A :class:`~.HttpResponse` instance should then be returned for successful
+        responses in the 100-3xx status code range. Otherwise, an exception of
+        type :class:`web_poet.exceptions.http.HttpResponseError` will be raised.
```
kmike, line 129: I think it reads like this: successful response with 100-3xx status code? HttpResponse is returned. Otherwise? HttpResponseError. 

It's not entirely correct, because HttpRequestError can also be raised (connection error is not a "successful response with 100-3xx status code").

docs/advanced/additional-requests.rst
BurnzZ: Thanks for raising this @kmike ! I didn't that this was the default behavior of Scrapy. 

I think the easiest route would be for **scrapy-poet** to set `meta={'dont_redirect'=True}` when it sees a `HEAD` request method. Added a note in the TODO section of https://github.com/scrapinghub/scrapy-poet/pull/62 so we wouldn't forget.

docs/advanced/additional-requests.rst
BurnzZ: I've corrected this in https://github.com/scrapinghub/web-poet/pull/38/commits/c5244b9fe04201529f3d78733e7cdeffcf7a4d54. Thanks!

web_poet/exceptions/http.py
BurnzZ: Thanks for raising this! I've updated this on https://github.com/scrapinghub/web-poet/pull/38/commits/a209b4d83932261026937c9f27122d987d97f550

docs/advanced/additional-requests.rst
BurnzZ: Good call! Updated the docs in https://github.com/scrapinghub/web-poet/pull/38/commits/b889fb974f935fe0c27902cfdc9cbb2b186fad0f to follow RFC 2119's guidelines.

web_poet/requests.py
BurnzZ: A good suggestion! Used this on https://github.com/scrapinghub/web-poet/pull/38/commits/1401a514106514d2910bcb6dc15d4d5909d5b453.

web_poet/requests.py
BurnzZ: A good catch! Updated the language in https://github.com/scrapinghub/web-poet/pull/38/commits/0eefdf872f4aa9cbf5ad9c52a29626dbbe49e666 to explicitly mention `HttpRequestError` as well.

docs/advanced/additional-requests.rst
```diff
@@ -304,7 +308,7 @@ Executing a HttpRequest instance
             return item
 
 As the example suggests, we're performing an additional request that allows us
-to extract more images in a product page that might not be otherwise be possible.
+to extract more images in a product page that MAY NOT be otherwise be possible.
```
kmike, line 311: I think in cases like this we shouldn't follow RFC terms. Here we're just explaining the example. I think RFC applies where we explain the rules user needs to follow to implement something, and where we need to be really specific on what's optional and what's not optional.

docs/advanced/additional-requests.rst
```diff
@@ -919,3 +893,115 @@ an :class:`~.HttpClient` instance:
...
+
+The Downloader MUST be able to accept an instance of :class:`~.HttpRequest`
+as the input and return an instance of :class:`~.HttpResponse`. This is important
+in order to handle and represent generic HTTP operations. The only time that
+it won't be returning :class:`~.HttpResponse` would be when it's raising exceptions
+(see :ref:`framework-exception-handling`).
+
+The Downloader MUST be able to properly resolve **redirections** except when
+the method is ``HEAD``. This means that the :class:`~.HttpResponse` that it'll
+be rendering is already the end of the redirection trail.
```
kmike, line 908:
```suggestion
The Downloader MUST resolve Location-based **redirections** when the HTTP 
method is not ``HEAD``. In other words, for non-``HEAD`` requests the 
returned :class:`~.HttpResponse` must be the final response, after all redirects. 
For ``HEAD`` requests redirects MUST NOT be resolved. 
```

docs/advanced/additional-requests.rst
```diff
@@ -919,3 +893,115 @@ an :class:`~.HttpClient` instance:
...
+
+Lastly, the Downloader MUST also be able to fully support the ``async/await``
+syntax in order to enable developers to perform additional requests asynchronously.
```
kmike, line 911:
```suggestion
Lastly, the Downloader MUST support the ``async/await``
syntax in order to enable developers to perform additional requests asynchronously.
```

docs/advanced/additional-requests.rst
BurnzZ: Reverted some of the updated terms in https://github.com/scrapinghub/web-poet/pull/38/commits/29aa8cc6d81fec598451e8493e91045661e9d9b8.

web_poet/exceptions/http.py
```diff
@@ -6,14 +6,61 @@
...
+    since it could denote problems either in the HTTP Request or Response.
+
+    For more specific errors, it would be better to use :class:`.HttpRequestError`
+    and :class:`.HttpResponseError`.
+
+    :param request: The :class:`~.HttpRequest` instance that was used.
+    :type request: HttpRequest
+    """
+
+    def __init__(self, *args, request: HttpRequest = None, **kwargs):
```
kmike, line 25: hey! What do you think about simplifying the parameters - why are *args and **kwargs needed here?

web_poet/page_inputs/client.py
```diff
@@ -45,18 +51,55 @@ class HttpClient:
...
+        if (
+            allow_all_status
+            or response.status is None  # allows serialized responses from tests
+            or response.status < 400
+            or str(response.status) in allow_status_normalized
+        ):
+            return
+
+        status = HTTPStatus(response.status)
+        msg = f"{response.status} {status.name} response for {response.url}"
```
kmike, line 75: just reaslised that this code can be moved to HttpResponseError

kmike (approved):
Looks good @BurnzZ!

web_poet/exceptions/http.py
BurnzZ: Good call. Addressed this in https://github.com/scrapinghub/web-poet/pull/38/commits/8b01cf7590f904047dba1851fb4d0ff18a43da69.

web_poet/page_inputs/client.py
BurnzZ: Good idea. Although I think we could make the default message more generic than this one. There are cases to raise `HttpResponseError` that are not specific to the status code and thus, exceptions messages like this would be invalid. For example, corrupt/malformed/truncated responses even on 200 status.

Updated this on https://github.com/scrapinghub/web-poet/pull/38/commits/43bb5d9998c0e0ca51141f34a5c1b57bf2c6c628.

kmike:
Thanks @BurnzZ!

PR #42, opened 2022-05-21
RequestURL and ResponseURL

kmike:
This is a basic version of URL dependency; it'd allow Page Objects to request just the URL, without the response.
RequestURL is especially useful, as it allows Page Objects to skip the initial download.

My concerns about this PR: I really wanted to have some "smart" object for URL, not just a str subclass. For this, I considered 3 options:

1. https://github.com/aio-libs/yarl (a top contender). 
2. https://github.com/python-hyper/hyperlink
3. roll out a new implementation, based on w3lib

For yarl the issue is that it declares yarl.URL as "final", and also prohibits subclassing at runtime. So it's not possible to have a "marker" subclass like RequestURL. To overcome this, I tried using https://docs.python.org/3/library/typing.html#newtype; it works at runtime, but I was unable to make typing work well for attr converters; e.g. creating HttpRequest with anything but RequestURL instance was problematic.

Next, I looked into hyperlink, but the API is not perfect. For example, they provide url.click method, which is essentially a re-implementation of urljoin. I think the method name is misleading, especially in web-poet context, because to click properly you need to take base url in account, not just response url.

Also, if I'm not mistaken, these libraries don't handle all the corner cases (e.g. a case when path encoding is different from query encoding, which could be the case for non-utf8 pages).

The third option is also something I don't like :) I'd much prefer to use something existing, something like yarl, and contribute to it if needed. But it is still an option to roll out a new impementation. In this case, I think we should create w3lib.URL object, not an object in web-poet. w3lib.URL could expose functions from w3lib.url in a nice way; it doesn't have to be very complete (minimal is fine), but it should expose features needed for web scraping. I think that it'd be fine even if this object only exposes some methods to work with query parameters, and nothing more; it'd be enough for a first version.

This PR contains a very basic implementation. It may be sufficient for web-poet needs, but I'm worried about backwards compatibility if we move to an URL object in future.

Thoughts?

### DIFF
--- a/tests/test_page_inputs.py
+++ b/tests/test_page_inputs.py
@@ -6,7 +6,6 @@ import requests
 import parsel
 
 from web_poet import RequestUrl, ResponseUrl
-from web_poet._base import _Url
 from web_poet.page_inputs import (
     HttpRequest,
     HttpResponse,
@@ -445,31 +444,3 @@ def test_browser_html():
     assert html.xpath("//p/text()").getall() == ["Hello, ", "world!"]
     assert html.css("p::text").getall() == ["Hello, ", "world!"]
     assert isinstance(html.selector, parsel.Selector)
-
-
-def test_url_base_class():
-    url_str = "http://example.com"
-    url = _Url(url_str)
-    assert str(url) == url_str
-    assert repr(url) == "_Url('http://example.com')"
-
-    with pytest.raises(TypeError):
-        _Url(123)
-
-
-def test_url_subclass():
-    url_str = "http://example.com"
-
-    class MyUrl(_Url):
-        pass
-
-    class MyUrl2(_Url):
-        pass
-
-    url = MyUrl(url_str)
-    assert str(url) == url_str
-    assert url._url == url_str
-    assert repr(url) == "MyUrl('http://example.com')"
-
-    url2 = MyUrl2(url)
-    assert str(url2) == str(url)
--- a/tests/test_page_inputs.py
+++ b/tests/test_page_inputs.py
@@ -3,8 +3,10 @@ import json
 import aiohttp.web_response
 import pytest
 import requests
-
 import parsel
+
+from web_poet import RequestUrl, ResponseUrl
+from web_poet._base import _Url
 from web_poet.page_inputs import (
     HttpRequest,
     HttpResponse,
@@ -72,7 +74,7 @@ def test_http_defaults(cls, body_cls):
     http_body = body_cls(b"content")
 
     obj = cls("url", body=http_body)
-    assert obj.url == "url"
+    assert str(obj.url) == "url"
     assert obj.body == b"content"
     assert not obj.headers
     assert obj.headers.get("user-agent") is None
@@ -164,7 +166,8 @@ def test_http_headers_init_dict(cls, headers_cls):
 
 def test_http_request_init_minimal():
     req = HttpRequest("url")
-    assert req.url == "url"
+    assert str(req.url) == "url"
+    assert isinstance(req.url, RequestUrl)
     assert req.method == "GET"
     assert isinstance(req.method, str)
     assert not req.headers
@@ -189,10 +192,18 @@ def test_http_request_init_full():
     http_body = HttpRequestBody(b"body")
     req_2 = HttpRequest("url", method="POST", headers=http_headers, body=http_body)
 
-    assert req_1.url == req_2.url
+    assert str(req_1.url) == str(req_2.url)
     assert req_1.method == req_2.method
     assert req_1.headers == req_2.headers
     assert req_1.body == req_2.body
+
+
+def test_http_request_init_with_response_url():
+    resp = HttpResponse("url", b"")
+    assert isinstance(resp.url, ResponseUrl)
+    req = HttpRequest(resp.url)
+    assert isinstance(req.url, RequestUrl)
+    assert str(req.url) == str(resp.url)
 
 
 def test_http_response_headers_from_bytes_dict():
@@ -434,3 +445,31 @@ def test_browser_html():
     assert html.xpath("//p/text()").getall() == ["Hello, ", "world!"]
     assert html.css("p::text").getall() == ["Hello, ", "world!"]
     assert isinstance(html.selector, parsel.Selector)
+
+
+def test_url_base_class():
+    url_str = "http://example.com"
+    url = _Url(url_str)
+    assert str(url) == url_str
+    assert repr(url) == "_Url('http://example.com')"
+
+    with pytest.raises(TypeError):
+        _Url(123)
+
+
+def test_url_subclass():
+    url_str = "http://example.com"
+
+    class MyUrl(_Url):
+        pass
+
+    class MyUrl2(_Url):
+        pass
+
+    url = MyUrl(url_str)
+    assert str(url) == url_str
+    assert url._url == url_str
+    assert repr(url) == "MyUrl('http://example.com')"
+
+    url2 = MyUrl2(url)
+    assert str(url2) == str(url)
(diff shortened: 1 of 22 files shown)


BurnzZ:
+1 on using `yarl` as a lot of developers that use Scrapy has at least interacted or used it in the past. This leads to better familiarity with the new Smart URL's API. It's also well-tested and has been battle-tested in many applications. This makes option 3 of creating a new implementation from scratch more time consuming and would have possibly more bugs.

>For yarl the issue is that it declares yarl.URL as "final", and also prohibits subclassing at runtime. So it's not possible to have a "marker" subclass like RequestURL.

We could do something like this:
```python
import yarl

# As suggested in https://github.com/aio-libs/yarl/issues/312
PatchedURL = yarl.URL
del PatchedURL.__init_subclass__

class RequestURL(PatchedURL):
    pass

class ResponseURL(PatchedURL):
    pass
```

kmike:
The issue with yarl workaround is that typing still doesn't work properly; yarl hardcodes various methods as returning and receiving yarl.URL. If I recall properly, this is what was breaking it usage in HttpResponse.

Gallaecio (approved):

Gallaecio:
> I'm worried about backwards compatibility if we move to an URL object in future.

I believe this is technically a valid concern, e.g. `URL("foo")[0]` works here and may not work in the future. Maybe we could instead use a custom class with an internal string, and only define what we expect not no change in a backward-incompatible way: `__init__` and `__str__`.

kmike:
> I believe this is technically a valid concern, e.g. URL("foo")[0] works here and may not work in the future. Maybe we could instead use a custom class with an internal string, and only define what we expect not no change in a backward-incompatible way: __init__ and __str__.

It's a bit worse than that: neither yarl.URL nor hyperlink.URL are string subclasses.

Gallaecio:
> It's a bit worse than that: neither yarl.URL nor hyperlink.URL are string subclasses.

Yes, I am suggesting we make it so that ours is not one either:

```python
class RequestUrl:
    def __init__(self, url: str):
        self._url = url
    def __str__(self):
        return self._url
```

That way `str(url)` works (I presume) the same as with yarl and hyperlink, and our interface is as minimal as possible (as opposed to exposing the whole `str` API), so that we can grow it as needed putting thought into every API change.

BurnzZ:
> I am suggesting we make it so that ours is not one either

What about wrapping `yarl.URL` into a class?

```python
class RequestURL:
    def __init__(self, url: str):
        self._url = yarl.URL(url)

    def __str__(self) -> str:
        return str(self._url)
```

This solves:
- yarl's the typing issue
- yarl's subclassing issue 
- > I'm worried about backwards compatibility if we move to an URL object in future.

We don't need to match every dunder method to the equivalent `yarl.URL` functionality if we won't need it. For example, if we find the `.path` property handy, then we could simply add it as:
```python
class RequestURL: 
    ... # same code as above

    @property
    def path(self) -> str:
        return self._url.path

url = RequestURL("https://example.com/category")
url.path  # /category
```

We're also free to define our own interface and not be tied directly to `yarl`. 

> Also, if I'm not mistaken, these libraries don't handle all the corner cases (e.g. a case when path encoding is different from query encoding, which could be the case for non-utf8 pages).

More specifically, we can solve this on our own.

kmike:
> Finally, I thought I had mentioned it already somewhere, so sorry if I did and I am repeating myself, but for consistency with existing classes the names should probably be Url and not URL. Which I find unpythonic, but I find consistency more important.

I was wondering if anyone would notice that :)

web_poet/page_inputs/client.py
```diff
@@ -72,12 +73,12 @@ def _handle_status(
             return
 
         status = HTTPStatus(response.status)
-        msg = f"{response.status} {status.name} response for {response.url}"
+        msg = f"{response.status} {status.name} response for {response.url!s}"
```
Gallaecio, line 76: I don’t think this change makes a difference, I believe f-strings cast their parameters to string by default.

> The only reason to use '!s' is if you want to specify a format specifier that applies to str, not to the type of the expression. — [PEP 498 – Literal String Interpolation](https://peps.python.org/pep-0498/)

Gallaecio (approved):

tests/test_requests.py
```diff
@@ -17,7 +17,7 @@ def async_mock():
     """Workaround since python 3.7 doesn't ship with asyncmock."""
 
     async def async_test(req):
-        return HttpResponse(req.url, body=b"")
+        return HttpResponse(str(req.url), body=b"")
```
BurnzZ, line 20: I don't think we should be casting this to `str` as the tests that rely on this mock are expecting `ResponseUrl`. We'd need to update the downstream tests instead to validate their expectations.

kmike: I must admit I failed to make this test working, and applied this hack instead :) Mocks are not my friends.

kmike: I was thinking it's harmless, because tests should be getting ResponseUrl, as HttpResponse should be converting str to ResponseUrl; without having mocks in mind the change doesn't really change the behavior. But I'm likely missing some interactions with mocks which are in effect here.

web_poet/page_inputs/client.py
kmike: Nice catch!

tests/test_page_inputs.py
```diff
@@ -72,7 +74,7 @@ def test_http_defaults(cls, body_cls):
     http_body = body_cls(b"content")
 
     obj = cls("url", body=http_body)
-    assert obj.url == "url"
+    assert str(obj.url) == "url"
```
BurnzZ, line 77: The tests so far only checks for equality when casted to a string. We could add assertions for inequality between a `_Url` and `str` to ensure they won't be equal.

tests/test_page_inputs.py
```diff
@@ -189,12 +192,20 @@ def test_http_request_init_full():
...
     assert req_1.body == req_2.body
 
 
+def test_http_request_init_with_response_url():
+    resp = HttpResponse("url", b"")
+    assert isinstance(resp.url, ResponseUrl)
+    req = HttpRequest(resp.url)
+    assert isinstance(req.url, RequestUrl)
+    assert str(req.url) == str(resp.url)
+
```
BurnzZ, line 207: @Gallaecio has raised a good point if `HttpResponse("url")` and `HttpRequest("url")` are equal. We should add a test for such.

tests/test_page_inputs.py
kmike: A good point, done!

tests/test_page_inputs.py
kmike: I've added a test which shows that they're not equal.

But I'm not sure this behavior is useful. What kind of errors would it prevent?

I wonder if it'd better to leave it "undefined", instead of fixing it in tests (if that's even a reasonable approach :)

pathlib does some magic when comparing paths, it's not a simple string match. E.g. on Windows path components are lower-cased before comparison.

BurnzZ: The test simply solidifies the expectation to users in case they're unsure. As a user, I might expect when initially using it for the first time that the expression `response.url == request.url` holds True. However, I need to cast it to its string value to bear True.

I also expect such test to conveniently break when we override `__eq__` to another behavior. It helps future PRs denote that an equality expectation will change.

BurnzZ (approved):
LGTM @kmike ! 🙌

BurnzZ:
It just occurred to me after hitting approve that the docs needed to be updated as well. I took the liberty of creating a PR for it here: https://github.com/scrapinghub/web-poet/pull/48

kmike:
Thanks @BurnzZ and @Gallaecio! I'm merging it, to unblock the releases.

PR #43, opened 2022-05-22
BrowserHtml

kmike:
In this PR url-less BrowserHtml is added. Unlike HttpResponseBody, its type is str, not bytes; this means selectors can be supported directly.

In autoextract-poet we had AutoextractHtml (https://github.com/scrapinghub/autoextract-poet/blob/aac08746c7ca9bc0baf07cfbf7773d616c26b1fb/autoextract_poet/page_inputs.py#L23), which is more similar to HttpResponse, as it contains URL.

I think we can add a similar class later, if needed. A design challenge would be to figure out wht should be a URL class - is the same as ResponseURL, or is it separate (BrowserURL) - assuming we're moving forward with https://github.com/scrapinghub/web-poet/pull/42.

### DIFF
--- a/web_poet/mixins.py
+++ b/web_poet/mixins.py
@@ -7,15 +7,14 @@ from w3lib.html import get_base_url
 
 class SelectableMixin(abc.ABC):
     """
-    Inherit from this mixin, implement ``.text`` property,
+    Inherit from this mixin, implement ``._selector_input`` method,
     get ``.selector`` property and ``.xpath`` / ``.css`` methods.
     """
     __cached_selector = None
 
-    @property
     @abc.abstractmethod
-    def text(self) -> str:
-        raise NotImplementedError()
+    def _selector_input(self) -> str:
+        raise NotImplementedError()  # pragma: nocover
 
     # XXX: see https://github.com/python/mypy/issues/1362
     @property   # type: ignore
@@ -26,7 +25,7 @@ class SelectableMixin(abc.ABC):
         if self.__cached_selector is not None:
             return self.__cached_selector
         # XXX: should we pass base_url=self.url, as Scrapy does?
-        sel = parsel.Selector(text=self.text)
+        sel = parsel.Selector(text=self._selector_input())
         self.__cached_selector = sel
         return sel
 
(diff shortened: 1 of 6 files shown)


web_poet/mixins.py
```diff
@@ -1,11 +1,48 @@
...
+        raise NotImplementedError()  # pragma: nocover
+
+    @property
+    def selector(self) -> parsel.Selector:
+        """Cached instance of :external:class:`parsel.selector.Selector`."""
+        # XXX: caching is implemented in a manual way to avoid issues with
+        # non-hashable classes, where memoizemethod_noargs doesn't work
+        if self.__cached_selector is not None:
+            return self.__cached_selector
+        # XXX: should we pass base_url=self.url, as Scrapy does?
```
BurnzZ, line 26: I think we can remove this comment since it's being used by `BrowserHtml` which doesn't rely on a **url**. Or do you foresee a need for it later on @kmike ?

BurnzZ (approved):
LGTM @kmike !

web_poet/mixins.py
kmike: If that's a right thing to do (which I'm not sure about - it seems it's not needed), we'd need to have URL for selectors to work properly. In this case, having a class like BrowserResponse, which contains both URL and HTML (similar to what we had with AutoextractHtml), might be better.

That said, it won't be a part of BrowserHtml, so it does make sense to remove the comment, thanks @BurnzZ!

kmike: Actually I think it might be better to keep the comment, as it's a part of SelectableMixin, not of BrowserHtml class.

web_poet/page_inputs/browser.py
```diff
@@ -0,0 +1,10 @@
+from web_poet.mixins import SelectableMixin
+
+
+class BrowserHtml(str, SelectableMixin):
```
Gallaecio, line 4: :lipstick: It should not be a problem given the current interface, but I would expect:

```suggestion
class BrowserHtml(SelectableMixin, str):
```

web_poet/page_inputs/browser.py
```diff
@@ -0,0 +1,10 @@
...
+    """ HTML returned by a web browser,
+    i.e. snapshot of the DOM tree in an HTML format.
```
Gallaecio, line 6:
```suggestion
    i.e. snapshot of the DOM tree in HTML format.
```

Gallaecio (approved):

web_poet/page_inputs/browser.py
kmike: nice catch!

gatufo (approved):
LGMT!

PR #44, opened 2022-05-25
web_poet.page_object decorator

kmike:
This is an experiment, to see how "functional" Page Objects could look like, if it's possible to implement them, and to get feedback about the idea.

### DIFF
--- a/web_poet/functional.py
+++ b/web_poet/functional.py
@@ -1,0 +1,52 @@
+from typing import Callable, Type
+
+from web_poet.pages import ItemPage
+
+
+def page_object(func: Callable) -> Type[ItemPage]:
+    """
+    Decorator to create a Page Object from a function.
+    Note that the original function is destroyed, and
+    a class is created instead.
+
+    .. code-block:: python
+
+        @web_poet.page_object
+        def MyPage(resp: HttpResponse):
+            return {"title": resp.css("title::text").get()}
+
+    is a shortcut for
+
+    .. code-block:: python
+
+        class MyPage(web_poet.ItemPage):
+            def __init__(self, resp: HttpResponse):
+                self.resp = resp
+
+            def to_item(self):
+                return {"title": self.resp.css("title::text").get()}
+
+    """
+    class PageObject(ItemPage):
+        def __init__(self, *args, **kwargs):
+            # FIXME: save arguments as properly named attributes
+            # FIXME: __init__ method doesn't fail if parameters are wrong
+            self.args = args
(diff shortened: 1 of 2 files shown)


web_poet/functional.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    Decorator to create a Page Object from a function.
+    Note that the original function is destroyed, and
+    a class is created instead.
+
+    .. code-block:: python
+
+        @web_poet.page_object
+        def MyPage(resp: HttpResponse):
+            return {"title": resp.css("title::text").get()}
```
kmike, line 16: alternative syntax, which doesn't destroy the function:

```py
def parse_foo(resp: HttpResponse):
    return {"title": resp.css("title::text").get()}

FooPage = web_poet.page_object(parse_foo)
```

web_poet/functional.py
```diff
@@ -0,0 +1,52 @@
...
+            def __init__(self, resp: HttpResponse):
+                self.resp = resp
+
+            def to_item(self):
+                return {"title": self.resp.css("title::text").get()}
+
+    """
+    class PageObject(ItemPage):
+        def __init__(self, *args, **kwargs):
+            # FIXME: save arguments as properly named attributes
```
Gallaecio, line 32: Could you elaborate?

web_poet/functional.py
```diff
@@ -0,0 +1,52 @@
...
+
+    # FIXME: robustness, edge cases
+    PageObject.__module__ = func.__module__
+    PageObject.__name__ = func.__name__
+    PageObject.__qualname__ = func.__qualname__
+    PageObject.__doc__ = func.__doc__
+    PageObject.__init__.__annotations__ = func.__annotations__
+
+    # TODO: check that type annotations work properly
+    # TODO: preserve the original function as an attribute?
```
Gallaecio, line 50: Is it to keep the implementation closer to `functools.update_wrapper`? Maybe we could leave this for a future change.

web_poet/functional.py
kmike: This is optional, to keep the generated class as close as possible to the hand-written class. See example above - MyPage has self.resp attribute.

web_poet/functional.py
kmike: I was thinking more about this: if you decorate function, you can't use it anymore, it becomes a class with a completely different API. https://github.com/scrapinghub/web-poet/pull/44/files#r882977954 could be another way to address it.

PR #45, opened 2022-06-01
RequestUrl and ResponseUrl using yarl

BurnzZ:
This PR attempts to scrutinize the idea as noted down in https://github.com/scrapinghub/web-poet/pull/42#issuecomment-1141831361.

It's built on top of this PR's branch:
- https://github.com/scrapinghub/web-poet/pull/42


web_poet/page_inputs/http.py
```diff
@@ -18,12 +19,46 @@
...
+
+    def __str__(self) -> str:
+        return str(self._url)
+
+    def __repr__(self) -> str:
+        return str(self._url)
+
+    def __eq__(self, other) -> bool:
+        return str(self._url) == str(other)
+
```
BurnzZ, line 34: The property methods below can be added mapped dynamically with `yarl`'s. However, we lose the benefit of defining docstrings within them.

web_poet/page_inputs/http.py
```diff
@@ -18,12 +19,46 @@
 _AnyStrDict = Dict[AnyStr, Union[AnyStr, List[AnyStr], Tuple[AnyStr, ...]]]
 
 
-class ResponseURL(str):
+class _URL:
```
BurnzZ, line 22: We could move this later to `w3lib` when we're comfortable in the "smart" URL functionalities that we like.

web_poet/page_inputs/http.py
```diff
@@ -18,12 +19,46 @@
...
-class ResponseURL(str):
+class _Url:
+    def __init__(self, url: Union[str, yarl.URL]):
+        self._url = yarl.URL(str(url))
+
+    def __str__(self) -> str:
+        return str(self._url)
+
+    def __repr__(self) -> str:
+        return str(self._url)
```
Gallaecio, line 30:
```suggestion
    def __repr__(self) -> str:
        return f'{type(self).__name__}({str(self._url)!r})'
```

e.g.

```
ResponseUrl('https://example.com')
```

web_poet/page_inputs/http.py
Gallaecio: We would also loose API governance, so +1 to manual definition.

However, I wonder if we should define them at all for the initial implementation. We want to make sure we get the API right encoding-wise, and if we expose a part of the Yarl interface already as is, I imagine we are introducing the encoding issue in our implementation, with the caveat of not supporting `encoded=True` in `__init__` to at least prevent Yarl from messing things up.

Maybe the initial implementation should use a string internally instead, and we can convert it into Yarl later.

web_poet/page_inputs/http.py
```diff
@@ -18,12 +19,46 @@
...
+
+    def __eq__(self, other) -> bool:
+        return str(self._url) == str(other)
```
Gallaecio, line 33: We should probably make sure we implement this in line with our API expectations.

Mind the following:

```python
>>> example_url_1 = URL("https://example.com")
>>> example_url_2 = URL("https://example.com/")
>>> example_url_1 == example_url_2
True
>>> str(example_url_1) == str(example_url_2)
False
```

Gallaecio: While that is true for `https://example.com/foo` and `https://example.com/foo/`, the case of ``https://example.com` and `https://example.com/` is special, since the path of a URL, when not specified, is `/`.

```python
>>> URL('https://example.com').path
'/'
>>> URL('https://example.com/').path
'/'
>>> URL('https://example.com/foo').path
'/foo'
>>> URL('https://example.com/foo/').path
'/foo/'
```

web_poet/page_inputs/http.py
BurnzZ: A good point about handling the encoding. What do you think about setting `encoded=False`  by default to prevent yarl from messing things up due to incorrect encoding? https://github.com/scrapinghub/web-poet/pull/45/commits/be37f395c2618a068987b3a743f28a80eff054af

web_poet/page_inputs/http.py
BurnzZ: Ahhh thanks for making the distinction clear! Handled this on https://github.com/scrapinghub/web-poet/pull/45/commits/292a3b4b9e15041bc0a7a39ec06e2f8f20c29132.

web_poet/page_inputs/http.py
kmike: I think the goal shouldn't be to implement a general URL class; the goal is to implement URL class useful for web scraping. 

If that's hard to use yarl or other library's URL class directly, and we're defining the API anyways, we probably should think about it from the API point of view: what's the API we want, what are the features commonly used in web scraping? After figuring out how we'd like API to look like, we can see what's the best way to implement it - wrap yarl, wrap w3lib, do something else.

Based on our previous discussions, I think a scraping-ready URL class should have:

* a way to manipulate query string: add/remove/get/update query parameters
* some kind of urljoin method, probably via `/` operation
* probably - a way to extract the domain name?
* anything else?

In addition to this, there is whole bunch of questions about encoding, normalization, converting URLs to ascii-only encoded strings suitable for downloading, etc. The best API to handle all that might require some thought. I wonder if we can side-step it for now somehow.

At the same time, I'm not sure properties like .scheme are that essential. They're are essential for a general-purpose URL class, but are people who write scraping code commonly parse URLs to get their scheme? We can add such methods and properties for sure, but we can do it later. These methods are probably useful for authors of web scraping frameworks / http clients, but less so for people who write web scraping code.

web_poet/page_inputs/http.py
```diff
@@ -18,13 +19,64 @@
...
+    def __str__(self) -> str:
+        return str(self._url)
+
+    def __repr__(self) -> str:
+        return f'{type(self).__name__}({str(self._url)!r})'
+
+    def __eq__(self, other) -> bool:
+        if self._url.path == "/":
+            if isinstance(other, str):
+                other = _Url(other)
```
kmike, line 35: If we follow pathlib route, probably there shouldn't be magic like this:

```py
In [1]: import pathlib

In [2]: p = pathlib.Path("/etc")

In [3]: p == "/etc"
Out[3]: False
```

web_poet/page_inputs/http.py
```diff
@@ -18,13 +19,64 @@
 _AnyStrDict = Dict[AnyStr, Union[AnyStr, List[AnyStr], Tuple[AnyStr, ...]]]
 
 
-class ResponseURL(str):
-    """ URL of the response """
+class _Url:
+    def __init__(self, url: Union[str, yarl.URL], encoded=True):
```
kmike, line 23: `url: Union[str, yarl.URL]` - this type annotation doesn't allow to create RequestUrl from RequestUrl or from ResponseUrl

web_poet/page_inputs/http.py
kmike: > The property methods below can be added mapped dynamically with yarl's. However, we lose the benefit of defining docstrings within them.

It's also about return types - some yarl.URL methods are going to return yarl.URL objects, while here it would make more sense to return `_Url` objects.

web_poet/page_inputs/http.py
BurnzZ: You're right. Initially, I was trying to solve the tiresome casting of `str()` everytime, similar to the concern noted in https://github.com/scrapinghub/web-poet/pull/42#discussion_r888256415. One thing I realized now would be the problem it presents to end users later on. They might be expecting a `"https://example.com"` but misses out that it's actually `ResponseUrl("https://example.com")` due to this equality "magic".

Updated this on: https://github.com/scrapinghub/web-poet/pull/45/commits/912ba772bbb0d1a00955068228e515e8bcd99caf

web_poet/page_inputs/http.py
BurnzZ: Nice catch! Fixed it in https://github.com/scrapinghub/web-poet/pull/45/commits/d869024bd2db2c5f32faf56afbb0f93502316d2a.

BurnzZ: I forgot to annotate the other methods in other modules. In any case, we'd be rebasing to #42 if we should proceed with this route. This would include the annotations like https://github.com/scrapinghub/web-poet/pull/42/commits/5b0e889f4800b1313877fb60e37406889e4ec413#diff-e91005d8dfc7a61689963aafad8b742554243e973f4943ec394cab425bbb61f6R81.

PR #52, opened 2022-06-30
Add urljoin to HttpRequest and HttpResponse

Gallaecio:


web_poet/mixins.py
```diff
@@ -78,3 +81,30 @@ def urljoin(self, url: str) -> str:
         url and baseurl of the response"""
         # FIXME: move it to HttpResponse
         return urljoin(self.base_url, url)
+
```
Gallaecio, line 84: Should we deprecate `ResponseShortcutsMixin` above?

web_poet/mixins.py
```diff
@@ -78,3 +81,30 @@ def urljoin(self, url: str) -> str:
...
+    url: Union[str, RequestUrl, ResponseUrl]
+
+    @property
+    def _base_url(self) -> str:
+        if self._cached_base_url is None:
+            text = getattr(self, "text", "")[:4096]
+            self._cached_base_url = get_base_url(text, str(self.url))
+        return self._cached_base_url
+
+    def urljoin(self, url: Union[str, RequestUrl, ResponseUrl]) -> RequestUrl:
```
Gallaecio, line 105: Is `RequestUrl` the right choice for the output? Should it instead match the input type? Should it always be `str`?

web_poet/utils.py
```diff
@@ -1,7 +1,114 @@
...
+from warnings import warn
+
+
+def _clspath(cls, forced=None):
+    if forced is not None:
+        return forced
+    return f"{cls.__module__}.{cls.__name__}"
+
+
+def _create_deprecated_class(
```
Gallaecio, line 15: Copied from Scrapy, only removing the ability to customize the deprecation warning class, which is here hardcoded as `DeprecationWarning` instead.

web_poet/mixins.py
```diff
@@ -78,3 +81,30 @@ def urljoin(self, url: str) -> str:
...
```
BurnzZ, line 105: Instead of adding it as mixin, I was thinking that `urljoin()`  and `basepath` could perhaps be added in the smart "url" class that we were discussing before.

Gallaecio: `response.url / arg` syntax? Should `response / arg` also work?

web_poet/mixins.py
```diff
@@ -78,3 +81,30 @@ def urljoin(self, url: str) -> str:
...
+class _UrlMixin(abc.ABC):
+    """Common shortcut methods for working with the URL of HTTP requests and
+    responses.
+
+    This mixin could be used with Page Object base classes.
+
+    It requires the ``url`` attribute to be present.
+    """
+
+    _cached_base_url = None
```
kmike, line 95: For mixin it could make sense to use `__cached_base_url` name, so that there won't be a name clash in a subclass

web_poet/mixins.py
```diff
@@ -78,3 +81,30 @@ def urljoin(self, url: str) -> str:
...
+    url: Union[str, RequestUrl, ResponseUrl]
+
+    @property
+    def _base_url(self) -> str:
+        if self._cached_base_url is None:
+            text = getattr(self, "text", "")[:4096]
```
kmike, line 101: It seems having "text" attribute is another requirement for this mixin - should we document it? 

In SelectableMixin it's solved with a private abc method, this allows the name of the attribute to change between classes (e.g. ResponseShortcutsMixin uses "html"). But this may be not important here.

web_poet/page_inputs/__init__.py
```diff
@@ -7,7 +7,6 @@
     HttpResponse,
     HttpResponseBody,
     HttpResponseHeaders,
-    RequestUrl,
-    ResponseUrl,
 )
 from .page_params import PageParams
+from .url import RequestUrl, ResponseUrl
```
kmike, line 12: 👍

web_poet/page_inputs/url.py
```diff
@@ -0,0 +1,13 @@
+from web_poet._base import _Url
```
kmike, line 1: we may move _Url here as well

web_poet/utils.py
kmike: should we copy tests as well?

web_poet/mixins.py
BurnzZ: @kmike Good point, I haven't considered the `<base>` tag in this case. What do you think about the following behaviors?

1. Using `response.url.join("path/somewhere")` and `response.url / "path/somewhere"` would only take into account the URL value. It's similar to https://yarl.readthedocs.io/en/latest/api.html?highlight=join#yarl.URL.join.

2. When using `response.urljoin("path/somewhere")` _(which has access to the response body/html)_, it adds a new feature of checking the base tag via https://w3lib.readthedocs.io/en/latest/w3lib.html#w3lib.html.get_base_url.

However, there's something that I don't like about the names above, where `response.url.join()` and `response.urljoin()` look very similar and might cause some debugging issues later on.

web_poet/mixins.py
Gallaecio: Nevermind, it is still used by `web_poet.pages.WebPage`.

web_poet/mixins.py
Gallaecio: I will document it.

SelectableMixin’s solution makes sense for a public mixin, but I don‘t think we need to do that here, this being a private mixin. We can always do it later on if we want our built-in classes to expose `text` with different names.

web_poet/mixins.py
```diff
@@ -78,3 +81,34 @@ def urljoin(self, url: str) -> str:
         url and baseurl of the response"""
         # FIXME: move it to HttpResponse
         return urljoin(self.base_url, url)
+
+
+class _UrlMixin(abc.ABC):
```
kmike, line 86: There are no abstract methods in this class, is abc.ABC base class needed?

web_poet/mixins.py
```diff
@@ -78,3 +81,34 @@ def urljoin(self, url: str) -> str:
...
+    """Common shortcut methods for working with the URL of HTTP requests and
+    responses.
+
+    This mixin could be used with Page Object base classes.
+
+    It requires the ``url`` attribute to be present.
```
kmike, line 92: I wonder if it'd be better to just move the relevant methods and attributes to HttpResponse class, without creating a mixin. 

It seems there are 2 use cases for having a mixin instead of implementing necessary methods directly:

1. To use it for HttpRequest.
2. To use it as a mixin for page objects.

For HttpRequest there is no base_url, it seems the implementation of .urljoin can be pretty straightforward - just a shortcut to urljoin. It doesn't look like there is going to be code duplication.

For custom page objects the interface doesn't look straightforward - user needs to implement url and text properties. I think we can think about a better shortcut, if that's needed (e.g. a base class). A shortcut may proxy the call to self.response.urljoin, so there won't be duplication either.

Also, we can always add a mixin later, if it turns out there are more advantages to it. But currently for me it looks like moving this logic to HttpResponse class makes the code simpler, and removes these implicit "url" and "text" dependencies, as well as indirection.

web_poet/mixins.py
kmike: A good question.

RequestUrl and str seems reasonable choices; I don't see a use case for matching the input type.

RequestUrl looks like a right choice, but we need to make sure that it works well with zyte-common-items. For example, it must be possible to extract image or other URLs using .urljoin, and create items with the result, without boilerplate.

web_poet/mixins.py
kmike: I think regardless of `response.url / "foo"` behavior, it's fine to have response.urljoin, and the existing logic (using base url) is good. So, we can handle ResponseUrl case separately.

PR #53, opened 2022-07-20
web_poet.field decorator

kmike:
An experiment: a minimal version of `@field` decorator.


tests/test_fields.py
```diff
@@ -0,0 +1,73 @@
...
+        "http://example.com",
+        b"""
+    <html>
+    <head><title>Hello!</title>
+    </html>
+    """,
+    )
+    page = Page(response=response)
+
+    assert page.name() == "Hello!"
```
kmike, line 60: Note that fields in these PRs are preserved as extraction methods, they're not converted to properties. I don't have a super-strong opinon about it, but it’s done like this for a few reasons:

1. It makes it clear that computation could be expensive; IMO it’s much easier to miss it when accessing a property. With a method it's explicit that there is a potentially heavy computation which can happen. Less opportunity to forget about cache when it’s needed.
2. It allows to add optional arguments to these methods, if really needed. These arguments won’t be used in to_item, but could be used elsewhere.
3. Implementation is simpler (?), no need for concepts like “async property”.

web_poet/fields.py
```diff
@@ -0,0 +1,64 @@
...
+    def __get__(self, obj, objtype=None):
+        return MethodType(self.meth, obj)
+
+
+async def item_from_fields(obj, item_cls=dict):
+    """Return an item of ``item_cls`` type, with its attributes populated
+    from the ``obj`` methods decorated with :class:`field` decorator.
+    """
+    data = item_from_fields_sync(obj, item_cls=dict)
+    return item_cls(**{name: await maybe_await(value) for name, value in data.items()})
```
BurnzZ, line 58: This logic is nifty as we could extend it so that fields that need additional requests could be ignored if needed. 💡

web_poet/fields.py
```diff
@@ -0,0 +1,63 @@
...
+        @field
+        def price(self):
+            return self.response.css(".price").get()
+
+        @field
+        def currency(self):
+            return "USD"
+
+        async def to_item(self):
+            return await item_from_fields(self)
```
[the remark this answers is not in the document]
kmike: It could also be 

```py
    # ...
```

This should allow to set proper typing info for the `to_item` method, and provide a default implementation.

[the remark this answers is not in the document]
kmike: We should also think about interaction with handle_urls decorator: in the example below it seems there is no need for `overrides` keyword argument:

```py
@handle_url("foo.com")
    # ...
```

tests/test_create_deprecated_class.py
```diff
@@ -0,0 +1,255 @@
+import inspect
```
kmike, line 1: This is not anything new, I moved tests related to create_deprecated_class to a separate file.

kmike: See also: https://github.com/scrapy/scrapy/issues/2194 :)

docs/advanced/fields.rst
```diff
@@ -0,0 +1,206 @@
...
+        @field
+        async def price(self):
+            resp = self.http_client.get("...")
+            return resp.json()['price']
+
+        async def to_item(self) -> dict:
+            return item_from_fields(self)
+
+Because :func:`~.item_from_fields` supports both sync and async extraction
+methods, it's recommended to use it over :func:`~.item_from_fields_sync`, even
```
BurnzZ, line 120: What are the specific scenarios that we're only allowed to use the sync function instead of the async? (given that the latter supports both cases)

For context, I'm wondering if we should not expose the sync function to make the choice easier for users.

docs/advanced/fields.rst
```diff
@@ -0,0 +1,206 @@
...
+if there are no async extraction methods yet.
+
+Item clasess
```
BurnzZ, line 123:
```suggestion
Item classes
```

docs/advanced/fields.rst
```diff
@@ -0,0 +1,206 @@
...
+    @attrs.define
+    class MyPage(ItemPage):
+        # ...
+        def to_item(self) -> Item:
+            return Item(
+                name=self.name(),
+                price=self.price()
+            )
+
+web_poet.fields support it, by allowing to pass an item class to the
```
BurnzZ, line 150:
```suggestion
:mod:`web_poet.fields` support it, by allowing to pass an item class to the
```

docs/advanced/fields.rst
Gallaecio: The only justification for the `_sync` version I can think of would be a slight performance improvement. However, if it is significant and we can not reduce the performance gap between both methods, I think the documentation needs to indicate so, and otherwise, I agree that it may be best not to provide `_sync` in the first place.

kmike: I'm fine with making it private. Main motivation to include it was to allow using field decorators with page objects where to_item method is not async.

web_poet/fields.py
```diff
@@ -0,0 +1,65 @@
...
+class field:
+    """
+    Page Object methods decorated with ``field`` decorator
+    are called by :func:`item_from_fields` or :func:`item_from_fields_sync`
+    to populate item attributes.
+    """
+
+    def __init__(self, meth):
+        if not callable(meth):
+            raise TypeError(f"@field decorator must be used on methods, " f"{meth!r} is decorated instead")
```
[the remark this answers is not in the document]
kmike: ouch, this is an auto-formatting issue (string was split by me, and then moved to the same line by black)

kmike, line 41:
```suggestion
            raise TypeError(f"@field decorator must be used on methods, {meth!r} is decorated instead")
```

tests/test_fields.py
Gallaecio: Point 3 sells it for me, as I have a feeling things can get complicated when mixing properties and both sync and async code.

I do wonder, thinking about 1, if [caching](https://github.com/scrapinghub/web-poet/pull/53/files#r927287307) wouldn’t be a good default (i.e. opt-out, rather than opt-in) feature of `@field`, as I imagine in most cases it will be desirable.

And as for 2, in real page objects I would not expect that, I would expect methods needing arguments to not be field-decorated, i.e. be some private methods that field-decorated methods call.

Which makes me think, if 3 were to be a non-issue (I have a *feeling* it can be an issue, but I do not know for a fact), maybe going for properties would not be such a bad idea, as it would discourage 2.

kmike: About (1), there is a note in PEP8 (see https://peps.python.org/pep-0008/#designing-for-inheritance):

> Note 2: Avoid using properties for computationally expensive operations; the attribute notation makes the caller believe that access is (relatively) cheap.

kmike: Having cache opt-out, not opt-in, sounds good.

web_poet/fields.py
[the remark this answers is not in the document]
Gallaecio: This is @kmike writing code with PEP8 guidelines (79 max length) and black not being smart when going wider. Happens to me all the time :facepalm:

I thought newer versions of Black were addressing this, but [it seems it is experimental and needs explicit enabling](https://github.com/psf/black/issues/2188).

docs/advanced/fields.rst
Gallaecio: Right, that is a valid point.

I have no strong opinion on providing both vs forcing users to use `async def to_item`, although I am slightly in favor of the later.

Gallaecio (approved):

kmike:
I'm converting it to draft to make cache opt-out, because otherwise it'd be a backwards incompatible change later.

tests/test_fields.py
kmike: I tried to add cache by default to field, but I'm not sure if that's good or not. 

It seems it won't hurt much to have such cache. But, at the same time, it seems it's most useful to cache things which are not fields. For example, there could be an API call made with additional requests, and multiple attributes might be extracted from this call. User should create a method (or a property?) to do this call, and then use it in attributes:

```py

class MyPage(ItemPage):
    response: HttpResponse
    http: HttpClient

    @i_want_to_be_cached
    async def api_result(self):
        api_url = self.response.css("...").get()
        return self.http.get(api_url).json()

    @field
    def price(self):
        res = await self.api_result()
        return res['price']   

    @field
    def name(self):
        res = await self.api_result()
        return res['name']   
```

It's not that important to cache name and price attributes, but caching api_result is. Again, having cache for attributes probably won't hurt, but that doesn't seem to be the main concern.

kmike: Actually, cache is more tricky for methods than for properties. I'll try to be careful with the implementation. For properties there is cached_property available in stdlib since Python 3.8, but for methods there is only lru_cache, which doesn't work properly for methods - `self` argument is used as a key in cache, and it prevents deallocation of instances.

Gallaecio: Maybe opt-out caching is not that good of an idea. I am OK with merging without any caching. Giving users freedom to choose their caching approach may work best. And implicit caching could be considered too magical, possibly confusing.

kmike: Caching of method resuts is surprisingly tricky: built-in lru_cache doesn't work properly on methods, and workarounds which use weakrefs to `self` don't work either for Page Object methods, as they require POs to be hashabe, while they're not always hashable - e.g. if they have HttpResponse dependency, then they're not, because headers are not hashable.

I'd like to get some solution for this before merging this PR. The problem is solved more easily for properties.

kmike: I've reverted field to not use cache by default, added `cached_method` decorator, and `field(cached=True)` shortcut. Todo: add documentation. 

Opt-out doesn't seem to be a good idea - it seems for most fields it's not going to be useful, and we may be increasing memory usage for no good reason.

web_poet/fields.py
```diff
@@ -24,31 +24,48 @@ async def to_item(self):
...
-            raise TypeError(f"@field decorator must be used on methods, {meth!r} is decorated instead")
-        self.meth = meth
 
-    def __set_name__(self, owner, name):
-        if not hasattr(owner, "_auto_item_fields"):
-            # dict is used instead of set to preserve the insertion order
-            owner._auto_item_fields = {}
-        owner._auto_item_fields[name] = True
+    By default, decorating a method with ``@field`` decorator make them cache
+    their results. Use ``@field(cached=False)`` to disable this behavior.
```
Gallaecio, line 40:
```suggestion
    Use ``@field(cached=True)`` to cache the method result.
```

web_poet/utils.py
```diff
@@ -126,6 +133,63 @@ def new_method(self, *args, **kwargs):
...
+
+    .. note::
+
+        :func:`cached_method` is needed because :func:`functools.lru_cache`
+        doesn't work well on methods: self is used as a cache key,
+        so a reference to an instance is kept in the cache, and this
+        prevents deallocation of instances.
+    """
+    name_hash = hashlib.sha1(method.__name__.encode("utf8")).hexdigest()[:4]
+    cached_meth_name = f"_cached_{method.__name__}_{name_hash}"
```
Gallaecio, line 151: I read SHA-1 is quite fast, but why do we need a hash here in the first place?

web_poet/utils.py
```diff
@@ -126,6 +133,63 @@ def new_method(self, *args, **kwargs):
...
+    if inspect.iscoroutinefunction(method):
+        meth = _cached_method_async(method, cached_meth_name)
+    else:
+        meth = _cached_method_sync(method, cached_meth_name)
+
+    meth.cached_method_name = cached_meth_name
+    return meth
+
+
+def _cached_method_sync(method, cached_method_name, maxsize=None):
```
Gallaecio, line 161: It seems pointless to implement `maxsize` in these private functions if we are not exposing it through the public functions that use them.

web_poet/fields.py
BurnzZ: We could also mention that the cache replacement policy is LRU.

web_poet/utils.py
BurnzZ: If we have:
```python
class Foo:
    def bar(self):
        print("hi")
```

Then `hashlib.sha1(Foo().bar.__name__.encode("utf-8")).hexdigest()[:4]` would always be `62cd`. In turn, `cached_meth_name` would have a value of `"_cached_bar_62cd'"`.

I'm also curious as to why the 4-character permutation at the end is needed, if they should be constant anyways.

web_poet/fields.py
kmike: It's not, the cache is unbounded.

web_poet/utils.py
kmike: I added it as poor man's approach to avoid collisions with user-defined attributes, as an additional safeguard. You may have `self._cached_price` defined by user, but it's unlikely user defines `self._cached_price_62cd`.

kmike: Speed shoudn't be an issue, as this code runs once, on import time.

web_poet/utils.py
kmike: Right; the plan was to expose them, but I was a bit lazy :) I'll either remove it, or implement it.

web_poet/utils.py
Gallaecio: I think using `_cached_method_` as prefix and covering this in the documentation of the `cached_method` function should be enough safeguard. I mean, if anyone decorates `bar` with `@cached_method` and then defines `_cached_method_bar`, they are looking for trouble even without documentation warning them.

Gallaecio:
:heart: the docs

tests/test_fields.py
```diff
@@ -0,0 +1,141 @@
...
+    price: str
+
+
+@attrs.define
+class Page(ItemPage):
+    response: HttpResponse
+
+    @field
+    def name(self):  # noqa: D102
+        return self.response.css("title ::text").get()
```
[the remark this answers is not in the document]
kmike: Added field(cached=True) and cached_method decorators.

web_poet/utils.py
```diff
@@ -160,14 +160,14 @@ def cached_method(method):
...
-def _cached_method_sync(method, cached_method_name, maxsize=None):
+def _cached_method_sync(method, cached_method_name):
     @wraps(method)
     def inner(self, *args, **kwargs):
         if not hasattr(self, cached_method_name):
             # on a first call, create a lru_cache-wrapped method,
             # and store it on the instance
             bound_method = MethodType(method, self)
-            cached_meth = lru_cache(maxsize=maxsize)(bound_method)
+            cached_meth = lru_cache(maxsize=None)(bound_method)
```
Gallaecio, line 170: I just found out that there is `functools.cache`, equivalent to `functools.lru_cache(maxsize=None)`.

kmike: Right, but it's 3.9+ only :) Implementation is literally `functools.lru_cache(maxsize=None)` anyways.

Gallaecio: :facepalm:

Gallaecio (approved):
1. I am OK with methods. However, if properties are trivial to implement, and we expect no issue with them, it may be worth implementing now, because we could implement them into `@field` itself. Otherwise, future property support would require users to combine `@field` and `@property`. Properties would be slightly cleaner, I think.
2. I am OK with merging with the “bug”.
3. I think feature-wise this is quite complete, and should cover most use cases. I don’t think caching was even necessary for the initial implementation, but it is great to have it. Post-processing and schema validation (?) are better addressed/discussed later, separately.
4. Very happy with the naming :slightly_smiling_face: 
5. :+1:. I can think of use cases that could take advantage of field caching, e.g. fields using complex regular expressions, or things like JavaScript evaluation or OCR. But I think the way it is covered in the documentation is perfect.

BurnzZ (approved):
I believe this PR is good to go. 👍 Nice work @kmike !

docs/advanced/fields.rst
```diff
@@ -0,0 +1,367 @@
...
+        response: HttpResponse
+
+        @field
+        def nane(self):
+            return self.response.css(".name").get()
+
+        async def to_item(self) -> Item:
+            return await item_from_fields(self, item_cls=Item)
+
+Because Item class is used, a typo ("nane" instead of "name") is detected
```
gatufo, line 243: Love the idea of detecting missing required arguments, but I'm **not sure about raising errors for fields not present in the Items**. However, I might be biased by my approach of seeing POs as field "providers"...

Take this Item+PO:

```python
@attrs.define
class Item:
    name: str
    price: str


@attrs.define
class DaaSProduct(ItemPage):

    @field
    def name(self):
        return self.response.css(".name").get()

    @field
    def price(self):
        return self.response.css(".price").get()

    async def to_item(self) -> Item:
        return await item_from_fields(self, item_cls=Item)

```

Let's say we want to create a custom PO subclassing it in order to reuse some fields:

```python
@attrs.define
class ClientItem:
    name: str
    description: str


@attrs.define
class ClientProduct(DaaSProduct):

    @field
    def description(self):
        return self.response.css(".price").get()

    async def to_item(self) -> ClientItem:
        return await item_from_fields(self, item_cls=ClientItem)
```
An error will be raised for `price` field here, as is not present in `ClientItem` definition.

gatufo (approved):
1. No need to answer to that one 😄  You know I see POs as field providers, and with that approach in mind, properties are cleaner to me.
2. Ok to merge it as-is 👍 
3. I don't see any missing feature for fields so far, I think some parts can be simplified (like not needing to implement the `to_item` method as it can be automatically generated), but that is something that can be build later on top of this.
4. Ok with names.
5. I don't have a strong opinion about this, in `eddie-poc` I'm providing `po.cached_property` decorator for fields to use cached properties when needed.

Great work as usual @kmike 👏 ❤️ 

My only concern in order to replace `field` decorator in `eddie-poc`/`sue` with this one, is to allow usage of POs as field providers and not raise errors for fields not present in the Item definition.

docs/advanced/fields.rst
kmike: That's a great point @gatufo, I haven't considered it.

I'm unsure about the default, but it looks like we need to provide an option to skip the fields not present in item. In order to do so, we need to get a list of item fields; I can't find a way around it, if we want to handle required fields as well.

Itemadapter has a method to get item fields, but it works on instances, not on classes. So, it seems the way to go is to implement `ItemAdapter.get_field_names_from_class`; then, modify `item_from_fields_sync` function like this:

```py
def item_from_fields_sync(obj, item_cls=dict, only_item_fields=False):
    """Synchronous version of :func:`item_from_fields`."""
    page_field_names = list(getattr(obj, "_auto_item_fields", {}))
    if only_item_fields:
        try:
            item_field_names = itemadapter.ItemAdapter.get_field_names_from_class(item_cls)
            # fixme: preserve field order
            field_names = set(page_field_names) | set(item_field_names)
        except itemadapter.FieldNamesUnknown:  # e.g. when item_cls is dict
            field_names = page_field_names
    else:
        field_names = page_field_names
    return item_cls(**{name: getattr(obj, name) for name in field_names})

```

kmike: PR to itemadapter: https://github.com/scrapy/itemadapter/pull/64

kmike: It is partially addressed here: https://github.com/scrapinghub/web-poet/pull/53/commits/80473c2a732db44a7d24ab9945dae3eb003da0a2. TODO: tests.

kmike: @gatufo it's ready for another round of review! The most important piece to check is the documentation - "Item classes" section and its sub-sections, especially "Changing Item type".

Don't use the commit linked above, there are some changes since then.

docs/advanced/fields.rst
```diff
@@ -0,0 +1,484 @@
...
+
+.. code-block:: python
+
+    page = MyPage(...)
+    price = await page.price
+
+Using Page Objects with async fields
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you want to a Page Object with async fields without calling its
```
Gallaecio, line 142: There is a verb missing here, at least.

web_poet/fields.py
```diff
@@ -0,0 +1,104 @@
...
+    if item_cls_fields:
+        field_names = _without_unsupported_field_names(item_cls, field_names)
+    return item_cls(**{name: getattr(obj, name) for name in field_names})
+
+
+def _without_unsupported_field_names(item_cls, field_names):
+    item_field_names = ItemAdapter.get_field_names_from_class(item_cls)
+
+    if item_field_names is not None:
+        field_names = list(set(field_names) & set(item_field_names))
```
Gallaecio, line 102: Wouldn’t this cause, in the example from @gatufo, the new `description` field to also be dropped? If so, I think we need to adjust this set operation accordingly and add a test for that scenario.

Gallaecio: Nevermind.

In the example from @gatufo, `field_names = item_field_names` would have worked the same, so I was trying to think of different scenarios. But the `and` operations seems to make sense in all of them.

web_poet/fields.py
```diff
@@ -0,0 +1,104 @@
...
+    else:
+        # @field(...) syntax
+        return _field
+
+
+async def item_from_fields(obj, item_cls=dict, *, item_cls_fields=False):
+    """Return an item of ``item_cls`` type, with its attributes populated
+    from the ``obj`` methods decorated with :class:`field` decorator.
+
+    If ``item_cls_fields`` is True, ``@fields`` which names don't match
```
Gallaecio, line 78:
```suggestion
    If ``item_cls_fields`` is True, ``@fields`` whose names don't match
```

Gallaecio (approved):
Great work!

gatufo (approved):
hey @kmike, I still think that just "provided" fields should be the default option here, but `item_cls_fields` works for me too 😄 

LGTM! Nice work! 👍

kmike:
Thanks for the reviews @Gallaecio @gatufo @BurnzZ!

@gatufo
 
> I still think that just "provided" fields should be the default option here, but item_cls_fields works for me too

Are you concerned only about the default value for `item_cls_fields` parameter, or do you have concerns about the guidelines, on when to use `item_cls_fields=True` vs `item_cls_fields=False`?

web_poet/fields.py
kmike: I hope the updated implementation (https://github.com/scrapinghub/web-poet/pull/53/commits/d8c77bebcf1a315350537a8b1190022fac708c42) is more readable.

gatufo:
> Are you concerned only about the default value for item_cls_fields parameter, or do you have concerns about the guidelines, on when to use item_cls_fields=True vs item_cls_fields=False?

@kmike Is more about guidelines and how I see usage...
For me when combining both Item and Page Objects, Item defines the fields that what you want to extract, and Page Objects become just providers of those fields. That's why I think the default value should be `True` here to allow (or encourage) this use by default.

kmike:
Let's discuss it in https://github.com/scrapinghub/web-poet/issues/59

PR #55, opened 2022-07-20
New Retry exception

Gallaecio:
I think this is the direction that we want to go.

I was worrying at first about how this could impact additional requests, but then I realized that page objects can handle retries of additional requests on their own when those do not require first retrying the original request, and when they do, they can raise `Retry` to trigger that. Retries trigger a new page object, so they cannot keep state for additional request retries across original request retries, but that should not be a big issue.

It does feel like examples like https://web-poet.readthedocs.io/en/stable/intro/tutorial.html#final-result are becoming increasingly misleading, and I am starting to wonder if we should start providing some non-Scrapy reference framework implementation that better illustrates what the simplest web-poet framework could look like. But that is a separate topic.

No idea how to cover this change in tests.

### DIFF
--- a/docs/advanced/retries.rst
+++ b/docs/advanced/retries.rst
@@ -4,14 +4,22 @@
 Retries
 =======
 
-The webpages of some websites can be unreliable. For example, sometimes
+The responses of some websites can be unreliable. For example, sometimes
 a request can get a response that may only include a part of the data to be
-extracted, or no data at all, but sending a follow-up, identical request can
-get you the expected data.
+extracted, no data at all, or even data unrelated to your request, but sending
+a follow-up, identical request can get you the expected data.
 
-You can write your page object so that it raises
-:exc:`~web_poet.exceptions.core.Retry` when it detects missing data that may become
-available after a request retry:
+Pages objects are responsible for handling these scenarios, where issues with
+response data can only be detected during extraction.
+
+.. _retries-input:
+
+Retrying Page Object Input
+==========================
+
+When the bad response data comes from the inputs that your web-poet framework
+supplies to your page object, your page object must raise
+:exc:`~web_poet.exceptions.core.Retry`:
 
 .. code-block:: python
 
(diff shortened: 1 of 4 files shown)


kmike (approved):

docs/advanced/retries.rst
```diff
@@ -0,0 +1,57 @@
+.. _retries:
+
+=======
+Retries
+=======
```
BurnzZ, line 5: What do you think about having subsections that cover the topic of "how to retry":
1. the entire Page Object itself, and
2. Additional Requests

For the 2nd one, we can do that in another PR.

docs/advanced/retries.rst
```diff
@@ -0,0 +1,57 @@
...
+
+The webpages of some websites can be unreliable. For example, sometimes
+a request can get a response that may only include a part of the data to be
+extracted, or no data at all, but sending a follow-up, identical request can
+get you the expected data.
+
+You can write your page object so that it raises
```
BurnzZ, line 12: minor: should we use "Page Object" (capitalized first letters) to be consistent with how it's being written in other parts of the doc like https://web-poet.readthedocs.io/en/stable/advanced/additional-requests.html ? or the other way around?

BurnzZ (approved):
LGTM @Gallaecio ! I got a few very minor comments but we could address them in another PR or discuss separately.

kmike:
>  I am starting to wonder if we should start providing some non-Scrapy reference framework implementation 

I think it'd be pretty cool to have a framework which you can use to run Page Objects in IPython notebooks. This would allow to develop extraction code interactively, and it may be an alternative to old GSoC projects which aimed to integrate Scrapy with IPython.

docs/advanced/retries.rst
Gallaecio: I think the first one is covered in this page.

The second one I think we can address in this pull request, but I think it makes more sense in the page about additional requests, and not in a section here.

docs/advanced/retries.rst
Gallaecio: My thinking is that Page Object is correct when referring to the design pattern, as it is a proper noun, but for the actual objects “page object” would be the right capitalization. So I don’t see “Page Object” as accurate in this context.

The documentation is already quite inconsistent in this regard (see https://web-poet.readthedocs.io/en/stable/intro/from-ground-up.html#page-objects), so I felt entitled to pick the lowercase option.

I am in favor of making the wording consistent across the documentation, even if that means capitalizing the term always, but I think it would be best done in a separate pull request.

BurnzZ:
> I think it'd be pretty cool to have a framework which you can use to run Page Objects in IPython notebooks
+1 

Such a framework would be especially useful when having POs behind an API.

docs/advanced/retries.rst
BurnzZ: >  but I think it makes more sense in the page about additional requests, and not in a section here.

We could have a dedicated section for it in additional requests and have it linked inside this Retry page (to prevent writing the same things).

I believe it would be much easier for users to find the content if they'd like to perform "retries" for ARs since "Retry" is it's own section, as opposed to clicking "Additional Requests" and stumble in the subsections from the sidebar.

image

Gallaecio: I ended up going for your original approach, just so that I did not had to introduce the scenarios where retries need to be handled in both pages. As a result, I also moved the retry page below the additional requests section in the table of contents.

docs/advanced/retries.rst
```diff
@@ -0,0 +1,117 @@
...
+
+    import attrs
+    from tenacity import retry, stop_after_attempt
+    from web_poet import HttpClient, HttpRequest, ItemWebPage
+
+    @attrs.define
+    class MyPage(ItemWebPage):
+        http_client: HttpClient
+
+        @retry(stop=stop_after_attempt(3))
```
kmike, line 71: nice!

kmike:
Looks good @Gallaecio! I'll leave it to @BurnzZ to merge it if he's happy :)

BurnzZ:
Looking good @Gallaecio ! Cheers for going the extra mile on this PR! 🙌

PR #58, opened 2022-07-28
Fields as properties

kmike:
The code changes turn out to be trivial :) 

It might be possible to simplify the implementation, and don't use cached_method to implement `field` decorator, but I'm not sure it worth it.

For properties there might be a stronger cause for making them cached by default. I'm worried about code like this:

```py
if page.name:
    item.product_name = page.name
```

Some features might be a bit harder to implement for properties later, e.g. exposing cache stats for cached properties or fields.

### DIFF
--- a/docs/advanced/fields.rst
+++ b/docs/advanced/fields.rst
@@ -21,16 +21,18 @@ individual attributes, a method or property per attribute:
     class MyPage(ItemPage):
         response: HttpResponse
 
+        @property
         def name(self):
             return self.response.css(".name").get()
 
+        @property
         def price(self):
             return self.response.css(".price").get()
 
         def to_item(self) -> dict:
             return {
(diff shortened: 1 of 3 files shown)


docs/advanced/fields.rst
```diff
@@ -151,8 +166,8 @@ function when accessing the fields:
     from web_poet.utils import ensure_awaitable
 
     page = MyPage(...)
-    name = await ensure_awaitable(page.name())
-    price = await ensure_awaitable(page.price())
+    name = await ensure_awaitable(page.name)
+    price = await ensure_awaitable(page.price)
```
kmike, line 170: I think this example is nicer now.

Gallaecio (approved):

BurnzZ (approved):

PR #66, opened 2022-08-17
declare web-poet as a package with type information

### LINKED ISSUE #65: PEP 561 compliance
https://peps.python.org/pep-0561/ should be followed in order for the `web-poet` to properly work on **mypy**. Otherwise, the following error is raised: `Skipping analyzing "web_poet": module is installed, but missing library stubs or py.typed marker`.

kmike:
Fixes #65.


BurnzZ:
Thanks @kmike ! I didn't realize that adding it would be this quick. 🚀

BurnzZ (approved):

Issue #67, opened 2022-08-17
Use default black options


kmike:
Currently we override line length to be 120. I'm not sure what's the reason for this, and propose to use black defaults.

PR #68, opened 2022-08-17
Use default black options

kmike:
Fixes https://github.com/scrapinghub/web-poet/issues/67.

I'm assuming @Gallaecio is not against it, based on https://github.com/scrapinghub/scrapy-poet/pull/68#discussion_r901532514 :)


BurnzZ (approved):
Let's add the commit to the `.git-blame-ignore-revs` file. 

Other than that, LGTM @kmike !

kmike:
> Let's add the commit to the .git-blame-ignore-revs file.

A good catch, thanks @BurnzZ!

Issue #69, opened 2022-08-17
web_poet.fields doesn't work properly with subclasses


kmike:
The implementation uses a class attribute to store information about fields. But when an object is subclassed, the attribute is already present, so it's reused. It means that any field defined in a subclass also changes the fields in the base class.

PR #62, opened 2022-08-12
allow to store meta information about Page Object fields

kmike:
Motivation:

1. Allow custom extensions in third-party code, a point where additional information about a field can be stored
2. Introduce a structured FieldInfo, which can be used in other features

PR #70, opened 2022-08-17
Fixed fields in subclasses.

kmike:
Fixes GH-69.

~~I'm sending this PR against #62 because the same code is touched, and it seems #62 is close enough.~~

### DIFF
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -329,3 +329,17 @@ def test_field_subclassing():
 
     assert get_fields_dict(Page)["field3"].meta is None
     assert list(get_fields_dict(Page)) == ["field1", "field3"]
+
+    class Page3(Page2):
+        @field
+        def field3(self):
+            return 2
+
+    assert get_fields_dict(Page3)["field3"].meta is None
+    assert list(get_fields_dict(Page3)) == ["field1", "field3", "field2"]
+
+    assert get_fields_dict(Page)["field3"].meta is None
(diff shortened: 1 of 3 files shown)


tests/test_fields.py
```diff
@@ -297,3 +300,32 @@ def to_item(self):
...
+
+        @field(meta={"foo": "bar"})
+        def field3(self):
+            return 1
+
+    assert get_fields_dict(Page2)["field3"].meta == {"foo": "bar"}
+    assert list(get_fields_dict(Page2)) == ["field1", "field3", "field2"]
+
+    assert get_fields_dict(Page)["field3"].meta is None
+    assert list(get_fields_dict(Page)) == ["field1", "field3"]
```
BurnzZ, line 331: Should we also add a case for symmetry, where `Page` has a field like:
```python
@field(meta={"a": "b"})
def field4(self):
    return 1
```

But then on the `class Page2(Page)` subclass, it became:

```python
@field
def field4(self):
    return 1
```

BurnzZ (approved):
Aside from a small comment below, LGTM @kmike ! 🎉

tests/test_fields.py
kmike: Done in https://github.com/scrapinghub/web-poet/pull/70/commits/26b9508a450cb975251de59bf7218fa8a3173dd3

Gallaecio (approved):

PR #71, opened 2022-08-17
Improve mypy config

kmike:
A minor cleanup.


kmike:
sorry, let me send it against master instead

PR #72, opened 2022-08-17
improve mypy setup:

kmike:
* move config to pyproject.toml
* add show_error_codes, to make it easier to write ignores
* bump mypy version


BurnzZ (approved):
LGTM @kmike !

I was also thinking if we should add `check_untyped_defs = true` next time to cleanup some of the issues that **mypy** has found.  I've created https://github.com/scrapinghub/web-poet/issues/73 to track this.

PR #74, opened 2022-08-18
New ItemPage

kmike:
This PR is built on top of https://github.com/scrapinghub/web-poet/pull/62 and https://github.com/scrapinghub/web-poet/pull/70, and should be only considered for merging after these 2 PRs. It could also make sense to review those 2 first.

The idea here is to extend ItemPage class, so that it provides a default to_item implementation, which uses web_poet fields if they're available. 

By default, dicts are created, but ItemPage can be parametrized with the item type. 

```py
class ProductPage(ItemPage[Product]):
    pass
``` 

In the example above, `await product_page.to_item()` would output a Product item, not a dict. Mypy is inferring types correctly for this case, i.e. it knows that item would be Product.

It's possible to change the item type in a subclass, by inheriting from `Returns[MyItem]` (see the tutorial). This feature works from the output point of view, but I was unable to make typing work properly in this case (commented inline). It could be a mypy bug though, I'm not sure.

Use cases with extended items (adding new fields) and with custom items (removing / changing fields) are covered in tests. There is `skip_nonitem_fields=True` argument exposed for the latter use case. https://github.com/scrapinghub/web-poet/pull/63 is a related PR which covers a remaining case (extra fields in the base class), but I'm not sure if that's a good pattern or not; I think it's better to discuss extra fields in the base classes separately..

The change is almost backwards compatible. I don't expect it to break any code, but it might break some type validation using mypy. There are 2 slight incompatibilities:

1. abstractmethod validation for to_item method is removed, because there is a default implementation now. It shouldn't affect user code.
2. to_item is defined as async. Code-wise, it shouldn't break user code, because existing ItemPage subclasses must implement their own to_item methods, and I can't imagine why they might call `super().to_item()`. It might break typing though, because mypy may issue an error (or a warning, not sure), if a regular `def to_item(self)` is defined in `ItemPage` subclass.

TODO:

* [x] decide if we should rename item_cls_fields parameter, and follow the decision 
* [x] API docs
* [x] Make sure Returns is in API docs
* [x] Rewrite fields tutorial
* [x] WebPage should supports fields. Should we deprecate ItemWebPage?
* [x] Check that other tutorials which mention ItemPage are valid


tests_typing/test_item_page.mypy-testing
```diff
@@ -0,0 +1,90 @@
...
+        price: float
+
+    class Subclass(BasePage, ItemPage[MyItem]):
+        @field
+        def price(self):
+            return 123
+
+    page = Subclass()
+    item = await page.to_item()
+    reveal_type(item)  # R: MyItem
```
kmike, line 90: This typing failure is my main concern with this PR. For some reason (maybe for a valid one), mypy thinks the type of item is going to be Item, not MyItem; that's why a test case is marked as xfail. 

What's interesting is that Pycharm thinks it's MyItem, so IDE works properly. 

I haven't checked yet if this should be considered a mypy bug or not, it needs a bit more exploration.

web_poet/pages.py
```diff
@@ -34,15 +36,33 @@ def is_injectable(cls: typing.Any) -> bool:
...
+        """Item class"""
+        param = get_generic_parameter(self.__class__)
+        if isinstance(param, typing.TypeVar):  # class is not parametrized
+            return dict  # type: ignore[return-value]
+        return param
+
+    async def to_item(self) -> ItemT:
         """Extract an item from a web page"""
+        return await item_from_fields(
+            self, item_cls=self.item_cls, item_cls_fields=self._skip_nonitem_fields
```
kmike, line 64: I wonder if it makes sense to rename `item_cls_fields` argument to `skip_nonitem_fields`

web_poet/pages.py
```diff
@@ -34,15 +36,33 @@ def is_injectable(cls: typing.Any) -> bool:
...
 
-    @abc.abstractmethod
-    def to_item(self):
+    _skip_nonitem_fields: bool
+
+    def __init_subclass__(cls, skip_nonitem_fields: bool = False, **kwargs):
+        super().__init_subclass__(**kwargs)
+        cls._skip_nonitem_fields = skip_nonitem_fields
+
+    @property
```
kmike, line 53: It seems this can be made a "class property", but it requires a bit more work

tests_typing/test_item_page.mypy-testing
kmike: The alternative I tried (https://github.com/scrapinghub/web-poet/pull/75) looks worse, because typing of to_item() method results doesn't work at all there. Here it works at least in some cases.

docs/advanced/fields.rst
```diff
@@ -314,63 +292,62 @@ to the item:
...
+    from my_library import FooPage, StandardItem
+    from web_poet import SetItemType, HttpResponse, field, ensure_awaitable
+
     @attrs.define
     class CustomItem(StandardItem):
         new_field: str
 
     @attrs.define
-    class CustomFooPage(FooPage):
+    class CustomFooPage(FooPage, SetItemType[CustomItem]):
```
kmike, line 304: Another naming idea:

```py
class CustomFooPage(FooPage, Returns[CustomItem]):
```

tox.ini
```diff
@@ -29,6 +29,15 @@ deps =
 
 commands = mypy web_poet tests
 
+[testenv:types]
+deps =
+    {[testenv]deps}
+    {[testenv:mypy]deps}
+    https://github.com/kmike/pytest-mypy-testing/archive/refs/heads/async-support.zip
```
kmike, line 36: it's needed because of https://github.com/davidfritzsche/pytest-mypy-testing/pull/31

web_poet/fields.py
```diff
@@ -73,33 +88,55 @@ def __get__(self, instance, owner=None):
...
+    obj, item_cls: Type[T] = dict, *, skip_nonitem_fields: bool = False  # type: ignore[assignment]
+) -> T:
     """Return an item of ``item_cls`` type, with its attributes populated
     from the ``obj`` methods decorated with :class:`field` decorator.
 
-    If ``item_cls_fields`` is True, ``@fields`` whose names don't match
-    any of the ``item_cls`` attributes are not passed to ``item_cls.__init__``.
-    When ``item_cls_fields`` is False (default), all ``@fields`` are passed
-    to ``item_cls.__init__``.
+    If ``skip_nonitem_fields`` is True, ``@fields`` whose names are not
```
kmike, line 111: I've renamed item_cls_fields to skip_nonitem_fields in this PR; does it look ok?

tests_typing/test_item_page.mypy-testing
BurnzZ: Interesting, it would seem **mypy** has still got some edge cases that are uncovered. I've tried looking deeper into this specific issue but to no avail. I'm wondering if we should also test this against other static type checkers like:
- https://github.com/microsoft/pyright
- https://github.com/facebook/pyre-check
- https://github.com/google/pytype

docs/intro/from-ground-up.rst
```diff
@@ -178,32 +187,40 @@ The same, but using web-poet
...
-  is created: ``BookPage(response=response)``.
-* ``BookPage`` inherits from :class:`~.WebPage` base class. This base class
-  is not doing much: it
+* instead of ``extract_book`` function we've got a ``BookPage`` class,
+  which inherits from :class:`~.ItemPage` base class, receives response
+  data in its ``__init__`` method, and returns the extracted item
+  in the ``to_item()`` method.
+
+Receiving a ``response`` argument in ``__init__`` is very common for page
+objects, so ``web-poet`` provides a shortcur for it: inherit from
```
BurnzZ, line 209:
```suggestion
objects, so ``web-poet`` provides a shortcut for it: inherit from
```

docs/intro/from-ground-up.rst
```diff
@@ -135,35 +141,38 @@ The same, but using web-poet
...
+    class BookPage(ItemPage):
         """
         A book page on http://books.toscrape.com website, e.g.
         http://books.toscrape.com/catalogue/a-light-in-the-attic_1000/index.html
         """
-        def extract_book(self):
+        def __init__(self, response: HttpResponse):
+            self.response = response
+
+        def to_item(self):
```
BurnzZ, line 162:
```suggestion
        def to_item(self) -> dict:
```

docs/intro/from-ground-up.rst
```diff
@@ -178,32 +187,40 @@ The same, but using web-poet
...
-     * defines ``__init__`` method which receives :class:`~.HttpResponse`, and
-     * provides shortcut methods like :meth:`~.WebPage.css`, which work by
-       creating :external:py:class:`parsel.selector.Selector` behind the scenes
-       (so that you don't need to create a selector in the ``extract_book`` method).
+.. code-block:: python
+
+    from web_poet import WebPage
+
+    class BookPage(WebPage):
+        def to_item(self):
```
BurnzZ, line 217:
```suggestion
        def to_item(self) -> dict:
```

docs/intro/from-ground-up.rst
```diff
@@ -215,41 +232,15 @@ to_item() method
 It is common to have Page Objects for a web page where a single main
```
BurnzZ, line 232: Should we also move this section much earlier? since the `to_item()` method has been already defined multiple times up top.

docs/intro/from-ground-up.rst
```diff
@@ -301,22 +292,35 @@ It might be easier to read the code written this way. Also, this style
...
-    in this style; see :ref:`web-poet-fields` .
-
-You may even write some base class to make it nicer - e.g. helper descriptors
-to define properties from CSS selectors, and a default ``to_item``
-implementation (so, no need to define ``to_item``).
-This is currently not implemented in ``web-poet``, but
-nothing prevents us from having a DSL like this:
+web-poet provides a small framework to simplify writing Page Objects
+in this style; see :ref:`web-poet-fields`. The example above can be simplified
+using web-poet fields - there is no need to write ``to_item`` boilerplate:
```
BurnzZ, line 297:
```suggestion
using web-poet :func:`web_poet.fields` - there is no need to write ``to_item`` boilerplate:
```

docs/intro/from-ground-up.rst
```diff
@@ -178,32 +187,40 @@ The same, but using web-poet
...
 
-* instead of ``extract_book`` function we got ``BookPage`` class,
-  which receives response data in its ``__init__`` method - see how it
-  is created: ``BookPage(response=response)``.
-* ``BookPage`` inherits from :class:`~.WebPage` base class. This base class
-  is not doing much: it
+* instead of ``extract_book`` function we've got a ``BookPage`` class,
+  which inherits from :class:`~.ItemPage` base class, receives response
+  data in its ``__init__`` method, and returns the extracted item
+  in the ``to_item()`` method.
```
Gallaecio, line 206:
```suggestion
* instead of the ``extract_book`` function we've got a ``BookPage`` class,
  which inherits from the :class:`~.ItemPage` base class, receives response
  data in its ``__init__`` method, and returns the extracted item
  in the ``to_item()`` method.
```

docs/intro/from-ground-up.rst
Gallaecio, line 209:
```suggestion
objects, so ``web-poet`` provides a shortcut for it: inherit from
```

docs/intro/from-ground-up.rst
```diff
@@ -301,22 +292,35 @@ It might be easier to read the code written this way. Also, this style
...
+        @field
+        def description(self):
+            return self.response.css("#product_description+ p").get().strip()
+
+        @field
+        def url(self):
+            return self.response.url
+
+.. note::
+    ``BookPage.to_item()`` method is ``async`` in the example above.
```
Gallaecio, line 322:
```suggestion
    The ``BookPage.to_item()`` method is ``async`` in the example above.
```

tests/test_pages.py
```diff
@@ -60,3 +72,100 @@ def to_item(self) -> dict:
...
+        @field
+        def name(self):
+            return "hello"
+
+        @field
+        def price(self):
+            return 123
+
+    # Item only contains "name", but not "price"
+    class Subclass(BasePage, Returns[Item], skip_nonitem_fields=True):
```
BurnzZ, line 164: Let's also add a test similar to this but `skip_nonitem_fields` is left out being `False` to assert that an exception is raised because of it.

docs/intro/from-ground-up.rst
```diff
@@ -708,10 +711,10 @@ Then, framework's role is to:
    (a common case is ``to_item``).
 
 For example, ``web-poet`` + Scrapy integration package (scrapy-poet_)
-may inspect a WebPage subclass you defined, figure out it needs
-:class:`~.HttpResponse` and nothing else, fetch scrapy's ``TextResponse``,
-create :class:`~.HttpResponse` instance from it, create your
-Page Object instance, and pass it to a spider callback.
+inspects a WebPage subclass you defined, figures out it needs
+:class:`~.HttpResponse` and nothing else, fetches scrapy's ``TextResponse``,
+creates :class:`~.HttpResponse` instance from it, creates your
```
Gallaecio, line 716:
```suggestion
:class:`~.HttpResponse` and nothing else, fetches Scrapy's ``TextResponse``,
creates an :class:`~.HttpResponse` instance from it, creates your
```

web_poet/fields.py
```diff
@@ -118,33 +96,47 @@ def get_fields_dict(cls_or_instance) -> Dict[str, FieldInfo]:
     return getattr(cls_or_instance, _FIELDS_INFO_ATTRIBUTE_READ, {})
 
 
-async def item_from_fields(obj, item_cls=dict, *, item_cls_fields=False):
+T = TypeVar("T")
```
BurnzZ, line 99: Should we utilize the same `ItemT` from `web_poet/pages.py`? as they're essentially the same

web_poet/pages.py
```diff
@@ -35,33 +37,47 @@ def is_injectable(cls: typing.Any) -> bool:
...
 
-class ItemPage(Injectable, abc.ABC):
-    """Base Page Object with a required :meth:`to_item` method.
-    Make sure you're creating Page Objects with ``to_item`` methods
-    if their main goal is to extract a single data record from a web page.
+ItemT = typing.TypeVar("ItemT")
+
+
+class Returns(typing.Generic[ItemT]):
+    """Inherit from this generic mixin to change item type used by
```
Gallaecio, line 44:
```suggestion
    """Inherit from this generic mixin to change the item type used by
```

docs/advanced/fields.rst
```diff
@@ -47,23 +47,15 @@ This approach has 2 main advantages:
...
----------------
-
+@field decorator
+----------------
 To aid writing Page Objects in this style, ``web-poet`` provides
-a few utilities:
-
-* :func:`@web_poet.field <web_poet.fields.field>` decorator,
-* :func:`web_poet.item_from_fields <web_poet.fields.item_from_fields>`
-  and :func:`web_poet.item_from_fields_sync <web_poet.fields.item_from_fields_sync>`
```
Gallaecio, line 58: It seems we are removing `item_from_fields_sync` from the documentation here. If so, should we remove the function altogether? Otherwise, maybe we should cover it somewhere in the documentation, even if briefly.

Also, should `item_from_fields` become a private function?

docs/intro/from-ground-up.rst
Gallaecio: [Already suggested](https://github.com/scrapinghub/web-poet/pull/74#discussion_r953426758).

docs/advanced/fields.rst
```diff
@@ -78,64 +70,51 @@ We can rewrite the example like this:
...
     import attrs
-    from web_poet import ItemPage, HttpResponse, HttpClient, field, item_from_fields
+    from web_poet import ItemPage, HttpResponse, HttpClient, field
 
 
     @attrs.define
     class MyPage(ItemPage):
         response: HttpResponse
-        http_client: HttpClient
+        http: HttpClient
```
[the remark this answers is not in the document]
BurnzZ: I think I like this shorter version as it could save 7 keystrokes. 😄 

Though if we change the naming convention, the other parts of the doc should also reflect it.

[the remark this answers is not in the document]
kmike: For me the main thing was its usage: `self.http.get(...)`, `self.http.post()` looks quite nice. But no strong opinion either, happy to revert it.

docs/advanced/fields.rst
kmike: They are currently covered in API docs for web_poet.fields. 
No strong opinon on keeping them public vs private; I wonder if they can be useful for implementing custom to_item methods.

docs/intro/from-ground-up.rst
kmike: web-poet fields are referred here as a concept, not as a module; I'm not sure it adding a link helps. The link doesn't look right (web_poet.fields is a module, not a function).

web_poet/fields.py
kmike: That's an interesting point, I have't considered it. Are there benefits in reusing typevars like this? 

In this particular case it seems there are scenarios when T can be not the same as ItemT, so they shouldn't be reused.

BurnzZ: No real benefit aside from ensuring that the types are consistent. But if some scenarios would result in diverging types, then we shouldn't reuse them. 👍

tests/test_pages.py
kmike: A good call, added more tests in https://github.com/scrapinghub/web-poet/pull/74/commits/98c2627d0b38ccd56c1c241661dee044db1eb0bc

docs/intro/from-ground-up.rst
kmike: it makes sense to add a note earlier; added it in https://github.com/scrapinghub/web-poet/pull/74/commits/28f917a3b7913c69ce3774e7d1d55105ebc6ff14. 

I'm not sure we can move the whole section earlier though. Currently the flow is this:

1. Show how code can be refactored without web-poet
2. Replicate the same thing with web-poet ("The same, but using web-poet")
3. Explain various concepts, starting from to_item.

We can't switch the sections, because web-poet is just introduced in a previous section. 
Maybe it's possible to write it all in some different way, but I'm out of ideas now :)

BurnzZ: I think that's fine. What do you think about adding a link pointing to the to_item() section instead to lead the read to more info? Though I think this would be a bit unnecessary since it would already be in the next section.

Gallaecio (approved):

docs/intro/from-ground-up.rst
kmike: I think it might be unnecessary; users should better keep reading, not jumping ahead :)

tests_typing/test_item_page.mypy-testing
kmike: I tried it with VS code + pylance, it agrees with mypy unfortunately. But I currently don't see a way around it, and alternatives look worse.

BurnzZ: I was also trying some of the workarounds mentioned in mypy's GH issue threads but couldn't make these particular examples work. Let's go with having them marked as xfail for now. 👍

BurnzZ (approved):
🚀

docs/advanced/fields.rst
```diff
@@ -314,63 +292,63 @@ to the item:
...
-To solve it, you can either
+One way to solve it is to make the orignal Page Object a dependency
+instead of inheriting from it, as explained in the beginning.
 
-* make the orignal Page Object a dependency instead of inheriting from it
-  (as explained in the beginning), or
-* use ``item_cls_fields=True`` argument of :func:`item_from_fields`:
-  when ``item_cls_fields`` parameter is True, ``@fields`` which
-  are not defined in the item are skipped.
+Alternatively, you can use ``skip_nonitem_fields=True`` class argument - it tells
```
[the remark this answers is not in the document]
kmike: I'm on fence about this suggestion :) On one hand, it makes sense. On the other hand, class arguments are so rarely used, so that thinking about default values of class arguments could be too much. I think for most users it'd be just "copy-paste ``skip_nonitem_fields=True``". No strong opinon on it though.

kmike:
Thanks for the review @Gallaecio @BurnzZ!

Issue #73, opened 2022-08-18
Fix mypy issues


BurnzZ:
Currently, a lot of the tests that we have don't have type annotations, which means the tests aren't scrutinized by **mypy**. Aside from annotating all of the test functions, we could set `check_untyped_defs = true` in the config.

In either way, we'd need to address all of the **mypy** issues that will be raised.

PR #78, opened 2022-08-30
ensure mypy validates tests

BurnzZ:
Resolves https://github.com/scrapinghub/web-poet/issues/73

Not too bad as I thought it would be since my initial tests were breaking due to `.mypy_cache`. 😅

### DIFF
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -6,7 +6,7 @@ from web_poet.page_inputs import HttpRequest, HttpResponse
 URL = "https://example.com"
 
 
-def test_http_error_init():
+def test_http_error_init() -> None:
     exc = HttpError()
     assert exc.request is None
     assert exc.args
@@ -16,7 +16,7 @@ def test_http_error_init():
     assert exc.request == request
 
 
-def test_http_request_error_init():
+def test_http_request_error_init() -> None:
     exc = HttpRequestError()
     assert exc.request is None
     assert exc.args
@@ -27,10 +27,10 @@ def test_http_request_error_init():
 
     response = HttpResponse(URL, b"")
     with pytest.raises(TypeError):
-        HttpRequestError(request=request, response=response)
+        HttpRequestError(request=request, response=response)  # type: ignore[call-arg]
 
 
-def test_http_response_error_init():
+def test_http_response_error_init() -> None:
     exc = HttpResponseError()
     assert exc.request is None
     assert exc.response is None
--- a/tests/test_overrides.py
+++ b/tests/test_overrides.py
@@ -11,7 +11,7 @@ from web_poet import OverrideRule, PageObjectRegistry, consume_modules, default_
 POS = {POTopLevel1, POTopLevel2, POModule, PONestedPkg, PONestedModule}
 
 
-def test_override_rule_uniqueness():
+def test_override_rule_uniqueness() -> None:
     """The same instance of an OverrideRule with the same attribute values should
     have the same hash identity.
     """
@@ -34,7 +34,7 @@ def test_override_rule_uniqueness():
     assert hash(rule1) == hash(rule2)
 
 
-def test_list_page_objects_all():
+def test_list_page_objects_all() -> None:
     rules = default_registry.get_overrides()
     page_objects = {po.use for po in rules}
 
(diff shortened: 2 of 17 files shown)


Gallaecio:
> my initial tests were breaking due to `.mypy_cache`.

Today I Learn that can happen :open_mouth:

tests/test_exceptions.py
```diff
@@ -27,10 +27,10 @@ def test_http_request_error_init():
 
     response = HttpResponse(URL, b"")
     with pytest.raises(TypeError):
-        HttpRequestError(request=request, response=response)
+        HttpRequestError(request=request, response=response)  # type: ignore[call-arg]
```
Gallaecio, line 30: This seems like a valid warning from mypy.

It seems HttpRequestError expects (msg, request) as parameters. Maybe we need to define `__init__` in HttpRequestError to allow an optional `response` parameter?

https://github.com/scrapinghub/web-poet/blob/master/web_poet/exceptions/http.py#L12-L40

tests/test_overrides.py
```diff
@@ -53,17 +53,17 @@ def test_list_page_objects_all():
     # registry's @handle_urls annotation was used.
     assert page_objects == POS.union({POLibSub})
     for rule in rules:
-        assert rule.instead_of == rule.use.expected_overrides, rule.use
-        assert rule.for_patterns == rule.use.expected_patterns, rule.use
-        assert rule.meta == rule.use.expected_meta, rule.use
+        assert rule.instead_of == rule.use.expected_overrides, rule.use  # type: ignore[attr-defined]
```
kmike, line 56: Why is it failing, is a missing type annotation somewhere? I.e. why isn't mypy aware that rules contains Rule instances?

tests/test_pages.py
```diff
@@ -20,7 +20,7 @@ class Item:
     name: str
 
 
-def test_page_object():
+def test_page_object() -> None:
```
kmike, line 23: By the way, I think there is nothing wrong with this opt-in approach using `-> None` annotations, it looks fine.

tests/test_requests.py
```diff
@@ -88,17 +89,21 @@ async def stub(req):
...
     client._request_downloader = client_with_status(500)
 
     method = getattr(client, method_name)
 
-    url_or_request = "url"
+    url_or_request: Union[str, HttpRequest] = "url"
     if method_name == "execute":
-        url_or_request = HttpRequest(url_or_request)
+        # Similar mypy issue as noted in
+        # https://github.com/zytedata/zyte-common-items/blob/main/tests/test_mypy.py#L37-L44
```
kmike, line 105: This link may become outdated if zyte-common-items code changes. Could you please either copy-paste some content from this link, or add a commit hash?

web_poet/page_inputs/client.py
```diff
@@ -19,7 +19,7 @@
 _StrMapping = Dict[str, str]
 _Headers = Union[_StrMapping, HttpRequestHeaders]
 _Body = Union[bytes, HttpRequestBody]
-_Status = Union[str, int]
+_Status = Union[str, int, List[Union[str, int]]]
```
kmike, line 22: What do you think about renaming it to _StatusList? It seems passing a list is what we should be promoting. Not sure why we support ints and strings in a first place :) List + a special case Literal("*") would be a bit cleaner (see https://docs.python.org/3/library/typing.html#typing.Literal). But that's out of scope for this PR.

tests/test_exceptions.py
kmike: Isn't this code checking a TypeError because of incorrect parameters, and typing fails because of that (as it should)?

kmike (approved):
Thanks @BurnzZ, nice work! I also expected way more failures :)

tests/test_exceptions.py
Gallaecio: Oh, right! :facepalm:

tests/test_overrides.py
BurnzZ: The exact error here is `"Callable[..., Any]" has no attribute "expected_overrides"  [attr-defined]` which is mainly due to the test definitions in https://github.com/scrapinghub/web-poet/blob/master/tests/po_lib/__init__.py#L15-L17.

Speaking of this error, I realized that we're labelling POs as `Callable`. I've updated the code to make it `Type[ItemPage]` instead. https://github.com/scrapinghub/web-poet/pull/78/commits/ceab08e855f46dcfa97a79974139680e4c071eec This ensures that we're promoting the overrides to be a subclass of `ItemPage`. I'm also not quite sure why we were labelling POs as `Callable` before. 🤔  What do you think of this change?

The `type: ignore` is still needed here since it's expecting `Type[ItemPage]` which doesn't contain class attributes like `expected_overrides` which are only used in the tests.

tests/test_pages.py
BurnzZ: Yeah, it would seem so. The remaining challenge here is to ensure that all of our tests has this annotation. One of the advantage of the `--check-untyped-defs` that I see is that it's able to catch any unannotated code or tests.

tests/test_requests.py
BurnzZ: Removed the link and copy-pasted the contents directly instead. https://github.com/scrapinghub/web-poet/pull/78/commits/62edbb9475a2ea6994bc5b6ab3140407de477e52

Gallaecio (approved):

Issue #64, opened 2022-08-16
BOM should take precedence over Content-Type header when detecting the encoding


kmike:
As explained in https://github.com/scrapy/w3lib/issues/189 and https://github.com/scrapy/scrapy/issues/5601, BOM should take a precedence over Content-Type headers when detecting an encoding.

Currently web-poet.HttpResponse prefers Content-Type header:

```py
import codecs
import web_poet

body = codecs.BOM + "Привет".encode('utf8')
headers = {"Content-Type": "text/html; charset=cp1251"}
resp = web_poet.HttpResponse(url="http://example.com", headers=headers, body=body, status=200)

print(resp.encoding) # cp1251, expected utf-8
print(resp.text) # яюРџСЂРёРІРµС‚ expected 'Привет'

```

PR #81, opened 2022-09-22
ensure that the BOM is used to sniff the encoding first

BurnzZ:
Resolves https://github.com/scrapinghub/web-poet/issues/64.

Based from the https://github.com/scrapy/scrapy/issues/5601 Scrapy fix.

### DIFF
--- a/web_poet/page_inputs/http.py
+++ b/web_poet/page_inputs/http.py
@@ -37,6 +37,10 @@ class HttpRequestBody(bytes):
 class HttpResponseBody(bytes):
     """A container for holding the raw HTTP response body in bytes format."""
 
+    def bom_encoding(self) -> Optional[str]:
+        """Returns the encoding from the byte order mark if present."""
+        return read_bom(self)[0]
+
     def declared_encoding(self) -> Optional[str]:
         """Return the encoding specified in meta tags in the html body,
         or ``None`` if no suitable encoding was found"""
@@ -238,7 +242,7 @@ class HttpResponse(SelectableMixin):
         """Encoding of the response"""
         return (
             self._encoding
-            or self._bom_encoding()
+            or self._body_bom_encoding()
             or self._headers_declared_encoding()
             or self._body_declared_encoding()
             or self._body_inferred_encoding()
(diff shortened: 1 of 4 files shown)


Gallaecio (approved):

web_poet/page_inputs/http.py
```diff
@@ -261,6 +263,10 @@ def urljoin(self, url: Union[str, _RequestUrl, _ResponseUrl]) -> _RequestUrl:
         *self*."""
         return _RequestUrl(urljoin(self._base_url, str(url)))
 
+    @memoizemethod_noargs
+    def _bom_encoding(self):
+        return read_bom(self.body)[0]
```
kmike, line 268: I wonder if this code should be moved to HttpResponseBody class

BurnzZ: Good call, moved it in https://github.com/scrapinghub/web-poet/pull/81/commits/c39d176050229d4ccb3bbf639ad252e5943fa23e.

web_poet/page_inputs/http.py
```diff
@@ -36,12 +37,16 @@ class HttpRequestBody(bytes):
...
+        """Returns the encoding from the byte order mark if present."""
+        return read_bom(self)[0]
+
     def declared_encoding(self) -> Optional[str]:
         """Return the encoding specified in meta tags in the html body,
         or ``None`` if no suitable encoding was found"""
         return html_body_declared_encoding(self)
 
-    def json(self):
+    def json(self) -> Optional[str]:
```
kmike, line 49: This type annotation is wrong, any object can be returned, as it's result of json.loads

web_poet/page_inputs/http.py
```diff
@@ -233,17 +238,18 @@ def _selector_input(self) -> str:
...
             self._encoding
+            or self._body_bom_encoding()
             or self._headers_declared_encoding()
             or self._body_declared_encoding()
             or self._body_inferred_encoding()
         )
 
     @memoizemethod_noargs
-    def json(self):
+    def json(self) -> Optional[str]:
```
kmike, line 252: the same, it can return anything :)

web_poet/page_inputs/http.py
BurnzZ: Nice catch! I was on a copy-pasting spree on the past commit. 🤦 

Although I'm curious as to why mypy wasn't able to detect this one as the tests were comparing it to a `dict`.

web_poet/page_inputs/http.py
```diff
@@ -46,7 +46,7 @@ def declared_encoding(self) -> Optional[str]:
         or ``None`` if no suitable encoding was found"""
         return html_body_declared_encoding(self)
 
-    def json(self) -> Optional[str]:
+    def json(self) -> Optional[Any]:
```
kmike, line 49: Optional[Any] is the same as Any :)

kmike:
Thanks @BurnzZ!

PR #82, opened 2022-09-27
add test for @field with other decorators

BurnzZ:
Used as an example in https://github.com/zytedata/zyte-common-items/pull/15#issuecomment-1259094848


kmike:
Thanks @BurnzZ!

Issue #77, opened 2022-08-26
"Apply rules" instead of overrides


kmike:
hey! I’m exploring how to make the following work:

```py
from zyte_common_items import Product
from web_poet import ItemPage, handle_urls


@handle_urls("example.com")
class ExamplePage(ItemPage[Product]):
    # ...
```

In other words, how to make `@handle_urls("example.com")` work for an `ItemPage[Product]` subclass without a need to use `instead_of` in handle_urls, and without a need to use a base page object for `instead_of`.

I can see 2 main approaches here.

# Approach 1: support is directly

In the example, handle_urls doesn't really define any override rule. Instead, we have a declaration that ExamplePage can return Product item for example.com page. This information should be enough to allow creation of a `scrapy-poet` provider for items:

```py
def parse(self, response: DummyResponse, product: Product):
    # ...
```

We know the website, we know which item is needed, and can use Page Object registry to find a right page object, according to domain and priority.

To implement it, web-poet needs to be slightly refactored:
 
1. We should rename “overrides”, “override rule” to something like “apply rules”. This includes changes to class names, module names, method names, and changes to documentation as well. In addition to the old `ApplyRule(for_patterns=..., use=ExampleProductPage, instead_of=GenericProductPage)`, it’d be possible to specify `ApplyRule(for_patterns=..., use=ExampleProductPage, to_return=Product)`
2. `handle_urls` decorator would pick up `to_return=Product` from the `ItemPage[Product]` automatically (probably unless instead_of argument is defined? but that's not clear, there are use cases for supporting both at the same time).

When implementing it, we should make sure that priorities work as intended. For example, it should be possible to define and configure a Page Object which provides Product instances using AutoExtract API, and have it behaving in a sensible way (enabled for all domains, custom Page Objects should take priority over this default one).

# Approach 2: define standard generic Page Object classes

In this case, "override rules" stay override rules. There is a registry of `{item_type: generic_page_object_class}` defined somewhere, e.g. `{Product: ProductPage}`. Or, maybe, `{Product: [ProductPage, AnotherGenericProductPage]}`. There should be an API to extend and modify this registry. handle_urls looks up this registry, and picks `instead_of` argument based on it.

# Pros/cons
Overall, I like an approach with items more, because it seems to provide a nicer API: 

* less indirection;
* enables a use of items as dependencies in callbacks;
* only requires standardization of item classes, not base page objects.

[...]

PR #83, opened 2022-09-28
Add more test cases for subclasses using @field

BurnzZ:
Adding and fixing some use cases before **Approach 1** from https://github.com/scrapinghub/web-poet/issues/77 begins development.


kmike:
@BurnzZ should we merge or close the PR?

BurnzZ:
@kmike , I forgot about this, thanks for reminding. I think we could still merge it after reviewing.

PR #84, opened 2022-09-28
@handle_urls() with item return type

BurnzZ:
Addresses **approach 1** of https://github.com/scrapinghub/web-poet/issues/77

# TODO:
- [x] Documentation overhaul
    - [x] docstrings
    - [x] tutorials
        - [x] overriding `to_return` item using `@handle_urls`
- [x] Consider renaming `OverrideRule` to something else (candidate: `ApplyRule`)
- [x] ~Consider renaming `OverrideRule.instead_of` to something else~ (hopefully to match the `@handle_urls` parameter). Or the other way around: renaming `@handle_urls`'s `overrides` parameter.
- [x] Consider renaming `handle_urls`'s `data_type` to something else (ensure that it's consistent with `OverrideRule` as well)
- [x] Consistent use of the term "item type" vs "data type" vs "item class" in docstrings, variables, and parameters.
- [x] ~Handling the case of not needing all fields (recommendation for frameworks like `scrapy-poet`)~


kmike:
hey @BurnzZ! I think we can solve "Handling the case of not needing all fields (recommendation for frameworks like scrapy-poet)" separately.

web_poet/overrides.py
```diff
@@ -37,6 +38,7 @@ class OverrideRule:
           about the patterns.
         * ``use`` - The Page Object that will be **used**.
         * ``instead_of`` - The Page Object that will be **replaced**.
+        * ``data_type`` - The data container class that the Page Object returns.
```
kmike, line 41: I think that's fine to have different argument names in ApplyRule and handle_urls.

handle_urls parameters tell you something about the decorated page object:

```py
# MyPage handles "example.com" urls and overrides GenericProductPage:
@handle_urls("example.com", overrides=GenericProductPage)
    # ...
```

ApplyRule attributes describe the rule:

```py
# for example.com website use MyPage instead of GenericProductPage
ApplyRule(for_patterns=Patterns(["example.com"]), use=Mypage, instead_of=GenericProductPage)
```

Following this logic, we could do something like this:

```py
# MyPage handles "example.com" urls and can be used to provide Product items:
@handle_urls("example.com", provides=Product)  # provides should rarely be needed
    # ...


# for example.com website use MyPage to provide Product items
ApplyRule(for_patterns=Patterns(["example.com"]), use=MyPage, to_provide=Product)
```

In #77 it was `to_return` instead of `to_provide`; to_provide seems better.

---

After re-reading it though, the following example doesn't look too bad if we make argument names the same:

```py
# MyPage handles "example.com" instead of GenericProductPage:
@handle_urls("example.com", instead_of=GenericProductPage)
    # ...

# MyPage handles "example.com" in order to provide Product items
@handle_urls("example.com", to_provide=Product)
    # ...

```

kmike: What do you think?

BurnzZ: +1 on having `@handle_urls()` utilize `instead_of`. This term is much more clear about the intent compared to the old `overrides`. The old `overrides` param speaks more about the a specific concept used in **web-poet**. The `instead_of` uses plain language which is direct. It also helps that the rule which `@handle_urls()` produce is consistent with its parameters.

I find that this is a good opportunity to improve the mismatches of parameters:

```python
# Old one
@handle_urls("example.com", overrides=GenericProductPage)
    ...

ApplyRule(for_patterns=Patterns(["example.com"]), use=Mypage, instead_of=GenericProductPage)
```

From the old one above, there might be some confusion to newcomers regarding what happened to the class they passed to `overrides`. Is there an underlying logic which sets `overrides` apart from `instead_of`? Are they same?

This new one is more consistent which could help remove those types of confusions:

```python
# Proposed renaming
@handle_urls("example.com", instead_of=GenericProductPage)
    ...

ApplyRule(for_patterns=Patterns(["example.com"]), use=Mypage, instead_of=GenericProductPage)
```

---

On the other hand, I find `to_return` to be slightly better than `to_provide`. It also helps that it's consistent in both `@handle_urls` and `ApplyRule`:

```python
@handle_urls("example.com", to_return=Product) 
    ...

ApplyRule(for_patterns=Patterns(["example.com"]), use=MyPage, to_return=Product)
```

The `*_return` terminology is consistent with the behavior of the Page Object that it returns something. I think `*_provide` could also work. Although I think when it's used in the context of **scrapy-poet**, there might be some confusion that the _rule_ or _page object_ itself is the provider.

I might be missing something. What was the main benefits you see in using `to_provide` instead of `to_return`?

kmike: I'm fine with using `to_return` name, it sounds fine indeed. `to_provide` may require more knowledge from the users, and so it could be less clear.

BurnzZ: Updated these. Thanks @kmike !

web_poet/__init__.py
```diff
@@ -1,5 +1,5 @@
 from .fields import field, item_from_fields, item_from_fields_sync
-from .overrides import OverrideRule, PageObjectRegistry, consume_modules
+from .overrides import ApplyRule, OverrideRule, PageObjectRegistry, consume_modules
```
kmike, line 2: What do you think about renaming the module? e.g. rules.py?

web_poet/overrides.py
```diff
@@ -166,8 +200,8 @@ def wrapper(cls):
 
         return wrapper
 
-    def get_overrides(self) -> List[OverrideRule]:
-        """Returns all of the :class:`~.OverrideRule` that were declared using
+    def get_overrides(self) -> List[ApplyRule]:
+        """Returns all of the :class:`~.ApplyRule` that were declared using
```
BurnzZ, line 204: @kmike , in line with `OverrideRule` becoming `ApplyRule`, I was thinking if we should also rename the following:
- `get_overrides` to `get_rules`
- `search_overrides` to `search_rules`

I think "rules" is a more apt terminology with the introduction of the "to_return" since it means that "overrides/instead_of" can now be set to `None`. For example, `ApplyRule(for_patterns=Patterns(include=['example.com'], exclude=[], priority=500), use=CustomProductPage, to_return=Product)` loses the concept of overrides.

Do you see some possible divergence of the concepts if we do rename them?

web_poet/__init__.py
BurnzZ: I was thinking the same thing! 👍

web_poet/overrides.py
```diff
@@ -141,22 +162,35 @@ def handle_urls(
         """
 
         def wrapper(cls):
-            rule = OverrideRule(
+            final_data_type = to_return
+            derived_type = get_generic_parameter(cls)
+            if derived_type and derived_type != ItemT:
+                final_data_type = derived_type
```
kmike, line 168: I think we have the code which does the same in `Returns.item_cls`. It'd be good to have only one copy. I can see two ways to achieve it:

1) Make Returns.item_cls a class property (could be tricky), or expose another class method for it.
2) Extract this code to an utility function.

web_poet/_typing.py
```diff
@@ -18,7 +18,7 @@ def is_generic_alias(obj) -> bool:
 
 
 def get_generic_parameter(cls):
-    for base in cls.__orig_bases__:
+    for base in getattr(cls, "__orig_bases__", []):
```
kmike, line 21: Have you encountered an issue with `cls.__orig__bases__`?

web_poet/overrides.py
kmike: This makes sense 👍

web_poet/__init__.py
BurnzZ: Done in https://github.com/scrapinghub/web-poet/pull/84/commits/01112e7e316ea2756a8692b752d757aeae280f23.

web_poet/_typing.py
BurnzZ: Yes, it doesn't work on cases like:

```python
@handle_urls("example.com", to_return=Product)
class CustomProductPageDataTypeOnly(web_poet.Injectable):
    ...
```

web_poet/overrides.py
BurnzZ: Exposing it into another classmethod won't be usable for the following use case since it's not inheriting from `Returns`:
```python
@handle_urls("example.com", to_return=Product)
class CustomProductPageDataTypeOnly(web_poet.Injectable):
    ...
```

Assuming that we're able to extract it into another utility function like:
```python
def get_item_cls(cls) -> typing.Type[ItemT]:
    param = get_generic_parameter(cls)
    if isinstance(param, typing.TypeVar):
        return dict
    return param
```

This code-block would still be more or less the same:
```python
final_data_type = to_return
derived_type = get_item_cls(cls)
if derived_type and derived_type != dict:  # the only diff is we're checking against a dict
    final_data_type = derived_type
```

I believe we've already reduced this one to its core.

web_poet/overrides.py
```diff
@@ -1,313 +1,4 @@
...
-        gen = _walk_module(module)
-
-        # Inspired by itertools recipe: https://docs.python.org/3/library/itertools.html
-        # Using a deque() results in a tiny bit performance improvement that list().
-        deque(gen, maxlen=0)
-
 
-OverrideRule = _create_deprecated_class("OverrideRule", ApplyRule, warn_once=False)
+msg = "The 'web_poet.overrides' module has been moved into 'web_poet.rules'."
+warnings.warn(msg, DeprecationWarning, stacklevel=2)
```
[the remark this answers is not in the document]
BurnzZ: I missed that one, great catch!

web_poet/overrides.py
BurnzZ: I now see what you mean, thanks for explaining it further! Updated this in https://github.com/scrapinghub/web-poet/pull/84/commits/144e39ef36d5db891686bae8137f723ba4e1437f

> The idea is that to_return passed explicitly to the decorator should take priority over the item class defined by the page object.
I currently had it the other way around. Nice catch!

web_poet/rules.py
```diff
@@ -0,0 +1,324 @@
...
+          do anything for now but may be useful for future API updates.
+
+    .. tip::
+
+        The :class:`~.ApplyRule` is also hashable. This makes it easy to store
+        unique rules and identify any duplicates.
+    """
+
+    for_patterns: Patterns = attrs.field(converter=str_to_pattern)
+    use: Type[ItemPage]
```
kmike, line 52: What do you think about `use` argument keyword-only?

kmike: As well as all other arguments, except for `for_patterns`?

kmike: It won't be backwards compatible (unless we make fix it somehow), but it may be ok, we might just document it in the changelog.

BurnzZ: That's a good idea! I believe kwargs enforces better readability in PO codebases especially for distinguishing between `use` and `instead_of`. https://github.com/scrapinghub/web-poet/pull/84/commits/9b6f9c4d836b0335850c076e1eed66374d18ecc6

BurnzZ:
I think this is ready for a proper review since the new API has been more or less validated in **scrapy-poet**'s test cases: https://github.com/scrapinghub/scrapy-poet/pull/88/files#diff-a5ce1f1bde36b639be5be17fed1ed35ffb018c1fefcf4544c0e3fac120ca58eb

I'm thinking of moving the task of "splitting the documentation between devs creating POs vs devs creating frameworks for web-poet" into another PR to avoid a lot of diffs here.

CHANGELOG.rst
```diff
@@ -2,6 +2,43 @@
...
+
+    * It's now accepting a ``to_return`` parameter which signifies the data
+      container class that the Page Object returns.
+    * Passing a string to ``for_patterns`` would auto-convert it into
+      ``url_matcher.Patterns``.
+    * All arguments are now keyword-only except for ``for_patterns``.
+
+* Modify the call signature of ``handle_urls``:
+
+    * New ``to_return`` parameter which signifies the data container class that
```
kmike, line 19: I think we should downplay it a bit, because most users shouldn't use to_return argument directly. What do you think about documenting it this way? 

1. handle_urls now sets rule.to_return to the item type declared in the ItemPage
2. for advanced use cases, it's possible to change it by using `to_return` argument of `handle_urls` decorator

CHANGELOG.rst
```diff
@@ -2,6 +2,43 @@
...
+* The ``overrides`` parameter from ``@handle_urls`` is now deprecated.
+  Use the ``instead_of`` parameter instead.
+* The ``OverrideRule`` class is now deprecated. Use ``ApplyRule`` instead.
+* The ``from_override_rules`` method of ``PageObjectRegistry`` is now deprecated.
+  Use ``from_apply_rules`` instead.
+* The ``web_poet.overrides`` module is deprecated. Use ``web_poet.rules`` instead.
+* The ``PageObjectRegistry.get_overrides`` is deprecated.
+  Use ``PageObjectRegistry.get_rules`` instead.
+* The ``PageObjectRegistry.search_overrides`` is deprecated.
+  Use ``PageObjectRegistry.search_rules`` instead.
```
Gallaecio, line 40:
```suggestion
* The ``PageObjectRegistry.get_overrides`` method is deprecated.
  Use ``PageObjectRegistry.get_rules`` instead.
* The ``PageObjectRegistry.search_overrides`` method is deprecated.
  Use ``PageObjectRegistry.search_rules`` instead.
```

docs/intro/overrides.rst
```diff
@@ -3,13 +3,17 @@
...
+`url-matcher <https://url-matcher.readthedocs.io>`_.
 
 Using such rules establishes the core concept of Overrides wherein a developer
-could declare that a specific Page Object must be used *(instead of another)*
-for a given set of URL patterns.
+could declare that for a given set of URL patterns a specific Page Object must
+be used:
+
+    * instead of another Page Object, or 
+    * which returns an Item Class instead of some other Item Class.
```
kmike, line 16: It doesn't need to be insted of some other Item Class

docs/intro/overrides.rst
```diff
@@ -3,13 +3,17 @@
...
+matching rules is handled by another library called
+`url-matcher <https://url-matcher.readthedocs.io>`_.
 
 Using such rules establishes the core concept of Overrides wherein a developer
-could declare that a specific Page Object must be used *(instead of another)*
-for a given set of URL patterns.
+could declare that for a given set of URL patterns a specific Page Object must
+be used:
+
+    * instead of another Page Object, or
```
kmike, line 15: I think only this use case counts as "Override"

docs/advanced/fields.rst
```diff
@@ -151,6 +151,8 @@ function when accessing the fields:
 Now any field can be converted from sync to async, or the other way around,
 and the code would keep working.
 
+.. _`item-classes`:
```
Gallaecio, line 154: :lipstick: I did not know that they worked, but those quote marks are not needed here.

```suggestion
.. _item-classes:
```

docs/intro/overrides.rst
```diff
@@ -100,42 +129,156 @@ The code above declares that:
...
+    @handle_urls("anotherexample.com", to_return=Product, exclude="/digital-goods/")
+    class AnotherExampleProductPage(WebPage):
+        def to_item(self) -> Product:
+            ...  # more specific parsing
+
+
+    @handle_urls(["dualexample.com/shop/?product=*", "dualexample.net/store/?pid=*"], to_return=Product)
+    class DualExampleProductPage(WebPage):
+        def to_item(self) -> SimilarProduct:
+            ...  # more specific parsing
```
kmike, line 169: I think the main and promoted way of using this feature should be this:

```py
    @handle_urls("example.com")
    class ExampleProductPage(WebPage[Product]):
        def to_item(self) -> Product:
            ...  # more specific parsing
```

Using to_return argument directly is only for advanced use cases; it should be discouraged.

pyproject.toml
```diff
@@ -6,3 +6,7 @@ multi_line_output = 3
 show_error_codes = true
 ignore_missing_imports = true
 no_warn_no_return = true
+
+[[tool.mypy.overrides]]
+module = "tests.po_lib_to_return.*"
+ignore_errors = true
```
kmike, line 12: Could you please add a comment, explaining why these errors are ignored?

docs/intro/overrides.rst
```diff
@@ -3,13 +3,17 @@
 Overrides
 =========
 
-Overrides contains mapping rules to associate which URLs a particular
-Page Object would be used. The URL matching rules is handled by another library
-called `url-matcher <https://url-matcher.readthedocs.io>`_.
+Overrides are simply rules represented by a list of :class:`~.ApplyRule` which
```
Gallaecio, line 6: :lipstick: 

```suggestion
Overrides are rules represented by a list of :class:`~.ApplyRule` which
```

tests/po_lib/__init__.py
```diff
@@ -25,19 +26,22 @@ class POTopLevelOverriden2(ItemPage):
     ...
 
 
-# This first annotation is ignored. A single annotation per registry is allowed
-@handle_urls("example.com", overrides=POTopLevelOverriden1)
+# This first decorator is ignored. A single ``ApplyRule`` with the same Page
+# Object to be used per registry is allowed.
```
kmike, line 30: Hm, I wonder if we should lift this limitation. Consider this:

```py
# MyPage is registered as a page which returns Product items
@handle_urls("example.com")
class MyPage(WebPage[Product]):
    # ...

# and later: register the same page to be used instead of some page object
handles_urls(instead_of=GenericProductPage)(MyPage)
```

I think this is out of scope for this PR though; there are some detail we'd need to figure out to make it work.

docs/intro/overrides.rst
```diff
@@ -50,6 +54,28 @@ Let's see this in action by declaring the Overrides in the Page Objects below.
 Creating Overrides
 ------------------
 
+To simplify the code examples in the next few subsections, let's consider that
+these Item Classes has been predefined.
```
Gallaecio, line 58: :lipstick: 

```suggestion
these Item Classes has been predefined:
```

docs/intro/overrides.rst
```diff
@@ -58,29 +84,32 @@ Let's take a look at how the following code is structured:
...
-    @handle_urls(["dualexample.com/shop/?product=*", "dualexample.net/store/?pid=*"], overrides=GenericProductPage)
+    @handle_urls(["dualexample.com/shop/?product=*", "dualexample.net/store/?pid=*"], instead_of=GenericProductPage)
     class DualExampleProductPage(WebPage):
-        def to_item(self):
+        def to_item(self) -> SimilarProduct:
             ...  # more specific parsing
 
 The code above declares that:
 
+    - Page Objects return ``Product`` and ``SimilarProduct`` item class. Returning
```
Gallaecio, line 110:
```suggestion
    - Page Objects return ``Product`` and ``SimilarProduct`` item classes. Returning
```

docs/intro/overrides.rst
```diff
@@ -100,42 +129,156 @@ The code above declares that:
...
+            ...  # more specific parsing
+
+
+    @handle_urls("anotherexample.com", to_return=Product, exclude="/digital-goods/")
+    class AnotherExampleProductPage(WebPage):
+        def to_item(self) -> Product:
+            ...  # more specific parsing
+
+
+    @handle_urls(["dualexample.com/shop/?product=*", "dualexample.net/store/?pid=*"], to_return=Product)
```
Gallaecio, line 166: example.com is a special domain for documentation purposes, anotherexample.com and dualexample.com are not (e.g. they could be registered and point to NSFW content in the future, making it akward that we have them here). You could instead use the `.example` top-level domain, e.g. `some.example`, `another.example`, `dual.example`.

docs/intro/overrides.rst
```diff
@@ -100,42 +129,156 @@ The code above declares that:
...
+    class DualExampleProductPage(WebPage, Returns[SimilarProduct]):
+        def to_item(self) -> SimilarProduct:
+            ...  # more specific parsing
+
+Notice that the ``@handle_urls`` decorator for ``AnotherExampleProductPage`` and
+``DualExampleProductPage`` doesn't specify the ``to_return`` parameter at all.
+This is because it's able to directly derive the returned Item Class from the
+respective ``Returns[Product]`` and ``Returns[SimilarProduct]`` declarations.
+
+See the next :ref:`retrieving-overrides` section observe what are the actual
```
Gallaecio, line 241:
```suggestion
See the next :ref:`retrieving-overrides` section to observe what are the actual
```

docs/intro/overrides.rst
```diff
@@ -211,16 +354,16 @@ This can be done something like:
 
 .. _`intro-rule-subset`:
 
-Using only a subset of the available OverrideRules
+Using only a subset of the available ApplyRules
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Gallaecio, line 358:
```suggestion
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

tests/po_lib_to_return/__init__.py
```diff
@@ -0,0 +1,174 @@
...
+    price: float
+
+
+@attrs.define
+class ProductMoreFields(Product):
+    brand: str
+
+
+@attrs.define
+class ProductLessFields:
```
Gallaecio, line 25: :smile: 

```suggestion
class ProductFewerFields:
```

Gallaecio (commented):
Great work!

tests/test_rules.py
```diff
@@ -0,0 +1,273 @@
...
+        assert rule.meta == rule.use.expected_meta, rule.use  # type: ignore[attr-defined]
+
+
+def test_registry_get_overrides_deprecation() -> None:
+    msg = "The 'get_overrides' method is deprecated. Use 'get_rules' instead."
+    with pytest.warns(DeprecationWarning, match=msg):
+        rules = default_registry.get_overrides()
+
+    # It should still work as usual
+    assert len(rules) == 15
```
kmike, line 174:
```suggestion
    assert len(rules) == len(default_registry.get_rules())
```

CHANGELOG.rst
BurnzZ: Makes sense. Updated it in https://github.com/scrapinghub/web-poet/pull/84/commits/8efa813ca6cb8a396beafd552d0be01f4ca6a278.

tests/test_rules.py
```diff
@@ -0,0 +1,273 @@
...
+def test_consume_module_not_existing() -> None:
+    with pytest.raises(ImportError):
+        consume_modules("this_does_not_exist")
+
+
+def test_list_page_objects_all_consume() -> None:
+    """A test similar to the one above but calls ``consume_modules()`` to properly
+    load the ``@handle_urls`` decorators from other modules/packages.
+    """
+    consume_modules("tests_extra")
```
kmike, line 186: The order of tests might really matter here; if consume_modules is called, or tests are imported in some other way, it affects all other tests.

tests/test_rules.py
```diff
@@ -0,0 +1,273 @@
...
+    page_objects = {po.use for po in rules}
+    assert any(["po_lib_sub_not_imported" in po.__module__ for po in page_objects])
+
+
+def test_registry_search_rules() -> None:
+    rules = default_registry.search_rules(use=POTopLevel2)
+    assert len(rules) == 1
+    assert rules[0].use == POTopLevel2
+
+    rules = default_registry.search_rules(instead_of=POTopLevelOverriden2)
```
kmike, line 197: Could you please add a test where search_rules is used 

1) with to_return argument?
2) with more than 1 search argument?

tests/test_rules.py
```diff
@@ -0,0 +1,273 @@
...
+
+    with warnings.catch_warnings(record=True) as w:
+
+        @handle_urls("example.com", overrides=CustomProductPage)
+        class PageWithDeprecatedOverrides:
+            ...
+
+    w = [x for x in w if x.category is DeprecationWarning]
+    assert len(w) == 1
+    assert str(w[0].message) == (
```
kmike, line 261: Can `with pytest.warns` be used to simplify it?

web_poet/_typing.py
```diff
@@ -18,7 +18,16 @@ def is_generic_alias(obj) -> bool:
...
         if is_generic_alias(base):
             args = _get_args(base)
             return args[0]
+
+
+def get_item_cls(cls, default=None, preferred=None):
+    if preferred:
+        return preferred
```
kmike, line 29: I'd remove the `preferred` argument, and handle it on a caller side, but that's very minor; the current code is fine as well.

web_poet/rules.py
```diff
@@ -0,0 +1,338 @@
...
+          this rule. You can read the API documentation of the `url-matcher
+          <https://url-matcher.readthedocs.io/>`_ package for more information
+          about the patterns.
+        * ``use`` - The Page Object that will be **used** in cases where the URL
+          pattern represented by the ``for_patterns`` attribute is matched.
+        * ``instead_of`` - *(optional)* The Page Object that will be **replaced**
+          with the Page Object specified via the ``use`` parameter.
+        * ``to_return`` - *(optional)* The data container class that will be
+          **replaced** by the item returned by the Page Object specified via the
+          ``use`` parameter.
```
kmike, line 45: I think we should try to find a better wording. Unlike Page Objects, we're not replacing anything here. The idea is to mark a Page Object as something which can return an item of a certain type. Moreover, `to_return` is for an advanced use case where this declaration needs to be different from the item type which is really returned by the Page Object; most users should rely on automatic behavior.

web_poet/rules.py
```diff
@@ -0,0 +1,338 @@
...
+        * ``meta`` - *(optional)* Any other information you may want to store.
+          This doesn't do anything for now but may be useful for future API updates.
+
+    The main functionality of this class lies in the ``instead_of`` and ``to_return``
+    parameter. Should both of these be omitted, then :class:`~.ApplyRule` simply
+    tags which URL patterns the given Page Object is expected to be used.
```
kmike, line 51: Not sure that's valid; if to_return is not present, then ApplyRule should be using ItemPage's return class.

web_poet/rules.py
```diff
@@ -0,0 +1,338 @@
...
+
+    def __hash__(self):
+        return hash((self.for_patterns, self.use, self.instead_of, self.to_return))
+
+
+class PageObjectRegistry(dict):
+    """This contains the :class:`~.ApplyRule` that associates the Page Objects
+    alongside its Items for a given URL matching rule.
+
+    Note that it's simply a ``dict`` subclass with added functionalities on
```
kmike, line 75:
```suggestion
    PageObjectRegistry is a ``dict`` subclass with added functionalities on
```

web_poet/rules.py
```diff
@@ -0,0 +1,338 @@
...
+        overrides: Optional[Type[ItemPage]] = None,
+        instead_of: Optional[Type[ItemPage]] = None,
+        to_return: Optional[Type] = None,
+        exclude: Optional[Strings] = None,
+        priority: int = 500,
+        **kwargs,
+    ):
+        """
+        Class decorator that indicates that the decorated Page Object should be
+        used instead of the overridden one for a particular set the URLs.
```
kmike, line 150: I think we should focus on to_return a bit more here. The main purpose is not to declare that this page object could be used on a certain set of URLs. Overrides is optional now.

web_poet/rules.py
```diff
@@ -0,0 +1,338 @@
...
+
+        The Page Object that is **overridden** is declared using the ``instead_of``
+        parameter.
+
+        The Item Class could also be **overridden** using the ``to_return``
```
kmike, line 155: I'd not call it "overridden"; it's actually overridden if to_return is passed, but it's a very different type of override - to_return argument overrides the item type defined in the page object class. Using the same terminology could be a bit confusing.

web_poet/rules.py
```diff
@@ -0,0 +1,338 @@
...
+        """
+
+        def wrapper(cls):
+
+            if overrides is not None:
+                msg = (
+                    "The 'overrides' parameter in @handle_urls is deprecated. "
+                    "Use the 'instead_of' parameter."
+                )
+                warnings.warn(msg, DeprecationWarning, stacklevel=2)
```
kmike, line 185: Could you please check that this stacklevel points to the user code? I can't wrap my head around it, but suspect stacklevel=3 could be needed here :)

web_poet/rules.py
```diff
@@ -0,0 +1,338 @@
...
+
+            return cls
+
+        return wrapper
+
+    def get_rules(self) -> List[ApplyRule]:
+        """Returns all the :class:`~.ApplyRule` that were declared using
+        the ``@handle_urls`` decorator.
+
+        .. warning::
```
kmike, line 217:
```suggestion
        .. note::
```

web_poet/rules.py
```diff
@@ -0,0 +1,338 @@
...
+        return list(self.values())
+
+    def get_overrides(self) -> List[ApplyRule]:
+        """Deprecated, use :meth:`~.PageObjectRegistry.get_rules` instead."""
+        msg = "The 'get_overrides' method is deprecated. Use 'get_rules' instead."
+        warnings.warn(msg, DeprecationWarning, stacklevel=2)
+        return self.get_rules()
+
+    def search_rules(self, **kwargs) -> List[ApplyRule]:
+        """Returns any :class:`ApplyRule` that has any of its attributes
```
kmike, line 232:
```suggestion
        """Return any :class:`ApplyRule` that has any of its attributes
```

docs/intro/overrides.rst
BurnzZ: Good catch! I was fixated on one specific usecase on this. Updated in https://github.com/scrapinghub/web-poet/pull/84/commits/52a6f474bd97d1626eaf153241496a7a78c40df6

docs/intro/overrides.rst
BurnzZ: Updated in https://github.com/scrapinghub/web-poet/pull/84/commits/52a6f474bd97d1626eaf153241496a7a78c40df6 where the focus shifts into matching the URL pattern and `to_return` class type.

docs/intro/overrides.rst
BurnzZ: Updated in https://github.com/scrapinghub/web-poet/pull/84/commits/52a6f474bd97d1626eaf153241496a7a78c40df6 where it de-emphasizes this approach

tests/po_lib/__init__.py
BurnzZ: Yes, I encountered this as well when writing out the PoC tests in **scrapy-poet**. There are cases where users might want to declared multiple rules for the same PO but with different `to_return` values. This could be applicable when a developer is compiling different rules from multiple POPs and the `ApplyRule` could be used to unify them via `to_return`.

The same is also true when the developer wants a single Page Object to replace multiple POs using the `instead_of` parameter.

Created an issue in https://github.com/scrapinghub/web-poet/issues/86 to keep track of it.

tests/po_lib_to_return/__init__.py
BurnzZ: Good catch! I was having the more/less words in my mind. 😆

tests/test_rules.py
BurnzZ: Yes, this is one part of our test suite which is brittle as it relies on some code comments to keep it working. https://github.com/scrapinghub/web-poet/pull/84/files/9729e8f1b1691af417ebfe4c0719231231fa3ae7#diff-236257b347f2d255e260081e2980f584367de9fc7999ba32c08e47e4f7edcad5R145-R153.

Do you have some specific mechanism in mind to make it more robust?

pyproject.toml
BurnzZ: Updated in https://github.com/scrapinghub/web-poet/pull/84/commits/88c511d51cfdf64f1b4ae69563f7a97f8cf61cff.

tests/test_rules.py
BurnzZ: Good idea! Updated in https://github.com/scrapinghub/web-poet/pull/84/commits/88c511d51cfdf64f1b4ae69563f7a97f8cf61cff.

tests/test_rules.py
BurnzZ: Good catch! I'm not sure was this test didn't even use that 😅 Updated in https://github.com/scrapinghub/web-poet/pull/84/commits/88c511d51cfdf64f1b4ae69563f7a97f8cf61cff.

web_poet/_typing.py
BurnzZ: Updated in https://github.com/scrapinghub/web-poet/pull/84/commits/928188e2c1267654ff12492e129cc6ecf0a7e251

web_poet/rules.py
BurnzZ: I see, we were assigning `to_return=None` before on cases like:

```python
@handle_urls("example.com")
class SomePage(ItemPage):
    ...
```

Updated this behavior in https://github.com/scrapinghub/web-poet/pull/84/commits/3679b59720e53fa74b7f58568310930b75418ab7

web_poet/rules.py
BurnzZ: Updated in https://github.com/scrapinghub/web-poet/pull/84/commits/fc0ba50610cfc78c9c463e7eab52b20a365c3985

web_poet/rules.py
BurnzZ: Makes sense! Updated in https://github.com/scrapinghub/web-poet/pull/84/commits/fc0ba50610cfc78c9c463e7eab52b20a365c3985

web_poet/rules.py
BurnzZ: If we have the following code:

```python
from web_poet import handle_urls

class Foo:
    ...

def do_something():
    @handle_urls("example.com", overrides=Foo)
    class Bar:
        ...

do_something()
```

**stacklevel=2** would have:
```python
tmp.py:10: DeprecationWarning: The 'overrides' parameter in @handle_urls is deprecated. Use the 'instead_of' parameter.
  class Bar:
```

**stacklevel=3** would have:
```python
tmp.py:14: DeprecationWarning: The 'overrides' parameter in @handle_urls is deprecated. Use the 'instead_of' parameter.
  do_something()
```

I think **stacklevel=2** is more appropriate here as it points to the class being decorated.

web_poet/rules.py
kmike: Wait, I think in this case None is the right behavior.

web_poet/rules.py
kmike: Thanks for checking it, makes sense!

docs/intro/overrides.rst
BurnzZ: Good idea, that hasn't occurred to me. Yep, let's sanitize the URLs in the docs and tests. https://github.com/scrapinghub/web-poet/pull/84/commits/0626e578b73843c11ec306a92946ee5d3bf78040

web_poet/rules.py
BurnzZ: To clarify, you're thinking about the following case right? wherein the users haven't declared the Item class in both the Page Object and the Rule:

```python
@handle_urls("example.com")
class SomePage(ItemPage):
    ...
```

In this case, should `to_return` be assigned with `None` or `dict` as the default value?

In the https://github.com/scrapinghub/web-poet/commit/3679b59720e53fa74b7f58568310930b75418ab7 commit, I've changed it from `None` to `dict` since by default, the `Returns` class was using `dict` in a subtle way.

This means that ItemPage subclasses without any hints about what it's able to return is expected to return `dict`s by default.

Since we're also thinking about establishing a concept for Crawling POs, I think it could serve the purpose of having `to_return=None` since it's not really expected to return any time at all, but rather serve as a convenience object to extract links, paginations, etc used for crawling. On the other hand, Data POs should at least have `to_return=dict` as a default to establish the fact that it's capable of returning an item.

kmike: I think we should special-case dict, because it doesn't make sense from the providers point of view. We'd have a lot of page objects which can return dicts. Does it mean that if user declares a dependency on a dict (or just annotates something as a dict), we should call some random page object?

We can special-case it in a library like scrapy-poet (and in other similar frameworks, when we have them). But it seems web-poet can be a better place to do it. I was thinking about `dict` as a fallback for the cases where a return item class is not defined, not as a real item class.

> Since we're also thinking about establishing a concept for Crawling POs, I think it could serve the purpose of having to_return=None since it's not really expected to return any item at all

I'm not sure about it. Crawling page objects could return an item with information needed for crawling - e.g. ProductList item for products. We will need to standardize these data types.

BurnzZ: I see, I think I understand what you mean now as I got it the other way around before. 😅 Updated in https://github.com/scrapinghub/web-poet/pull/84/commits/de15a86b8497b1dd40b29668a184409cc80f7177

web_poet/pages.py
```diff
@@ -41,16 +41,13 @@ def is_injectable(cls: typing.Any) -> bool:
 
 
 class Returns(typing.Generic[ItemT]):
-    """Inherit from this generic mixin to change the item type used by
+    """Inherit from this generic mixin to change the Item Class used by
```
kmike, line 44: A nitpick: I'm not sure about the capitalization change here. In Scrapy and itemadapter docs "Item" usually means scrapy.Item, and "item" means an item, something which is supported by itemadapter.

web_poet/rules.py
```diff
@@ -0,0 +1,350 @@
...
+
+    def get_overrides(self) -> List[ApplyRule]:
+        """Deprecated, use :meth:`~.PageObjectRegistry.get_rules` instead."""
+        msg = "The 'get_overrides' method is deprecated. Use 'get_rules' instead."
+        warnings.warn(msg, DeprecationWarning, stacklevel=2)
+        return self.get_rules()
+
+    def search_rules(self, **kwargs) -> List[ApplyRule]:
+        """Return any :class:`ApplyRule` that has any of its attributes
+        match the rules inside the registry.
```
kmike, line 245: Could you please re-phrase it? I initially read it as "if any of the kwargs match attribute name of the rule, return the rule". But based on tests it seems all kwargs should be matched.

web_poet/pages.py
BurnzZ: I've made these terms in Proper Case to indicate the that we're talking about the concept of "Item Classes" in the perspective of `@handle_urls` and Page Objects.

Using "item class" could simply mean that its a literal class and could lack the concept of rules, the class that's returned, etc.

web_poet/rules.py
BurnzZ: Good catch! https://github.com/scrapinghub/web-poet/pull/84/commits/bce97be36bc4c6a6e2ca5889d3a2f2c2e440fef0

docs/intro/overrides.rst
```diff
@@ -58,84 +85,223 @@ Let's take a look at how the following code is structured:
...
+        - If a ``Product`` Item Class is requested for URLs matching with the
+          "some.example" pattern, then the ``Product`` Item Class would come from
+          the ``to_item()`` method of ``ExampleProductPage``.
+        - Similarly, if a page with a URL matches with "another.example" without
+          the "/digital-goods/" path, then the ``Product`` Item Class comes from
+          the ``AnotherExampleProductPage`` Page Object.
+        - However, if a ``Product`` Item Class is requested matching with the URL
+          pattern of "dual.example/shop/?product=*", a ``SimilarProduct``
+          Item Class is returned by the ``DualExampleProductPage``'s ``to_item()``
+          method instead.
```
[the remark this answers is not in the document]
BurnzZ: I believe it's still worth providing a standalone explanation here as it could be the case that the reader specifically traverses the docs via `Overrides -> Creating Overrides -> Item Class`.

But yeah, some of the information could be redundant if they've already read `Overrides -> Creating Overrides -> Page Object`.

I'm inclined to still leave this on as it doesn't bring much downsides apart from repeating information. But I consider that as an advantage as this subsection can stand on itself.

docs/intro/overrides.rst
```diff
@@ -58,84 +85,223 @@ Let's take a look at how the following code is structured:
...
+from using a Page Object into Item Classes instead (returned by the ``to_item()``
+method).
+
+The upside here is that you're able to directly access the item extracted by the
+Page Object for the given site. There's no need to directly call the ``to_item()``
+method to receive it.
+
+However, one downside to this approach is that you lose access to the actual Page
+Object. Aside from the item extracted by the Page Object, there might be some
+other convenience methods or other data from it that you want to access.
```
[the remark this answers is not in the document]
BurnzZ: Good idea! I like this change of perspective. 👍

docs/intro/overrides.rst
```diff
@@ -1,15 +1,20 @@
...
 
 Using such rules establishes the core concept of Overrides wherein a developer
-could declare that a specific Page Object must be used *(instead of another)*
-for a given set of URL patterns.
+could declare that for a given set of URL patterns, a specific Page Object must
+be used instead of another Page Object.
+
+The :class:`~.ApplyRule` also supports pointing to the item returned by a specific
+Page Object if it both matches the URL pattern and the Item Class specified in the
+rule.
```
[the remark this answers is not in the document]
BurnzZ: > I would keep this introduction to explain in as few words as possible what an
override is

I'm not too sure about this one, as this section functions like an **abstract** which summarizes what to expect from a given paper (in this case, the doc section). 

> I think both of those can be introduced later as needed.

It will definitely be introduced later on. But we still need a high-level overview to provide the motivation of why this section of the doc is important.

> it may be a good idea to make the first reference to Page Object here
link to the Page Objects section in intro/from-ground-up.rst.

Good idea!

docs/intro/overrides.rst
```diff
@@ -58,84 +85,223 @@ Let's take a look at how the following code is structured:
...
+    class DualExampleProductPage(WebPage[SimilarProduct]):
+        ...  # more specific parsing
+
+To recap some previously mentioned behaviors:
+
+    - In the ``@handle_urls`` decorator for ``ExampleProductPage``, we needed to
+      pass the ``to_return`` parameter since the Page Object didn't specify the
+      Item Class that it returns (i.e. ``class ExampleProductPage(WebPage[Product])``).
+      If it did, we could omit the ``to_return`` parameter entirely since it's
+      able to automatically derive the Item Class.
```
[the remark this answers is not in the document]
BurnzZ: It is to show that `to_return` is derived using the annotation in 
```python
@handle_urls(URL)
class ExampleProductPage(WebPage[Product]): ...
```

If that's not available, then the following could work:
```python
@handle_urls(URL, to_return=Product)
class ExampleProductPage(WebPage): ...
```

docs/intro/overrides.rst
```diff
@@ -58,84 +85,223 @@ Let's take a look at how the following code is structured:
...
+    the :class:`~.ApplyRule`. For example:
+
+    .. code-block:: python
+
+        @handle_urls("some.example", to_return=DifferentProduct)
+        class ProductPage(WebPage[Product])
+            ... # more specific parsing
+
+    The Page Object returns a ``Product`` item but the created :class:`~.ApplyRule`
+    has ``to_return=DifferentProduct``.
```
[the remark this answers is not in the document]
BurnzZ: > I don’t like that we mention “more advanced use cases” without specific examples of such use cases. I would either not mention the existence of such use cases at all, or add a later section that covers one such case, and link to that here.

You're right. Although adding such advanced use cases might encourage users to go that route. On the other hand, omitting the availability of the `to_return` option might not provide users with a clue that it's available for `@handle_urls`. Although I think the API refs showing it available as one of the parameters is already enough.

With this, I'm inclined to remove the mention of `to_return` in the docs. It could also simplify this tutorial a lot more. This would also promote the much preferred way of declaring them via:
```python
@handle_urls(URL)
class ProductPage(WebPage[Product]): ...
```

@kmike do you have some specific cases in mind that you think we should mention? If none, we can remove `to_return` from the docs.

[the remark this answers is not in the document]
kmike: > @kmike do you have some specific cases in mind that you think we should mention? If none, we can remove to_return from the docs.

I think that's fine to "undocument" to_return parameter of `handle_urls`.

docs/intro/overrides.rst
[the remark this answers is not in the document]
Gallaecio: I would avoid linking to stuff the used does not need to read at this point. But I have no strong opinion on this, so let’s move on.

docs/intro/overrides.rst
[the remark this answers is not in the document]
Gallaecio: > But I consider that as an advantage as this subsection can stand on itself.

I agree that this is a good goal to have, that it is good for sections to stand on their own. However, this section starts with “An alternative approach for the Page Object Overrides example above”, which I believe sets the expectation for the user to read the example above before they continue reading.

That said, I have no strong opinion here, so moving on.

web_poet/rules.py
```diff
@@ -0,0 +1,352 @@
...
+    The main functionality of this class lies in the ``instead_of`` and ``to_return``
+    parameters. Should both of these be omitted, then :class:`~.ApplyRule` simply
+    tags which URL patterns the given Page Object is expected to be used.
+
+    As much as possible, the ``to_return`` parameter should capture the Item Class
+    that the Page Object is capable of returning. Before passing it to :class:`~.ApplyRule`,
+    the ``to_return`` value is primarily derived from the return class specified
+    from Page Objects that are subclasses of :class:`~.ItemPage`. However, a
+    special case exists when a Page Object returns a ``dict`` as an item but then
+    the rule should have ``to_return=None`` and **NOT** ``to_return=dict``.
```
Gallaecio, line 57: Some questions that I still have about `to_return` after reading the whole docstring:

- Is `to_return` a replacement for `use`? Does it make sense to use both parameters? When to use one an not the other?

> As much as possible, the ``to_return`` parameter should capture the Item Class that the Page Object is capable of returning.

- Which page object, the one assigned to “use”? Why “as much as possible”, when would it make sense to set a different `to_return` value? Is this about the `dict` case mentioned later, or is there more to it?

> Before passing it to :class:`~.ApplyRule`

Should we replace “Before” with “When not”? Or does it change the meaning?

> primarily derived

Given “primarily” here, I assume there are other cases, and we should cover those here.

> the return class specified from Page Objects that are subclasses of :class:`~.ItemPage`

I think users reading this may not understand how those return classes are specified. Maybe we could link “specified” to a documentation piece covering how to declare the return type of a Page Object by subclassing ItemPage.

web_poet/rules.py
```diff
@@ -0,0 +1,352 @@
...
+                warnings.warn(msg, DeprecationWarning, stacklevel=2)
+
+            rule = ApplyRule(
+                for_patterns=Patterns(
+                    include=as_list(include),
+                    exclude=as_list(exclude),
+                    priority=priority,
+                ),
+                use=cls,
+                instead_of=instead_of or overrides,
```
Gallaecio, line 206: Should we have a special warning if both are used, indicating that overrides are being ignored?

web_poet/rules.py
```diff
@@ -0,0 +1,352 @@
...
+                to_return=to_return or get_item_cls(cls),
```
Gallaecio, line 207: Would this make sense?

```suggestion
                to_return=to_return or get_item_cls(cls) or get_item_cls(instead_of or overrides),
```

docs/intro/overrides.rst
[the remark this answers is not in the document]
BurnzZ: Removed `to_return` usage in `@handle_urls()` https://github.com/scrapinghub/web-poet/pull/84/commits/076e7bb5ae74b700789ea8a2e74882d66253747c

docs/intro/overrides.rst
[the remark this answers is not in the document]
BurnzZ: Removed `to_return` usage in `@handle_urls()` https://github.com/scrapinghub/web-poet/pull/84/commits/076e7bb5ae74b700789ea8a2e74882d66253747c

tests/test_rules.py
```diff
@@ -0,0 +1,280 @@
...
+    patterns = Patterns(include=["example.com"], exclude=["example.com/blog"])
+
+    rule = ApplyRule(
+        for_patterns=patterns,
+        use=POTopLevel1,
+        instead_of=POTopLevelOverriden2,
+    )
+
+    with pytest.raises(attrs.exceptions.FrozenInstanceError):
+        rule.use = POModule  # type: ignore[misc]
```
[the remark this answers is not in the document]
BurnzZ: https://github.com/scrapinghub/web-poet/pull/84/commits/1419f7a30aeeda580c9d3560de016bfde00b598f

tests/test_rules.py
```diff
@@ -0,0 +1,280 @@
...
+        assert rule.meta == rule.use.expected_meta, rule.use  # type: ignore[attr-defined]
+
+
+def test_registry_get_overrides_deprecation() -> None:
+    msg = "The 'get_overrides' method is deprecated. Use 'get_rules' instead."
+    with pytest.warns(DeprecationWarning, match=msg):
+        rules = default_registry.get_overrides()
+
+    # It should still work as usual
+    assert len(rules) == len(default_registry.get_rules())
```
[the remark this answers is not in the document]
BurnzZ: Great catch! https://github.com/scrapinghub/web-poet/pull/84/commits/1419f7a30aeeda580c9d3560de016bfde00b598f

tests/test_rules.py
```diff
@@ -0,0 +1,280 @@
...
+
+    # param: instead_of
+    rules = default_registry.search_rules(instead_of=POTopLevelOverriden2)
+    assert len(rules) == 1
+    assert rules[0].instead_of == POTopLevelOverriden2
+
+    # param: to_return
+    rules = default_registry.search_rules(to_return=Product)
+    assert len(rules) == 3
+    assert all([r for r in rules if r.use == Product])
```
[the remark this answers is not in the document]
BurnzZ: Good one! https://github.com/scrapinghub/web-poet/pull/84/commits/1419f7a30aeeda580c9d3560de016bfde00b598f

tests/test_rules.py
```diff
@@ -0,0 +1,280 @@
...
+
+    registry = PageObjectRegistry.from_apply_rules(rules)
+
+    assert registry.get_rules() == rules
+    assert default_registry.get_rules() != rules
+
+
+def test_from_override_rules_deprecation() -> None:
+    rules = [
+        ApplyRule(
```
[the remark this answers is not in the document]
BurnzZ: https://github.com/scrapinghub/web-poet/pull/84/commits/1419f7a30aeeda580c9d3560de016bfde00b598f

tests/test_rules.py
```diff
@@ -0,0 +1,280 @@
...
+def test_handle_urls_deprecation() -> None:
+    msg = (
+        "The 'overrides' parameter in @handle_urls is deprecated. Use the "
+        "'instead_of' parameter."
+    )
+    with pytest.warns(DeprecationWarning, match=msg):
+
+        @handle_urls("example.com", overrides=CustomProductPage)
+        class PageWithDeprecatedOverrides:
+            ...
```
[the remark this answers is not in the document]
BurnzZ: Yeah, it would, but it wouldn't affect how the other tests work due to the sequential reading of this test module. But that's a good point! as despite the deprecation warning, it should still be added successfully into the registry. https://github.com/scrapinghub/web-poet/pull/84/commits/1419f7a30aeeda580c9d3560de016bfde00b598f

web_poet/rules.py
```diff
@@ -0,0 +1,352 @@
...
+    You can access any of its attributes:
+
+        * ``for_patterns`` - contains the list of URL patterns associated with
+          this rule. You can read the API documentation of the `url-matcher
+          <https://url-matcher.readthedocs.io/>`_ package for more information
+          about the patterns.
+        * ``use`` - The Page Object that will be **used** in cases where the URL
+          pattern represented by the ``for_patterns`` attribute is matched.
+        * ``instead_of`` - *(optional)* The Page Object that will be **replaced**
+          with the Page Object specified via the ``use`` parameter.
```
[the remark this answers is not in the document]
BurnzZ: > I think we should clarify later in the docstring how this parameter affects the mapping. I assume that, if you do not specify instead_of, use is used for all matches of for_patterns, but if you specify instead_of, use is used only for matches of for_patterns whose page object would otherwise be instead_of.

The idea is almost there. I added some notes in the docstrings about this to remove the ambiguity and clarify on this further.

> I also wonder what happens when you have 2 levels of instead_of. If you have a @handle_urls decorator on B with instead_of=A, and on C with instead_of=B, is B used or is it C? Do we have a test for this?

That's an interesting point. I don't believe we have this in **web-poet** as the framework (e.g. **scrapy-poet**) is the one that should implement it. For **scrapy-poet**, I don't believe we test it explicitly on this. Created an issue in https://github.com/scrapinghub/scrapy-poet/issues/90 to keep track of it. Thanks!

tests/test_rules.py
```diff
@@ -0,0 +1,280 @@
...
+    )
+    rule2 = ApplyRule(
+        for_patterns=patterns,
+        use=POTopLevel1,
+        instead_of=POTopLevelOverriden2,
+        to_return=ProductSimilar,
+    )
+    # A different Item Class results in different hash.
+    assert hash(rule1) != hash(rule2)
+
```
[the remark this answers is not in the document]
Gallaecio: My thinking was to test them separately, i.e. test scenarios where everything matches except for those, not to test them all at once. If we test them all at once, the tests will pass even if only 1 of them makes the hash change.

tests/test_rules.py
```diff
@@ -0,0 +1,373 @@
...
+    )
+
+    with pytest.raises(attrs.exceptions.FrozenInstanceError):
+        rule.use = Patterns(include=["example.com/"])  # type: ignore[misc]
+
+    with pytest.raises(attrs.exceptions.FrozenInstanceError):
+        rule.use = POTopLevel2  # type: ignore[misc]
+
+    with pytest.raises(attrs.exceptions.FrozenInstanceError):
+        rule.use = POTopLevelOverriden2  # type: ignore[misc]
```
Gallaecio, line 127:
```suggestion
    with pytest.raises(attrs.exceptions.FrozenInstanceError):
        rule.for_patterns = Patterns(include=["example.com/"])  # type: ignore[misc]

    with pytest.raises(attrs.exceptions.FrozenInstanceError):
        rule.use = POTopLevel2  # type: ignore[misc]

    with pytest.raises(attrs.exceptions.FrozenInstanceError):
        rule.instead_of = POTopLevelOverriden2  # type: ignore[misc]
```

web_poet/rules.py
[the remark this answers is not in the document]
Gallaecio: > I added some notes in the docstrings about this to remove the ambiguity and clarify on this further.

I do not see any change, did you push it yet?

web_poet/rules.py
BurnzZ: Thanks for raising these! I've improved the docs to help alleviate the confusion and ambiguity on them. Let me know what you think! https://github.com/scrapinghub/web-poet/pull/84/commits/776cf0dd16a466504e64875307da86d4ee2a6115

> Which page object, the one assigned to “use”? Why “as much as possible”, when would it make sense to set a different to_return value? Is this about the dict case mentioned later, or is there more to it?

I removed the “as much as possible" phrase here to remove the ambiguity. We also want to encourage the use of declaring the item class via `ItemPage[Product]`.

> Should we replace “Before” with “When not”? Or does it change the meaning?

Let's stick with "Before" as it means that the value should be derived first before before instantiating `ApplyRule`.

> Given “primarily” here, I assume there are other cases, and we should cover those here.

The "primarily" suggests that we can't actually derive everything completely from the return class. For example, if the PO returns `dict`, we set it to `to_return=None`. I'm not sure if we should cover other advanced use cases apart from this special case as early as now. I'm also not sure what are the other special scenarios that deriving something won't actually work. We can cover that later on as we and our users stumble across them.

> I think users reading this may not understand how those return classes are specified. Maybe we could link “specified” to a documentation piece covering how to declare the return type of a Page Object by subclassing ItemPage.

Good idea. I've added a link pointing to the `Item Class` subsection in the Overrides tutorial.

web_poet/rules.py
BurnzZ: Updated the current warning message to also capture this notion. https://github.com/scrapinghub/web-poet/pull/84/commits/776cf0dd16a466504e64875307da86d4ee2a6115

web_poet/rules.py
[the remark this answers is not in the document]
BurnzZ: Oohh right sorry, forgot to push and mention it https://github.com/scrapinghub/web-poet/pull/84/commits/776cf0dd16a466504e64875307da86d4ee2a6115.

tests/test_rules.py
BurnzZ: Thank you for catching! 🙌

tests/test_rules.py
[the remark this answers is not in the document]
BurnzZ: Ohhh right, that's indeed much better. Thanks! https://github.com/scrapinghub/web-poet/pull/84/commits/36cd866263f2081fa221a00c218568c5a4026fd1

Gallaecio (approved):

web_poet/pages.py
kmike: Hey! Going back to it :) In itemadapter and in Scrapy a lower-case "item class" is used to talk about a concept; Item class means `scrapy.Item` class, or maybe some other concrete item class. Upper-case Item can't be used in itemadapter, because it would cause confusion with scrapy.Item class. So, it's the opposite approach in itemadapter docs.

BurnzZ: > Upper-case Item can't be used in itemadapter, because it would cause confusion with scrapy.Item class. So, it's the opposite approach in itemadapter docs.

Gotcha. I just re-read https://github.com/scrapy/itemadapter 👍 . Updated in https://github.com/scrapinghub/web-poet/pull/84/commits/383b4f74261388d149138804f4d0d41791f22691

web_poet/rules.py
```diff
@@ -0,0 +1,393 @@
...
+        1. Given a URL, match it against the ``for_patterns`` from the registry
+           rules.
+        2. This could give us a collection of rules. We need to select one based
+           on the highest priority set by `url-matcher`_.
+        3. When a single rule has been selected, use the the Page Object specified
+           in its ``use`` parameter.
+
+    If ``instead_of=None``, this simply means that the Page Object assigned in
+    the ``use`` parameter will be utilized for all URLs matching the URL pattern
+    in ``for_patterns``. However, if ``instead_of=ReplacedPageObject``, then it
```
kmike, line 62: This doesn't look correct: if `instead_of=None`, then Page Object won't be used to replace another Page Object, regardless of URL pattern. It might be used on the provided URL pattern if `to_return` is set, but this has nothing to do with `instead_of=None`, as it should work regardless of `instead_of` value. 

So, in the example below:

```py
@handle_urls("example.com")
class MyPage(ItemPage):
    # ...
``` 

handle_urls is a pure documentation. You can't use it instead of any other page object (as instead_of is not provided), and you can't use it to return a certain data type (because the returned item class is unknown).

kmike: @BurnzZ let me send a PR with some doc-related suggestions

kmike: https://github.com/scrapinghub/web-poet/pull/90

kmike:
Thanks @BurnzZ, great work!

PR #85, opened 2022-10-14
Add support for field processors.

wRAR:
This adds support for an `out=` argument to `@field`, as proposed in https://github.com/zytedata/zyte-common-items/issues/18 (option 3). It only supports a list of callables, though it should be easy to also add support for a single callable. It only supports synchronous processors as async processors don't seem too useful and would change the signature of the field from sync to async which I find problematic.

I'm not sure how idiomatic (wrt both Python and web-poet) is the implementation but it's simple enough to be able to adjust it.

### DIFF
--- a/docs/advanced/fields.rst
+++ b/docs/advanced/fields.rst
@@ -151,6 +151,34 @@ function when accessing the fields:
 Now any field can be converted from sync to async, or the other way around,
 and the code would keep working.
 
+Field processors
+----------------
+
+It's often needed to clean or process field values using reusable functions.
+``@field`` takes an optional ``out`` argument with a list of such functions.
+They will be applied to the field value before returning it:
+
+.. code-block:: python
+
+    from web_poet import ItemPage, HttpResponse, field
+
+    def clean_tabs(s):
+        return s.replace('\t', ' ')
+
+    class MyPage(ItemPage):
+        response: HttpResponse
+
+        @field(out=[clean_tabs, str.strip])
+        def name(self):
+            return self.response.css(".name ::text").get()
+
+Note that while processors can be applied to async fields, they need to be
+normal functions themselves.
+
+It's also possible to implement field cleaning and processing in ``to_item``
+but in that case accessing a field directly will return the value without
+processing, so it's preferable to use field processors instead.
+
 Item classes
 ------------
 
@@ -197,9 +225,6 @@ but item classes are of a great help when
 
 * you need to extract data in the same format from multiple websites, or
 * if you want to define the schema upfront.
-
-Item classes can also be used to hold common attribute
-pre-processing and validation logic.
 
 Error prevention
 ~~~~~~~~~~~~~~~~
@@ -488,27 +513,3 @@ returns a dictionary, where keys are field names, and values are
 
     print(field_names)  # dict_keys(['my_field'])
     print(my_field_meta)  # {'expensive': True}
-
-Field processors
-----------------
-
-It's often needed to clean or process field values using reusable functions.
-``@field`` takes an optional ``out`` argument with a list of such functions.
-They will be applied to the field value before returning it:
-
-.. code-block:: python
-
-    from web_poet import ItemPage, HttpResponse, field
-
-    def clean_tabs(s):
-        return s.replace('\t', ' ')
-
-    class MyPage(ItemPage):
-        response: HttpResponse
-
-        @field(out=[clean_tabs, str.strip])
-        def name(self):
-            return self.response.css(".name ::text").get()
-
-Note that while processors can be applied to async fields, they need to be
-normal functions themselves.
--- a/docs/advanced/fields.rst
+++ b/docs/advanced/fields.rst
@@ -173,7 +173,7 @@ They will be applied to the field value before returning it:
             return self.response.css(".name ::text").get()
 
 Note that while processors can be applied to async fields, they need to be
-normal functions themselves.
+sync functions themselves.
 
 It's also possible to implement field cleaning and processing in ``to_item``
 but in that case accessing a field directly will return the value without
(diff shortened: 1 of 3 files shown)


web_poet/fields.py
```diff
@@ -63,6 +73,28 @@ def __init__(self, method):
                 raise TypeError(
                     f"@field decorator must be used on methods, {method!r} is decorated instead"
                 )
+            if out:
+                orig_method = method
```
wRAR, line 77: This is needed because, as I've just found out, Python closures get the final value of the variable, not the one at the point where the closure is defined, so `method = processed` is applied when the closure is called.

web_poet/fields.py
```diff
@@ -63,6 +73,28 @@ def __init__(self, method):
...
+                else:
+
+                    @wraps(method)
+                    def processed(*args, **kwargs):
+                        result = orig_method(*args, **kwargs)
+                        for processor in out:
+                            result = processor(result)
+                        return result
+
+                method = processed
```
kmike, line 96: A very small cleanup: define a function like this:

```py
def _process(out, result):  # out can be omitted if you defined it in a closure
    for processor in out:
        result = processor(result)
    return result
```

And then the code can be simplified a bit:
```suggestion
                if inspect.iscoroutinefunction(method):

                    @wraps(method)
                    async def processed(*args, **kwargs):
                        return _process(await orig_method(*args, **kwargs))
                        
                else:

                    @wraps(method)
                    def processed(*args, **kwargs):
                        return _process(orig_method(*args, **kwargs))
                        
                method = processed
```

Not sure if it worths it though :)

web_poet/fields.py
```diff
@@ -55,6 +62,9 @@ def field(method=None, *, cached: bool = False, meta: Optional[dict] = None):
     The ``meta`` parameter allows to store arbitrary information for the field,
     e.g. ``@field(meta={"expensive": True})``. This information can be later
     retrieved for all fields using the :func:`get_fields_dict` function.
+
+    The ``out`` parameter is an optional list of field processors, which are
```
kmike, line 66: I wonder if it'd be useful to add `out` to FieldInfo

web_poet/fields.py
wRAR: Yeah, I thought about that too. Well, I think it simplifies the code, I'll implement it.

kmike: You can also move the whole block to a separate function; it'd be similar to cached_method decorator. It looks fine as-is though.

kmike: One more idea: use a single wraps() call (or update_wrapper), don't use a decorator

kmike: e.g. `method = wraps(method)(processed)`

kmike:
The implementation and tests makes sense to me, thanks @wRAR! 

Could you please add docs for this feature? it should be a section in https://github.com/scrapinghub/web-poet/blob/master/docs/advanced/fields.rst

kmike:
There is also a note in the current docs, which says 

> Item classes can also be used to hold common attribute pre-processing and validation logic.

I think we should remove it, and actually explain that having preprocessing logic in items may be problematic. This new docs section would be a good place for it.

web_poet/fields.py
wRAR: done

web_poet/fields.py
wRAR: done

wRAR:
I've added a docs section.

> Item classes can also be used to hold common attribute pre-processing and validation logic.

> I think we should remove it, and actually explain that having preprocessing logic in items may be problematic. This new docs section would be a good place for it.

Why would be it be problematic?

docs/advanced/fields.rst
```diff
@@ -488,3 +488,27 @@ returns a dictionary, where keys are field names, and values are
 
     print(field_names)  # dict_keys(['my_field'])
     print(my_field_meta)  # {'expensive': True}
+
+Field processors
```
kmike, line 492: Let's move this section above the Caching section; it seems important.

kmike:
> Why would it be problematic?

It breaks the following (from the docs):

> The resulting page object becomes more flexible and reusable: if not all data extracted in the to_item() method is needed, user can use properties for individual attributes. It’s more efficient than running to_item() and only using some of the result.

If field processing happens in the item, it means you can't use page_object.field instead of item.field, the result would be wrong (i.e. without cleaning).

docs/advanced/fields.rst
kmike: Or: what do you think about moving it above Items section as well?

kmike: It seems there's a lot of docs about items, sometimes documenting edge cases; field processors look like a straightforward feature which people need to be aware of.

web_poet/fields.py
```diff
@@ -63,6 +76,9 @@ def __init__(self, method):
                 raise TypeError(
                     f"@field decorator must be used on methods, {method!r} is decorated instead"
                 )
+            if out:
+                method = self._processed(method)
```
kmike, line 80: Another nitpick: this looks a bit weird because of closures :) It's a bit surprising that self_processed can use `out` value. I think it can be solved either by passing `out` as an argument, or by moving this `if out:` check to the `self._processed` method.

kmike (approved):
Looks good @wRAR!

docs/advanced/fields.rst
```diff
@@ -151,6 +151,34 @@ function when accessing the fields:
...
+
+    class MyPage(ItemPage):
+        response: HttpResponse
+
+        @field(out=[clean_tabs, str.strip])
+        def name(self):
+            return self.response.css(".name ::text").get()
+
+Note that while processors can be applied to async fields, they need to be
+normal functions themselves.
```
Gallaecio, line 176: We don’t want async functions to feel left out :smiley: 

```suggestion
sync functions themselves.
```

Gallaecio (approved):
Nice!

kmike:
After re-reading https://github.com/zytedata/zyte-common-items/issues/18 I realized that typing could be an issue. But it turns out proper typing is not implemented for `@field` at all. So, we should treat is as a separate issue (https://github.com/scrapinghub/web-poet/issues/87, https://github.com/scrapinghub/web-poet/pull/88).

Issue #86, opened 2022-10-14
Consider removing the limitation of only having a single PO in the registry


BurnzZ:
Stemming off from the discussion in https://github.com/scrapinghub/web-poet/pull/84#discussion_r995643314.

kmike:
Would it make sense to remove dict-like interface for PageObjectRegistry for 0.6.0 release? It seems it might get in a way if we want to implement this feature.

BurnzZ:
Related threads:
- https://github.com/scrapinghub/web-poet/pull/96/files#r1009636014
- https://github.com/scrapinghub/web-poet/pull/96/files#r1010083865

PR #96, opened 2022-10-31
updated PageObjectRegistry API

kmike:
It seems now is the best time to try to clean up API for PageObjectRegistry; there are already changes related to overrides -> rules move, and it seems it'd be nice to bundle it all together.

See https://github.com/scrapinghub/web-poet/issues/86 - the idea is not to solve it here, but to prepare the API for this.

### DIFF
--- a/web_poet/rules.py
+++ b/web_poet/rules.py
@@ -115,8 +115,8 @@ class RulesRegistry:
         context.
     """
 
-    def __init__(self, *, rules: Optional[List[ApplyRule]] = None):
-        self._page_object_rules: dict[Type[ItemPage], ApplyRule] = {}
+    def __init__(self, *, rules: Optional[Iterable[ApplyRule]] = None):
+        self._page_object_rules: Dict[Type[ItemPage], ApplyRule] = {}
         if rules is not None:
             self._page_object_rules = {rule.use: rule for rule in rules}
 


web_poet/rules.py
```diff
@@ -81,40 +81,33 @@ def __hash__(self):
         return hash((self.for_patterns, self.use, self.instead_of, self.to_return))
 
 
-class PageObjectRegistry(dict):
-    """This contains the :class:`~.ApplyRule` that associates the Page Objects
-    alongside its Items for a given URL matching rule.
-
-    PageObjectRegistry is a ``dict`` subclass with added functionalities on
```
kmike, line 88: The idea here is to deprecate or remove `dict` API.

@BurnzZ - a couple of questions for you:

1.  would it cause issues for scrapy-poet?
2. would it be ok to remove it without deprecation?

tests/test_rules.py
```diff
@@ -317,7 +317,7 @@ def test_from_override_rules_deprecation_using_OverrideRule() -> None:
         "Use 'from_apply_rules' instead."
     )
     with pytest.warns(DeprecationWarning, match=msg):
-        registry = PageObjectRegistry.from_override_rules(rules)
+        registry = RulesRegistry.from_override_rules(rules)
```
kmike, line 320: An option is to change the API to be `RulesRegistry(rules)`, if dict-like API is removed/deprecated. 

I think that's fine to do it later, as a separate step, because we can do it in a backwards-compatible way.

web_poet/rules.py
BurnzZ: 1. **scrapy-poet** wouldn't be affected at all as it's not relying on any `dict`-features of the registry. Moreover, its only interaction with the registry is via:

```python
SCRAPY_POET_OVERRIDES = default_registry.get_rules()
```

After retrieving the rules, it creates its own [OverridesRegistry](https://github.com/scrapinghub/scrapy-poet/blob/master/scrapy_poet/overrides.py#L28).

2. I think we should still retain the deprecation for backwards compatibility. I can imagine how users might be relying on the `dict`-interface for some of their needs.

If we remove it right away, there are some `dict` functionalities that would be left unattended:
- record deletion via key
- equality comparison (though `.get_rules()` could be used)
- length checks (though `.get_rules()` could be used)

web_poet/rules.py
```diff
@@ -17,7 +17,7 @@
 
 Strings = Union[str, Iterable[str]]
 
-PageObjectRegistryTV = TypeVar("PageObjectRegistryTV", bound="PageObjectRegistry")
+RulesRegistryTV = TypeVar("RulesRegistryTV", bound="RulesRegistry")
```
BurnzZ, line 20: For completeness of backwards-compatibility, we could also have `PageObjectRegistryTV` assuming users were relying on it.

web_poet/rules.py
```diff
@@ -288,13 +287,10 @@ def matcher(rule: ApplyRule):
         return results
 
     def search_overrides(self, **kwargs) -> List[ApplyRule]:
-        """Deprecated, use :meth:`~.PageObjectRegistry.search_rules` instead."""
-        msg = (
-            "The 'search_overrides' method is deprecated. "
-            "Use 'search_rules' instead."
-        )
+        """Deprecated, use :meth:`~.RulesRegistry.search` instead."""
+        msg = "The 'search_overrides' method is deprecated. " "Use 'search' instead."
```
BurnzZ, line 291:
```suggestion
        msg = "The 'search_overrides' method is deprecated. Use 'search' instead."
```

BurnzZ: Black needs to fix these. 😅

web_poet/rules.py
```diff
@@ -81,52 +81,51 @@ def __hash__(self):
...
 
     .. note::
 
-        It is encouraged to simply use and import the already existing registry
-        via ``from web_poet import default_registry`` instead of creating your
-        own :class:`~.PageObjectRegistry` instance. Using multiple registries
-        would be unwieldy in most cases.
+        It is encouraged to use the ``web_poet.default_registry`` instead of
+        creating your own :class:`~.RulesRegistry` instance. Using multiple
+        registries would be unwieldy in most cases.
```
BurnzZ, line 110:
```suggestion
        registries would be unwieldy in most cases (see :ref:`rules-custom-registry`).
```

web_poet/rules.py
kmike: > I can imagine how users might be relying on the dict-interface for some of their needs.

I hope that the use of this feature is minimal now; I wouldn't expent many people to be using overrides registry at the moment, or customizing it.

web_poet/rules.py
kmike: That's a good point. I'm not sure how to deprecate PageObjectRegistryTV though. It's also not intended to be used, and I doubt anyone is using it at the moment, given that it's early days.

web_poet/rules.py
```diff
@@ -81,52 +81,51 @@ def __hash__(self):
...
-class PageObjectRegistry(dict):
-    """This contains the :class:`~.ApplyRule` that associates the Page Objects
-    alongside its Items for a given URL matching rule.
-
-    PageObjectRegistry is a ``dict`` subclass with added functionalities on
-    storing, retrieving, and searching for the :class:`~.ApplyRule` instances.
-    The **value** represents the :class:`~.ApplyRule` instance from which the
-    Page Object in the **key** is allowed to be used. Since it's essentially a
-    ``dict``, you can use any ``dict`` operations with it.
+class RulesRegistry(dict):
```
BurnzZ, line 84: Related to https://github.com/scrapinghub/web-poet/issues/86#issuecomment-1297274048.

What do you think about retaining `PageObjectRegistry` for backwards compatibility with the `dict`-interface, but then introducing a new Registry that's not based on `dict`? In this way, we can support multiple rules for a single Page Object. 

I imagine that we need this before the **0.6.0** release since it's a blocker for use cases like when users want to use the same Page Object that supports different `to_return` values.

kmike: > What do you think about retaining PageObjectRegistry for backwards compatibility with the dict-interface, but then introducing a new Registry that's not based on dict? In this way, we can support multiple rules for a single Page Object.

I might be missing something, but for me it looks like the direct use of PageObjectRegistry is next to non-existent now, and so we can be pretty agressive deprecating and removing features. It'd be a larger issue after handle_urls cleanup is released, together with scrapy-poet changes, as I'd expect the usage to increase.

> I could see how we need this before the 0.6.0 release

I think cleaning up API is a blocker, but supporting multiple rules is not a blocker. We can add this support later, and there's enough useful code to release in web-poet.

web_poet/rules.py
Gallaecio: > I'm not sure how to deprecate PageObjectRegistryTV though.

Me neither. But I am thinking we could consider its deprecation implied by the deprecation of all API members that use it, and remove them along those API members.

I also thought about making these variables private, but seeing how scrapy-poet subclasses, it seems like it makes sense to keep them public.

web_poet/rules.py
kmike: Even if we don't go with #86, it still can be a good idea to make the exposed API smaller, as it allows us to make changes easier without breaking user code.

Gallaecio: This may be the cleanest, otherwise we would need to reimplement all methods of dict to mark them as deprecated, and I wonder if some of the [pitfalls of dict-inheritance](https://realpython.com/inherit-python-dict/#inheriting-from-the-python-built-in-dict-class) could be an issue for that.

kmike: Ok, let me give it a try

kmike: Done!

tests/test_rules.py
kmike: Implemented in https://github.com/scrapinghub/web-poet/pull/96/commits/fc8c09139c9f32d4decd42055ee6794ee862a664

BurnzZ (approved):
I got some minor comments regarding typing but otherwise, this LGTM! Nice work @kmike ! 🎉

web_poet/rules.py
```diff
@@ -81,71 +81,56 @@ def __hash__(self):
...
-    ) -> PageObjectRegistryTV:
-        """An alternative constructor for creating a :class:`~.PageObjectRegistry`
-        instance by accepting a list of :class:`~.ApplyRule`.
-
-        This is useful in cases wherein you need to store some selected rules
-        from multiple external packages. See this :ref:`example
-        <rules-custom-registry>`.
-        """
-        return cls({rule.use: rule for rule in rules})
+    def __init__(self, *, rules: Optional[List[ApplyRule]] = None):
```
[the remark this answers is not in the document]
kmike: Fixed in https://github.com/scrapinghub/web-poet/pull/96/commits/58ff02dcdd0c7efb4be87444c57b338167250297

web_poet/rules.py
```diff
@@ -81,71 +81,56 @@ def __hash__(self):
...
+        self._page_object_rules: dict[Type[ItemPage], ApplyRule] = {}
```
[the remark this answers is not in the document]
kmike: Fixed in https://github.com/scrapinghub/web-poet/pull/96/commits/58ff02dcdd0c7efb4be87444c57b338167250297

Gallaecio (approved):

PR #97, opened 2022-11-01
Add a section about writing frameworks

### LINKED ISSUE #54: Split documentation by target audience
At the moment we kind of mix documentation for page object writers and for framework writers. I think it would be best to split the two at the root of the documentation, so that the top-level sections of the table of contents become:

1. Getting started
2. Writing page objects
3. Writing frameworks
4. Reference

So, for example, the content under “Framework expectations” in the “Additional requests” topic would be split into a separate page, under “Writing frameworks”.

Gallaecio:
Minimal changes to address #54:

-  The “Framework Expectations” sections of the “Additional Requests” and “Retries” pages have been moved into pages of their own, listed in the table of contents under “Writing frameworks”. On the old pages, no other change has been made. On the new pages, nothing besides the top-level reference ID, page title and the introductory paragraphs have been touched.
- The “Advanced” section of the table of contents has been renamed to “Writing page objects”, with the corresponding path change for all affected pages.

Resolves #54.


docs/frameworks/retries.rst
```diff
@@ -0,0 +1,28 @@
...
+page object.
+
+When :exc:`~web_poet.exceptions.core.Retry` is caught:
+
+#.  The original request whose response was fed into the page object must be
+    retried.
+
+#.  A new page object must be created, of the same type as the original page
+    object, and with the same input, except for the response data, which must
+    be the new response.
```
kmike, line 18: I'm not sure that's valid. All inputs which correspond to the same page should be re-created, there shouldn't be a scenario when one input comes from an original page, and the other one comes from a retried result.

Gallaecio: I imagine this is worded like this to account for inputs that should not change during a retry, which I assume is the case of RequestUrl and PageParams. We could reword it to talk about updating inputs that can be updated, although it may be best to be specific about what input should and which inputs should not be updated during a retry.

docs/frameworks/additional-requests.rst
```diff
@@ -0,0 +1,228 @@
+.. _framework-additional-requests:
+
+==============================
+Supporting Additional Requests
+==============================
+
+To support :ref:`additional requests <advanced-requests>`, your framework must
+provide the request download implementation of :class:`~.HttpClient`.
```
kmike, line 8: Could you please check

> The key words "MUST”, "MUST NOT”, "REQUIRED”, "SHALL”, "SHALL NOT”, "SHOULD”,
"SHOULD NOT”, "RECOMMENDED”, "MAY”, and "OPTIONAL” in this document are to be
interpreted as described in RFC `2119 <https://www.ietf.org/rfc/rfc2119.txt>`_.

in the page-pbjects/additional-requests? I think this section applies mostly to the framework docs, and if we keep it, we should keep it for framework docs only.

docs/frameworks/additional-requests.rst
```diff
@@ -0,0 +1,228 @@
...
+
+.. _advanced-downloader-impl:
```
kmike, line 10: Is it a right time to clean up ref labels?

Gallaecio: I have no strong opinion either way. Since they are not user-visible, I tend not to worry about them until I need to touch related code.

docs/frameworks/additional-requests.rst
Gallaecio: I have removed it for now. Once we work on the framework docs, and have an introduction page for them, we *can* consider adding it back. Though I am not sure we need it.

kmike:
Thanks @Gallaecio!

PR #98, opened 2022-11-01
Added typing issues reported due to the latest version of Parsel

Gallaecio:


tests/test_pages.py
```diff
@@ -36,9 +36,12 @@ def to_item(self) -> dict:
 def test_web_page_object(book_list_html_response) -> None:
     class MyWebPage(WebPage):
         def to_item(self) -> dict:  # type: ignore
+            title = self.css("title::text").get()
+            if title is not None:
+                title = title.strip()
             return {
                 "url": self.url,
-                "title": self.css("title::text").get().strip(),
+                "title": title,
```
BurnzZ, line 44:
```suggestion
                "title": self.css("title::text").get("").strip(),
```

BurnzZ (approved):
I think there's a shorter solution to this as suggested but otherwise the fix looks good.

PR #100, opened 2022-11-02
move Apply Rules docs from intro to "page objects" section

kmike:


BurnzZ (approved):
I was thinking the same thing!

Gallaecio (approved):

PR #101, opened 2022-11-02
update RulesRegistry to allow multiple rules in a single Page Object

BurnzZ:
Resolves https://github.com/scrapinghub/web-poet/issues/86.

Related to the prior work in https://github.com/scrapinghub/web-poet/pull/96.

This unblocks the WIP in https://github.com/scrapinghub/scrapy-poet/pull/88 when considering use cases that needs multiple rules in a single PO.

It's not necessary to include this in the `0.6.0` release.

### DIFF
--- a/web_poet/rules.py
+++ b/web_poet/rules.py
@@ -215,7 +215,7 @@ class RulesRegistry:
             beforehand to recursively import all submodules which contains the
             ``@handle_urls`` decorators from external Page Objects.
         """
-        return self._rules
+        return self._rules[:]
 
     def get_overrides(self) -> List[ApplyRule]:
         """Deprecated, use :meth:`~.RulesRegistry.get_rules` instead."""
--- a/web_poet/rules.py
+++ b/web_poet/rules.py
@@ -116,9 +116,9 @@ class RulesRegistry:
     """
 
     def __init__(self, *, rules: Optional[Iterable[ApplyRule]] = None):
-        self._page_object_rules: Dict[Type[ItemPage], ApplyRule] = {}
+        self._rules: List[ApplyRule] = []
         if rules is not None:
-            self._page_object_rules = {rule.use: rule for rule in rules}
+            self._rules = list(rules)
 
     @classmethod
     def from_override_rules(
@@ -200,17 +200,7 @@ class RulesRegistry:
                 to_return=to_return or get_item_cls(cls),
                 meta=kwargs,
             )
-            # If it was already defined, we don't want to override it
-            if cls not in self._page_object_rules:
-                self._page_object_rules[cls] = rule
-            else:
-                warnings.warn(
-                    f"Multiple @handle_urls decorators for the same Page Object "
-                    f"are ignored in the same Registry. The following rule is "
-                    f"ignored:\n{rule}",
-                    stacklevel=2,
-                )
-
+            self._rules.append(rule)
             return cls
 
         return wrapper
@@ -225,7 +215,7 @@ class RulesRegistry:
             beforehand to recursively import all submodules which contains the
             ``@handle_urls`` decorators from external Page Objects.
         """
-        return list(self._page_object_rules.values())
+        return self._rules
 
     def get_overrides(self) -> List[ApplyRule]:
         """Deprecated, use :meth:`~.RulesRegistry.get_rules` instead."""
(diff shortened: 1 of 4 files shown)


Gallaecio (approved):

web_poet/rules.py
```diff
@@ -248,14 +238,6 @@ def search(self, **kwargs) -> List[ApplyRule]:
 
         """
 
-        # Short-circuit operation if "use" is the only search param used, since
```
kmike, line 251: I've checked scrapy-poet code, and it seems that currently fast lookups in the registry are not needed, as matching is implemented in a different way. So, it seems that's fine to drop this optimization. We still should mention this change in the changelog though.

web_poet/rules.py
```diff
@@ -225,7 +215,7 @@ def get_rules(self) -> List[ApplyRule]:
             beforehand to recursively import all submodules which contains the
             ``@handle_urls`` decorators from external Page Objects.
         """
-        return list(self._page_object_rules.values())
+        return self._rules
```
kmike, line 218:
```suggestion
        return self._rules[:]
```

web_poet/rules.py
kmike: Because of the algorithmic complexity change, I think it makes sense to consider it a blocker for 0.6.0

kmike: (i.e. bundle all changes like this in 0.6.0)

BurnzZ: > I've checked scrapy-poet code, and it seems that currently fast lookups in the registry are not needed,

I think one optimization we need is fast lookups for returning the rules for a given request (particularly its URL). Currently, it's looking through all the rules ([ref](https://github.com/scrapinghub/scrapy-poet/blob/master/scrapy_poet/overrides.py#L115-L121)):

```python
    def overrides_for(self, request: Request) -> Mapping[Callable, Callable]:
        overrides: Dict[Callable, Callable] = {}
        for instead_of, matcher in self.matcher.items():
            rule_id = matcher.match(request.url)
            if rule_id is not None:
                overrides[instead_of] = self.rules[rule_id].use
        return 
```

Although I'm optimizing this in the https://github.com/scrapinghub/scrapy-poet/pull/88 PR since it needs a refactor anyways for the new `to_return` functionality.

BurnzZ: > Because of the algorithmic complexity change, I think it makes sense to consider it a blocker for 0.6.0

Does this mean that we'll need to consider bundling this change into `0.6.0`?

kmike: > Currently, it's looking through all the rules

I think it's not that bad. `self.matcher` (not a great name, by the way) is dictionary of `{generic_cls: URLMatcher}`. So, the loop happens not against all rules, but against all generic (overridable) classes. For example, if you have GenericProductPage and GenericProductListPage, and there are page objects for 1000 websites, the loop would do only 2 iterations, not 1000. 

Picking a right rule is handled by url-matcher, and it should be optimized pretty well; it has a domain index, so only rules which are defined for this particular domain are iterated internally. In most cases, there would be very few of them.

So, based on this, I'm not concerned about impact on scrapy-poet.

But I do think we need to mention this algorithmic change in the web-poet changelog, just in case, as it's another backwards incompatible change in RulesRegistry: O(1) lookups using `registry.search(use=MyPage)` are no longer supported, they're O(N) now.

> Does this mean that we'll need to consider bundling this change into 0.6.0?

Yes, because of the above.

kmike: By the way, it looks like the code for URL matching which is currently in scrapy-poet could be a better fit in web-poet. Finding a rule in registry based on URL and instead_of / to_return argument is not specific to Scrapy.

kmike (approved):

web_poet/rules.py
BurnzZ: Updated the changelog in https://github.com/scrapinghub/web-poet/pull/101/commits/955870155b08f361234dcb057547d397ac214660

> `self.matcher` (not a great name, by the way) is dictionary of `{generic_cls: URLMatcher}`. So, the loop happens not against all rules, but against all generic (overridable) classes.

Right, I missed out on the fact that it's pooling all of the URL patterns that a given PO supports.

> By the way, it looks like the code for URL matching which is currently in scrapy-poet could be a better fit in web-poet. Finding a rule in registry based on URL and instead_of / to_return argument is not specific to Scrapy.

I agree. I had to subclass `web_poet.RulesRegistry` in **scrapy-poet**'s registry in https://github.com/scrapinghub/scrapy-poet/pull/88 to avoid re-inventing some of its features like search. We can move the matching functionality **web-poet** later on.

kmike:
Thanks @BurnzZ!

PR #103, opened 2022-11-02
Add SwitchPage

Gallaecio:
Changes:
- Implement a `SwitchPage` class, that can be subclassed to create special page object classes that call other page object classes based on the received input.
- New documentation page about webpage layouts as a web scraping concept, to provide context before introducing the new class.
  It took me a while to write it in a way that did not presume too much prior knowledge from the reader, and I am still unhappy with the result, so I would love some feedback here.

Arguments for the proposed API:
- I think the code that determines which page object class to use should be centralized (`switch` method in the current proposal), as opposed to defining it on each page object class. Otherwise, handling priorities and corner cases could become cumbersome.
- In regards to the eventual implementation of a mechanism to detect invalid layouts, I do not see a problem. Assuming that we implement it by raising some special exception, such an exception could be raised from `switch` if no right page object class is found, or from the `to_item` method of the selected page object class if, despite that page object class getting selected from `switch`, the input it gets does not match some other expectation.
- For cases where different candidate page object classes have different inputs, the documentation asks for the switch page object class to be declared with a combination of all of them, and then the right ones would be fed into the selected page object class (to be implemented; I assume it is feasible, but I would like some input on how to approach it).

To do:
- [x] Discuss the proposed API.
    - [ ] ~~Do we need to have some override mechanism for candidate page object classes? i.e. Does there need to be a way through rules to replace a candidate page object class instead of replacing the whole switch page object class?~~ (out of scope)
- [x] ~~Implement input passing to the selected candidate page object class (andi? inspect and simple type matching?).~~ (layout page object classes are now declared as inputs, letting web-poet frameworks handle this automatically through dependency injection)
- [x] ~~Provide a nice exception if there is a missing input?~~ (not relevant given the resolution of the bullet point above)

### DIFF
--- a/docs/page-objects/layouts.rst
+++ b/docs/page-objects/layouts.rst
@@ -185,8 +185,8 @@ Your multi-layout page object class should:
 
 You may use :class:`~web_poet.pages.MultiLayoutPage` as a base class for your
 multi-layout page object class, so you only need to implement the
-:class:`~web_poet.pages.MultiLayoutPage.layout` method that determines which
-page object to use. For example:
+:class:`~web_poet.pages.MultiLayoutPage.get_layout` method that determines
+which page object to use. For example:
 
 .. code-block:: python
 
@@ -222,7 +222,7 @@ page object to use. For example:
        h1: H1Page
        h2: H2Page
 
-       async def layout(self) -> ItemPage[Header]:
+       async def get_layout(self) -> ItemPage[Header]:
            if self.response.css("h1::text"):
                return self.h1
            return self.h2
--- a/docs/page-objects/layouts.rst
+++ b/docs/page-objects/layouts.rst
@@ -131,7 +131,6 @@ page object to use. For example:
     from web_poet import handle_urls, HttpResponse, ItemPage, MultiLayoutPage, WebPage
 
 
-    @attrs.define
     class Header:
         text: str
 
--- a/docs/page-objects/layouts.rst
+++ b/docs/page-objects/layouts.rst
@@ -49,11 +49,11 @@ there are at least 2 different webpage layouts.
 
 However, whether you continue to work as if everything uses the same webpage
 layout, or you split your page object class into 2 page object classes, each
-targetting one of the webpage layouts you have found, it is entirely up to you.
+targeting one of the webpage layouts you have found, it is entirely up to you.
 
 Ask yourself: Is supporting all webpage layout differences making your page
 object class implementation only a few lines of code longer, or is it making it
-an unmaintainable bowl of spagetti code?
+an unmaintainable bowl of spaghetti code?
 
 
 Mapping webpage layouts
@@ -74,10 +74,10 @@ When that is the case, you can :ref:`associate your page object class to the
 corresponding URL pattern <rules-intro>`.
 
 
-.. _switch:
+.. _multi-layout:
 
-Switch page object classes
---------------------------
+Multi-layout page object classes
+--------------------------------
 
 Sometimes it is impossible to know, based on the target URL, which webpage
 layout you are getting. For example, during `A/B testing`_, you could get a
@@ -85,39 +85,47 @@ random webpage layout on every request.
 
 .. _A/B testing: https://en.wikipedia.org/wiki/A/B_testing
 
-For these scenarios, we recommend that you create a special “switch” page
-object class, and use it to switch to the right page object class at run time
-based on the input you receive.
+For these scenarios, we recommend that you create different page object classes
+for the different layouts that you may get, and then write a special
+“multi-layout” page object class, and use it to select the right page object
+class at run time based on the input you receive.
 
-Your switch page object class should:
+Your multi-layout page object class should:
 
-#.  Request all the inputs that the candidate page object classes may need.
+#.  Declare attributes for the input that you will need to determine which page
+    object class to use.
 
-    For example, if there are 2 candidate page object classes, and 1 of them
-    requires browser HTML as input, while the other one requires an HTTP
-    response, your switch page object class must request both.
+    For example, declare an :class:`HttpResponse` attribute to select a page
+    object class based on the response content.
 
-    If combining different inputs is a problem, consider refactoring the
-    candidate page object classes to require similar inputs.
+#.  Declare an attribute for every page object class that you may depending on
+    which webpage layout you get from the target website.
+
+    Note that all inputs of all those page object classes will be resolved and
+    requested along with the input of your multi-layout page object class. For
+    example, if one page object class requires browser HTML as input, while
+    another requires an HTTP response, your multi-layout page object class asks
+    for both inputs.
+
+    If combining different inputs is a problem, consider refactoring your page
+    object classes to require similar inputs.
 
 #.  On its :meth:`~web_poet.pages.ItemPage.to_item` method:
 
-    #.  Determine, based on the inputs, which candidate page object class to
-        use.
+    #.  Determine, based on inputs, which page object to use.
 
-    #.  Create an instance of the selected candidade page object class with the
-        necessary input, call its :meth:`~web_poet.pages.ItemPage.to_item`
-        method, and return its result.
+    #.  Return the output of the :meth:`~web_poet.pages.ItemPage.to_item`
+        method of that page object.
 
-You may use :class:`~web_poet.pages.SwitchPage` as a base class for your switch
-page object class, so you only need to implement the
-:class:`~web_poet.pages.SwitchPage.switch` method that determines which
-candidate page object class to use. For example:
+You may use :class:`~web_poet.pages.MultiLayoutPage` as a base class for your
+multi-layout page object class, so you only need to implement the
+:class:`~web_poet.pages.MultiLayoutPage.layout` method that determines which
+page object to use. For example:
 
 .. code-block:: python
 
     import attrs
-    from web_poet import handle_urls, HttpResponse, Injectable, ItemPage, SwitchPage
+    from web_poet import handle_urls, HttpResponse, ItemPage, MultiLayoutPage, WebPage
 
 
     @attrs.define
@@ -126,29 +134,29 @@ candidate page object class to use. For example:
 
 
     @attrs.define
-    class H1Page(ItemPage[Header]):
-        response: HttpResponse
+    class H1Page(WebPage[Header]):
 
         @field
         def text(self) -> str:
-            return self.response.css("h1::text").get()
+            return self.css("h1::text").get()
 
 
     @attrs.define
-    class H2Page(ItemPage[Header]):
-        response: HttpResponse
+    class H2Page(WebPage[Header]):
 
         @field
         def text(self) -> str:
-            return self.response.css("h2::text").get()
+            return self.css("h2::text").get()
 
 
     @handle_urls("example.com")
     @attrs.define
-    class HeaderSwitchPage(SwitchPage[Header]):
+    class HeaderMultiLayoutPage(MultiLayoutPage[Header]):
         response: HttpResponse
+        h1: H1Page
+        h2: H2Page
 
-        async def switch(self) -> Injectable:
+        async def layout(self) -> ItemPage[Header]:
             if self.response.css("h1::text"):
-                return H1Page
-            return H2Page
+                return self.h1
+            return self.h2
--- a/docs/page-objects/layouts.rst
+++ b/docs/page-objects/layouts.rst
@@ -96,7 +96,13 @@ Your multi-layout page object class should:
     object class to use.
 
     For example, declare an :class:`HttpResponse` attribute to select a page
-    object class based on the response content.
+    object class based on the response content:
+
+    .. code-block:: python
+
+       class MyMultiLayoutPage(ItemPage):
+           response: HttpResponse
+           ...
 
 #.  Declare an attribute for every page object class that you may use depending
     on which webpage layout you get from the target website.
@@ -104,11 +110,51 @@ Your multi-layout page object class should:
     They all should return the same type of :ref:`item <item-classes>` as your
     multi-layout page object class.
 
+    For example:
+
+    .. code-block:: python
+
+       class MyItem:
+           ...
+
+       @attrs.define
+       class MyPage1(ItemPage[MyItem]):
+           ...
+
+       @attrs.define
+       class MyPage2(ItemPage[MyItem]):
+           ...
+
+       @attrs.define
+       class MyMultiLayoutPage(ItemPage[MyItem]):
+           ...
+           page1: MyPage1
+           page2: MyPage2
+
     Note that all inputs of all those page object classes will be resolved and
-    requested along with the input of your multi-layout page object class. For
-    example, if one page object class requires browser HTML as input, while
-    another requires an HTTP response, your multi-layout page object class asks
-    for both inputs.
+    requested along with the input of your multi-layout page object class.
+
+    For example, given:
+
+    .. code-block:: python
+
+       @attrs.define
+       class MyPage1(ItemPage):
+           response: HttpResponse
+
+       @attrs.define
+       class MyPage2(ItemPage):
+           response: BrowserHtml
+
+       @attrs.define
+       class MyMultiLayoutPage(ItemPage):
+           response: HttpResponse
+           page1: MyPage1
+           page2: MyPage2
+
+    Using ``MyMultiLayoutPage`` causes the use of both ``HttpResponse`` and
+    ``BrowserHtml``, because ``MyMultiLayoutPage`` requires ``MyPage2``, and
+    ``MyPage2`` requires ``BrowserHtml``.
 
     If combining different inputs is a problem, consider refactoring your page
     object classes to require similar inputs.
@@ -120,6 +166,23 @@ Your multi-layout page object class should:
     #.  Return the output of the :meth:`~web_poet.pages.ItemPage.to_item`
         method of that page object.
 
+    For example:
+
+    .. code-block:: python
+
+       @attrs.define
+       class MyMultiLayoutPage(ItemPage[MyItem]):
+           response: HttpResponse
+           page1: MyPage1
+           page2: MyPage2
+
+           async def to_item(self) -> MyItem:
+               if self.response.css(".foo"):
+                   page_object = self.page1
+               else:
+                   page_object = self.page2
+               return await page_object.to_item()
+
 You may use :class:`~web_poet.pages.MultiLayoutPage` as a base class for your
 multi-layout page object class, so you only need to implement the
 :class:`~web_poet.pages.MultiLayoutPage.layout` method that determines which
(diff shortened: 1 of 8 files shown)


docs/page-objects/layouts.rst
```diff
@@ -0,0 +1,154 @@
...
+
+The code that those webpages have in common is their **webpage layout**.
+
+Coding for webpage layouts
+==========================
+
+Webpage layouts should inform how you organize your data extraction code.
+
+A good practice to keep your code maintainable is to have a separate :ref:`page
+object class <page-objects>` per webpage layout.
```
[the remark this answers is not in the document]
Gallaecio: I plan to eventually address this as part of my on-going work on the documentation.

web_poet/pages.py
```diff
@@ -68,6 +68,28 @@ async def to_item(self) -> ItemT:
...
 
 
+class SwitchPage(Injectable, Returns[ItemT]):
+    """Base class for :ref:`switch page object classes <switch>`.
+
+    Subclasses must reimplement the :meth:`switch` method.
+    """
+
+    @abc.abstractmethod
+    async def switch(self) ->  Injectable:
```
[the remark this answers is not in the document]
Gallaecio: I agree. `SelectionPage`/`select` and `ChoicePage`/`choose` would work for me. Any other suggestion or preference?

web_poet/pages.py
```diff
@@ -68,6 +68,28 @@ async def to_item(self) -> ItemT:
...
+        """Return the right :ref:`page object class <page-objects>` based on
+        the received input."""
```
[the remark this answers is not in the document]
Gallaecio: I have no clear idea on what the best approach here is. I think your questions aligns a bit with one of my open question:

> Do we need to have some override mechanism for candidate page object classes?

docs/page-objects/layouts.rst
```diff
@@ -0,0 +1,154 @@
...
+        text: str
+
+
+    @attrs.define
+    class H1Page(ItemPage[Header]):
+        response: HttpResponse
+
+        @field
+        def text(self) -> str:
+            return self.response.css("h1::text").get()
```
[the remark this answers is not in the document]
kmike: I think self.response.css is fine. Also, self.css doesn't work in this example, because H1Page is an instance of ItemPage (not WebPage), so it doesn't have this shortcut :)

web_poet/pages.py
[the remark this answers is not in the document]
kmike: I think we don't need to support overrides for candidate page object classes.

docs/page-objects/layouts.rst
```diff
@@ -0,0 +1,154 @@
...
+object class, and use it to switch to the right page object class at run time
+based on the input you receive.
+
+Your switch page object class should:
+
+#.  Request all the inputs that the candidate page object classes may need.
+
+    For example, if there are 2 candidate page object classes, and 1 of them
+    requires browser HTML as input, while the other one requires an HTTP
+    response, your switch page object class must request both.
```
kmike, line 98: What do you think about making "layout" page objects dependencies of the main page object? If we do so, it'd remove the need.

docs/page-objects/layouts.rst
```diff
@@ -0,0 +1,154 @@
...
+    candidate page object classes to require similar inputs.
+
+#.  On its :meth:`~web_poet.pages.ItemPage.to_item` method:
+
+    #.  Determine, based on the inputs, which candidate page object class to
+        use.
+
+    #.  Create an instance of the selected candidade page object class with the
+        necessary input, call its :meth:`~web_poet.pages.ItemPage.to_item`
+        method, and return its result.
```
kmike, line 110: If these POs are dependencies, to_item won't need to create them - figure out which dependencies they need, handle dependency trees, etc.

docs/page-objects/layouts.rst
```diff
@@ -0,0 +1,154 @@
...
+
+    @handle_urls("example.com")
+    @attrs.define
+    class HeaderSwitchPage(SwitchPage[Header]):
+        response: HttpResponse
+
+        async def switch(self) -> Injectable:
+            if self.response.css("h1::text"):
+                return H1Page
+            return H2Page
```
kmike, line 154: An API idea (unfinished):

```py
class MyPage(MultiLayoutPage[Header]):
    class Layouts:
        h1: H1Page
        h2: H2Page 

    layouts: Layouts

    async def layout(self) -> ItemPage:
        if self.response.css("h1::text"):
            return self.layouts.h1
        return self.layouts.h2
```

Gallaecio:
Note to self: sublayout support.

docs/page-objects/layouts.rst
kmike: or simply

```py
class MyPage(MultiLayoutPage[Header]):
    h1: H1Page
    h2: H2Page 

    async def layout(self) -> ItemPage:
        if self.response.css("h1::text"):
            return self.h1
        return self.h2
```

web_poet/pages.py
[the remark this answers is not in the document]
Gallaecio: Gone for MultiLayoutPage and layout based on feedback from Mikhail which seemed also in line with similar work by @gatufo.

Gallaecio:
Applied the [improvement proposal](https://github.com/scrapinghub/web-poet/pull/103#discussion_r1016290595) by @kmike. It is very clean, and saves me a lot of trouble. It also makes nesting (i.e. multi-layout page object classes that use other multi-layout page object classes) work by default, I believe.

Considering the problems of input mixing and override handling out of scope, my only remaining thoughts are:
- Should we provide an example of layout nesting in the docs?
- Should we provide some alternative or complementary API to make it easy to define layout-selection rules on layouts themselves instead of centralizing it on MultiLayoutPage? If not, should we at least provide an example on how to achieve this with the current API?

web_poet/pages.py
```diff
@@ -68,6 +68,25 @@ async def to_item(self) -> ItemT:
         )
 
 
+class MultiLayoutPage(ItemPage[ItemT]):
```
kmike, line 71: I also wonder if we should support fields for MultiLayoutPage, or not. 
Currently all fields defined for MultiLayoutPage are silently ignored. 

There is also a use case when you need a "partial" layout, layout of a region of a page. In this case, only some fields are extracted by the layout; others are common. It seems we have a few options:

1. Use a base class with common fields. Create layouts which inherit from the base class and handle necessary regions. Use them in MultiLayoutPage.
2. Allow to define fields on the MultiLayoutPage. When layout is picked, combine the fields from the layout with the fields defined in the class itself, to compute the final item. Supporting to_item for layouts can be more tricky in this case, as it's not clear which item they should use.

A separate issue is if it's possible to use 2 or more regions, with different layouts, in the same page object.

If fields are supported, it seems it makes sense to move the logic to ItemPage. I think it may simplify typing, and inheritance as well. E.g. layouts can be used with ProductPage from https://github.com/zytedata/zyte-common-items/pull/19.

It seems that if fields are not supported, it's better to keep MultiLayoutPage as a separate class, and probably raise an error or issue a warning if fields are defined.

Sorry for a braindump :) What do you think?

kmike: A disadvantage of using MultiLayoutPage without fields: it's not possible to use fields in the code which uses the page object. So let's say we have ProductPage, it uses fields for data extraction. Then, it's refactored to use MultiLayoutPage as a base class. It means that the fields are no longer supported, and so the code which uses this page may break.

Gallaecio: I am slightly against field support in MultiLayoutPage.

In addition to those 2 options you mention, I can think of a 3rd that is a variation on 1 based precisely on how you amended my API proposal for MultiLayoutPage:

3. Use composition rather than inheritance: define a page object class that can extract the common fields, make it a dependency of the layout(s) that extract additional fields, and use it in that layout.

We would still have the same problem as with the lack of fields in MultiLayoutPage, i.e. you could not access the fields of the dependency layout through the layout that uses it (other than accessing the dependency directly, e.g. `layout.dependency.field`). But I wonder if we could implement a getattr fallback mechanism to solve this issue for both scenarios: allow MultiLayoutPage to expose the fields of the object that its layout() method returns, and allow layouts with other layouts as dependencies to expose the fields of their dependencies.

docs/page-objects/layouts.rst
```diff
@@ -0,0 +1,170 @@
...
+:class:`~web_poet.pages.MultiLayoutPage.layout` method that determines which
+page object to use. For example:
+
+.. code-block:: python
+
+    import attrs
+    from web_poet import handle_urls, HttpResponse, ItemPage, MultiLayoutPage, WebPage
+
+
+    @attrs.define
```
BurnzZ, line 134:
```suggestion
```

docs/page-objects/layouts.rst
```diff
@@ -0,0 +1,170 @@
...
+    class Header:
+        text: str
+
+
+    @attrs.define
```
BurnzZ, line 139:
```suggestion
```

docs/page-objects/layouts.rst
```diff
@@ -0,0 +1,170 @@
...
+random webpage layout on every request.
+
+.. _A/B testing: https://en.wikipedia.org/wiki/A/B_testing
+
+For these scenarios, we recommend that you create different page object classes
+for the different layouts that you may get, and then write a special
+“multi-layout” page object class, and use it to select the right page object
+class at run time based on the input you receive.
+
+Your multi-layout page object class should:
```
BurnzZ, line 93: A minor suggestion: I think it'd be easier to understand these itemized points if we could show the example first and then go over the points.

tests/test_pages.py
```diff
@@ -33,6 +34,72 @@ def to_item(self) -> dict:
     }
 
 
+@pytest.mark.asyncio
+async def test_multi_layout_page_object():
+    @attrs.define
+    class Header:
+        text: str
+
+    @attrs.define
```
BurnzZ, line 43:
```suggestion
```

tests/test_pages.py
```diff
@@ -33,6 +34,72 @@ def to_item(self) -> dict:
...
+    class H1Page(WebPage[Header]):
+        @field
+        def text(self) -> Optional[str]:
+            return self.css("h1::text").get()
+
+    @attrs.define
```
BurnzZ, line 49:
```suggestion
```

docs/page-objects/layouts.rst
```diff
@@ -0,0 +1,170 @@
...
+    @attrs.define
+    class Header:
+        text: str
+
+
+    @attrs.define
+    class H1Page(WebPage[Header]):
+
+        @field
+        def text(self) -> str:
```
BurnzZ, line 143:
```suggestion
        def text(self) -> Optional[str]:
```

docs/page-objects/layouts.rst
```diff
@@ -0,0 +1,170 @@
...
+            return self.css("h1::text").get()
+
+
+    @attrs.define
+    class H2Page(WebPage[Header]):
+
+        @field
+        def text(self) -> str:
```
BurnzZ, line 151:
```suggestion
        def text(self) -> Optional[str]:
```

docs/page-objects/layouts.rst
```diff
@@ -0,0 +1,170 @@
...
+    #.  Return the output of the :meth:`~web_poet.pages.ItemPage.to_item`
+        method of that page object.
+
+You may use :class:`~web_poet.pages.MultiLayoutPage` as a base class for your
+multi-layout page object class, so you only need to implement the
+:class:`~web_poet.pages.MultiLayoutPage.layout` method that determines which
+page object to use. For example:
+
+.. code-block:: python
+
```
BurnzZ, line 129:
```suggestion
from typing import Optional
```

web_poet/pages.py
```diff
@@ -68,6 +68,25 @@ async def to_item(self) -> ItemT:
...
+    """Base class for :ref:`multi-layout page object classes <multi-layout>`.
+
+    Subclasses must reimplement the :meth:`layout` method.
+    """
+
+    @abc.abstractmethod
+    async def layout(self) -> ItemPage[ItemT]:
+        """Return the :ref:`page object <page-objects>` to use based on the
+        received input."""
```
BurnzZ, line 80:
```suggestion
        received input (e.g. :class:`~.HttpResponse`)."""
```

just to clarify what type of input we're talking about.

web_poet/pages.py
BurnzZ: I'm also not sure about support `@fields` in `MultiLayoutPage`.

It would seem it'd be best to keep it's task simple wherein it simply identifies and returns the PO instance based on the layout.

tests/test_pages.py
```diff
@@ -33,6 +34,72 @@ def to_item(self) -> dict:
...
+
+    response1 = HttpResponse("https://example.com", body=html_h1)
+    h1_1 = H1Page(response=response1)
+    h2_1 = H2Page(response=response1)
+    response2 = HttpResponse("https://example.com", body=html_h2)
+    h1_2 = H1Page(response=response2)
+    h2_2 = H2Page(response=response2)
+
+    item1 = await HeaderMultiLayoutPage(response=response1, h1=h1_1, h2=h2_1).to_item()
+    item2 = await HeaderMultiLayoutPage(response=response2, h1=h1_2, h2=h2_2).to_item()
```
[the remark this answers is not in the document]
Gallaecio: > use the same `HttpResponse` instance for a given page

I did not check at all, but I was assuming this as a given.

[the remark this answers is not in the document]
BurnzZ: Currently, the `HttpResponseProvider` of `scrapy-poet` returns a fresh `HttpResponse` instance every time: https://github.com/scrapinghub/scrapy-poet/blob/master/scrapy_poet/page_input_providers.py#L174-L178.

Created https://github.com/scrapinghub/scrapy-poet/issues/95 to keep track of future discussion.

docs/page-objects/layouts.rst
Gallaecio: I take it from your feedback that `@attrs.define` is not needed when inheriting from a class that already uses it. But in this specific case, with no parent class, isn’t it needed?

docs/page-objects/layouts.rst
Gallaecio: I agree that it is not easy to understand these points on a first read as it is, but I am unsure about placing the example first, because the example uses MultiLayoutPage which is basically a shortcut to some of these points.

What about including separate, illustrative examples of each point?

docs/page-objects/layouts.rst
Gallaecio: Hmm, it seems `@attrs.define` is also needed when new attributes are defined, otherwise you get things like:

```
TypeError: object.__new__() takes exactly one argument (the type to instantiate)
TypeError: __init__() got an unexpected keyword argument 'price'
```

Gallaecio: Hmm, it seems `@attrs.define` is also needed when new attributes are defined, otherwise you get things like:

```
TypeError: object.__new__() takes exactly one argument (the type to instantiate)
TypeError: __init__() got an unexpected keyword argument 'price'
```

web_poet/pages.py
Gallaecio: > There is one argument for keeping it separate and not supporting fields: it'd allow to define fields named `layout`. If we provide a standard method named `layout`, and allow to define fields, it's not possible to have a field of the same name.

This is a valid point, but I think we could switch to ``get_layout`` to avoid this issue, and it would be a better method name anyway.

docs/page-objects/layouts.rst
BurnzZ: Ahhh yes, I marked this incorrectly while in a `@attrs.define` removal spree yesterday. 😅

docs/page-objects/layouts.rst
BurnzZ: > What about including separate, illustrative examples of each point?

Yes, I think that would help in a more digestible info. 👍

web_poet/pages.py
Gallaecio: I still think defining fields on a multi-layout page object should be discouraged, but the more I think about it, the more I feel we indeed need to allow addressing the suggested refactoring scenario:

> So let's say we have ProductPage, it uses fields for data extraction. Then, it's refactored to use MultiLayoutPage as a base class. It means that the fields are no longer supported, and so the code which uses this page may break.

With the current proposal, you would have to manually define every field as:

```python
        @field
        async def foo(self) -> Optional[str]:
            return await self.layout().foo
```

When I wrote [my earlier comment](https://github.com/scrapinghub/web-poet/pull/103/files#r1018112627), I was thinking that we could make `__getattr__` fall back to `self.layout().__getattr__` to address this, so we do not need to add code like this to every field.

But when you inherit from `MultiLayoutPage`, other issues arise. For example, the `to_item` method of the inheriting class becomes that of `MultiLayoutPage`, preventing fields from the inheriting class to be included in the output of `to_item`.

Gallaecio: I wonder if, rather than providing a `MultiLayoutPage`, we should document how to implement one instead by:
- Defining a cached, class-private ``__get_layout`` method for the layout selection code, so that subclasses can safely redefine it, for cases of multiple multi-layout classes in the inheritance tree.
- Defining every single field from the underlying layout as in my previous comment. It is tedious, but it also plays well with `ItemPage.to_item`.

tests/test_pages.py
```diff
@@ -33,6 +34,166 @@ def to_item(self) -> dict:
...
+
+    # Returned items work as expected.
+    item1 = await multilayoutpage1.to_item()
+    assert item1.url == url
+    assert item1.text == "a"
+    item2 = await multilayoutpage2.to_item()
+    assert item2.url == url
+    assert item2.text == "b"
+
+
```
BurnzZ, line 196: Let's also add a test for use cases when a `MultiLayoutPage` subclass also uses multiple `MultiLayoutPage`s underneath. I think this could be a use case when users import POs from different packages and repacking them.

```python
class MyMultiLayoutPage(MultiLayoutPage[SomeItem]):
    response: HttpResponse
    layout_page_us: LayoutPageUS
    layout_page_uk: LayoutPageUK

    async def get_layout(self) -> ItemPage[SomeItem]:
        if self.response.css(".origin::text") == "us":
            return self.layout_page_us.get_layout()
        return self.layout_page_uk.get_layout()
```

Might also be worth creating a doc about this as well.

tests/test_pages.py
```diff
@@ -33,6 +34,166 @@ def to_item(self) -> dict:
...
+        async def text(self) -> Optional[str]:
+            return self.css("h2::text").get()
+
+    @attrs.define
+    class MyMultiLayoutPage(MultiLayoutPage[FullItem]):
+        response: HttpResponse
+        page1: FullPage1
+        page2: FullPage2
+
+        async def get_layout(self) -> ItemPage[FullItem]:
```
BurnzZ, line 136:
```suggestion
        async def get_layout(self) -> ItemPage:
```

Removes the need for the `type: ignore[return-value]` below. 

This should be okay since we're expecting `ItemPage` (and its subclasses) to be returned. For some reason, mypy fixates on the `FullItem` return type.

web_poet/pages.py
```diff
@@ -68,6 +68,25 @@ async def to_item(self) -> ItemT:
...
 
 
+class MultiLayoutPage(ItemPage[ItemT]):
+    """Base class for :ref:`multi-layout page object classes <multi-layout>`.
+
+    Subclasses must reimplement the :meth:`layout` method.
+    """
+
+    @abc.abstractmethod
+    async def get_layout(self) -> ItemPage[ItemT]:
```
BurnzZ, line 78: Not sure if this sounds right but how about `to_layout()`? In that way, we establish some sort of consistency with the PO's API naming, i.e. `to_item()`.

web_poet/pages.py
wRAR: @Gallaecio can you please explain the last idea in more details, e.g. with some code?

I agree that not having fields in `MultiLayoutPage` is a better design and I also agree that there are problems with it.

Gallaecio: > @Gallaecio can you please explain the last idea in more details, e.g. with some code?

https://github.com/scrapinghub/web-poet/pull/103/commits/fc7867fb0d9a8c48cc43d3bf2ec295d09fbf724f

wRAR: Yeah, this looks nice.

kmike: We were discussing it today on a call with @proway2. They implemented a multi-layout page object, tried a few approaches. In short - having fields on the final page object is a must :) That's the reason the documented approach here won't work well for them. Taking union of all dependencies is fine.

kmike: Currently they define all the fields, and call to the self._layout in each field. It's a lot of boilerplate; exactly something a library should be solving.

kmike: It seems having fields availabe is required, but not necessarily being able to define top-level fields.

PR #107, opened 2022-11-25
Serialization support

wRAR:
This is WIP but ready for review. It doesn't have docs yet and wasn't tested as a cache replacement in scrapy-poet.

The basic idea:

* This is mostly for serializing page objects and their attributes recursively. So it has some custom support for web-poet classes but otherwise Injectable attr-classes are serialized recursively.
* This means that unless a custom function is written (and it's written for `HttpResponse`), every attr-class is converted into a tree where leaves are primitive types. `HttpResponse`, on the other hand, is just two leaves: the body and a dict with all other fields (we wanted to save the body as a plain HTML file and to make sure we support custom serialization).
* The tree (a dict with dicts as nodes and `bytes` objects as leaves, typed as `SerializedData`) does not contain any typing information, only the raw data and attribute names to map it to attributes. To recreate the object we need its type definition. Then we can find the raw data for every attribute by its name and recreate it using its type.
* The tree is flattened when written to the disk, using dots as separators of node names (the `SerializedData` structure cleanly maps to a file system tree structure but I see no reason so far to use subdirectories). A top-level class is thus represented by a directory with files inside. As an example, a directory for a `WebPage` object contains two files: `response.body.html` and `response.other.json`. A `WebPage` descendant that also has a `response_url: ResponseUrl` field serializes into three files, adding `response_url.txt`. (As an implementation detail, the leaves in the tree have the file extension as the key in the parent dict, this looked like the cleanest solution to the problem that serializing a variable shouldn't need the variable name.)
* The dynamic serialization is implemented with [functools.singledispatch](https://docs.python.org/3/library/functools.html#functools.singledispatch). The function signature is `_(o: T) -> SerializedData`, and dispatching is done by the type of `o`.
* The dynamic deserialization cannot use the same mechanism, as it needs to take the type, not the object, so we store the deserialization function as an attribute of the paired serialization one and find it using `serialize.dispatch()` which takes the type. The function signature is `_(t: Type[T], data: SerializedData) -> T`.
* We expose the singledispatch-decorated `serialize()`, the wrapper that finds and calls the correct deserialization function as `deserialize()` and a `register_serialization()` function that takes a function pair and registers them. 
* Currently, as a limitation, Injectable descendants are assumed to be attr-classes, this needs to be fixed (note that there is no common ancestor, and so no straightforward way to dispatch, for attr-classes).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,7 +2,7 @@ import os
 
 import pytest
 
-from web_poet.page_inputs import HttpResponse, HttpResponseBody, ResponseUrl
+from web_poet.page_inputs import HttpResponse, HttpResponseBody
 
 
 def read_fixture(path):
@@ -20,7 +20,5 @@ def book_list_html():
 def book_list_html_response(book_list_html):
     body = HttpResponseBody(bytes(book_list_html, "utf-8"))
     return HttpResponse(
-        url=ResponseUrl("http://books.toscrape.com/index.html"),
-        body=body,
-        encoding="utf-8",
+        url="http://books.toscrape.com/index.html", body=body, encoding="utf-8"
     )
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,7 +2,7 @@ import os
 
 import pytest
 
-from web_poet.page_inputs import HttpResponse, HttpResponseBody
+from web_poet.page_inputs import HttpResponse, HttpResponseBody, ResponseUrl
 
 
 def read_fixture(path):
@@ -20,5 +20,7 @@ def book_list_html():
 def book_list_html_response(book_list_html):
     body = HttpResponseBody(bytes(book_list_html, "utf-8"))
     return HttpResponse(
-        url="http://books.toscrape.com/index.html", body=body, encoding="utf-8"
+        url=ResponseUrl("http://books.toscrape.com/index.html"),
+        body=body,
+        encoding="utf-8",
     )
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -22,15 +22,15 @@ from web_poet.serialization import (
 )
 
 
-def _compare_webpages(p1: WebPage, p2: WebPage) -> None:
+def _assert_webpages_equal(p1: WebPage, p2: WebPage) -> None:
     assert p1.response.body == p2.response.body
     assert p1.response.status == p2.response.status
     assert p1.response.headers == p2.response.headers
     assert p1.response._encoding == p2.response._encoding
-    _compare_urls(p1.response.url, p2.response.url)
+    _assert_urls_equal(p1.response.url, p2.response.url)
 
 
-def _compare_urls(u1: _Url, u2: _Url) -> None:
+def _assert_urls_equal(u1: _Url, u2: _Url) -> None:
     assert type(u1) == type(u2)
     assert str(u1) == str(u2)
 
@@ -54,10 +54,10 @@ def test_serialization_unsup() -> None:
 
 
 def test_serialization_webpage(book_list_html_response) -> None:
-    po = WebPage(book_list_html_response)
+    po: WebPage = WebPage(book_list_html_response)
     serialized_po = serialize(po)
     deserialized_po = deserialize(WebPage, serialized_po)
-    _compare_webpages(po, deserialized_po)
+    _assert_webpages_equal(po, deserialized_po)
 
 
 def test_serialization_httpresponse_encoding(book_list_html) -> None:
@@ -127,7 +127,7 @@ def test_custom_functions() -> None:
     assert obj.value == deserialized_obj.value
 
 
-def test_extra_attrs():
+def test_extra_attrs() -> None:
     @attrs.define
     class C(Injectable):
         f1: Optional[str]
@@ -144,11 +144,11 @@ def test_extra_attrs():
 def test_write_data(book_list_html_response, tmp_path) -> None:
     directory = tmp_path / "ser"
     directory.mkdir()
-    po = WebPage(book_list_html_response)
+    po: WebPage = WebPage(book_list_html_response)
     serialized_po = serialize(po)
     write_serialized_data(serialized_po, directory)
     assert (directory / "response.body.html").exists()
     read_serialized_po = read_serialized_data(directory)
     deserialized_po = deserialize(WebPage, read_serialized_po)
     assert type(deserialized_po) == WebPage
-    _compare_webpages(po, deserialized_po)
+    _assert_webpages_equal(po, deserialized_po)
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -60,12 +60,17 @@ def test_serialization(book_list_html_response) -> None:
     url = ResponseUrl(url_str)
 
     serialized_deps = serialize([book_list_html_response, url])
-    assert serialized_deps["HttpResponse"]["body.html"] == bytes(
-        book_list_html_response.body
+    assert serialized_deps["web_poet.page_inputs.http.HttpResponse"][
+        "body.html"
+    ] == bytes(book_list_html_response.body)
+    other_data = json.loads(
+        serialized_deps["web_poet.page_inputs.http.HttpResponse"]["other.json"]
     )
-    other_data = json.loads(serialized_deps["HttpResponse"]["other.json"])
     assert other_data["url"] == url_str
-    assert serialized_deps["ResponseUrl"]["txt"] == url_str.encode()
+    assert (
+        serialized_deps["web_poet.page_inputs.url.ResponseUrl"]["txt"]
+        == url_str.encode()
+    )
 
     po = MyWebPage(book_list_html_response, url)
     deserialized_po = deserialize(MyWebPage, serialized_deps)
@@ -122,13 +127,13 @@ def test_write_data(book_list_html_response, tmp_path) -> None:
     directory.mkdir()
     serialized_deps = serialize([book_list_html_response, url])
     write_serialized_data(serialized_deps, directory)
-    assert (directory / "HttpResponse-body.html").exists()
-    assert (directory / "HttpResponse-body.html").read_bytes() == bytes(
-        book_list_html_response.body
-    )
-    assert (directory / "HttpResponse-other.json").exists()
-    assert (directory / "ResponseUrl.txt").exists()
-    assert (directory / "ResponseUrl.txt").read_text(
+    assert (directory / "web_poet.page_inputs.http.HttpResponse-body.html").exists()
+    assert (
+        directory / "web_poet.page_inputs.http.HttpResponse-body.html"
+    ).read_bytes() == bytes(book_list_html_response.body)
+    assert (directory / "web_poet.page_inputs.http.HttpResponse-other.json").exists()
+    assert (directory / "web_poet.page_inputs.url.ResponseUrl.txt").exists()
+    assert (directory / "web_poet.page_inputs.url.ResponseUrl.txt").read_text(
         encoding="utf-8"
     ) == "http://example.com"
 
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -1,23 +1,18 @@
-from typing import Optional, Type
+from typing import Type
 
 import attrs
 import pytest
 
-from web_poet import (
-    HttpResponse,
-    HttpResponseBody,
-    Injectable,
-    RequestUrl,
-    ResponseUrl,
-    WebPage,
-)
+from web_poet import HttpResponse, HttpResponseBody, ResponseUrl, WebPage
 from web_poet.page_inputs.url import _Url
 from web_poet.serialization import (
-    SerializedData,
+    SerializedLeafData,
     deserialize,
+    deserialize_leaf,
     read_serialized_data,
     register_serialization,
     serialize,
+    serialize_leaf,
     write_serialized_data,
 )
 
@@ -35,28 +30,34 @@ def _assert_urls_equal(u1: _Url, u2: _Url) -> None:
     assert str(u1) == str(u2)
 
 
-def test_serialization() -> None:
+def test_serialization_leaf() -> None:
     data = {"a": "b", "c": 42}
-    serialized_data = serialize(data)
-    deserialized_data = deserialize(dict, serialized_data)
+    serialized_data = serialize_leaf(data)
+    deserialized_data = deserialize_leaf(dict, serialized_data)
     assert data == deserialized_data
 
 
-def test_serialization_unsup() -> None:
+def test_serialization_leaf_unsup() -> None:
     class A:
         pass
 
     with pytest.raises(NotImplementedError):
-        serialize(A())
+        serialize_leaf(A())
 
     with pytest.raises(NotImplementedError):
-        deserialize(A, {})
+        deserialize_leaf(A, {})
 
 
-def test_serialization_webpage(book_list_html_response) -> None:
-    po: WebPage = WebPage(book_list_html_response)
-    serialized_po = serialize(po)
-    deserialized_po = deserialize(WebPage, serialized_po)
+def test_serialization(book_list_html_response) -> None:
+    @attrs.define
+    class MyWebPage(WebPage):
+        url: ResponseUrl
+
+    url = ResponseUrl("http://example.com")
+
+    serialized_deps = serialize([book_list_html_response, url])
+    po = MyWebPage(book_list_html_response, url)
+    deserialized_po = deserialize(MyWebPage, serialized_deps)
     _assert_webpages_equal(po, deserialized_po)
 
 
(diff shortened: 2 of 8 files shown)


tests/test_serialization.py
```diff
@@ -0,0 +1,154 @@
...
+    SerializedData,
+    deserialize,
+    read_serialized_data,
+    register_serialization,
+    serialize,
+    write_serialized_data,
+)
+
+
+def _compare_webpages(p1: WebPage, p2: WebPage) -> None:
```
wRAR, line 25: WebPage has `eq=False`, also `.url` is not comparable anyway.

tests/test_serialization.py
```diff
@@ -0,0 +1,154 @@
...
+    assert p1.response.body == p2.response.body
+    assert p1.response.status == p2.response.status
+    assert p1.response.headers == p2.response.headers
+    assert p1.response._encoding == p2.response._encoding
+    _compare_urls(p1.response.url, p2.response.url)
+
+
+def _compare_urls(u1: _Url, u2: _Url) -> None:
```
wRAR, line 33: _Url has no `__eq__` and different classes being not equal even has its own test.

tests/test_serialization.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+def test_serialization_webpage(book_list_html_response) -> None:
+    po = WebPage(book_list_html_response)
+    serialized_po = serialize(po)
+    deserialized_po = deserialize(WebPage, serialized_po)
+    _compare_webpages(po, deserialized_po)
+
+
+def test_serialization_httpresponse_encoding(book_list_html) -> None:
```
wRAR, line 63: The `encoding`/`_encoding` semantics were unclear to me so I wrote this. Looks like saving and restoring `_encoding` if it's not empty is correct.

web_poet/serialization/functions.py
```diff
@@ -0,0 +1,149 @@
...
+    if o is None:
+        return {}
+
+    assert attrs.has(o)  # FIXME
+    deps = attrs.asdict(o, recurse=False)
+    return {dep_name: serialize(dep) for dep_name, dep in deps.items()}
+
+
+def _deserialize_Injectable(t: Type[Injectable], data: SerializedData) -> Injectable:
+    def _strip_Optional(t: Type) -> Type:
```
wRAR, line 40: This is special support for Optional but other non-trivial field types could need additional support too.

tests/conftest.py
```diff
@@ -20,5 +20,7 @@ def book_list_html():
 def book_list_html_response(book_list_html):
     body = HttpResponseBody(bytes(book_list_html, "utf-8"))
     return HttpResponse(
-        url="http://books.toscrape.com/index.html", body=body, encoding="utf-8"
+        url=ResponseUrl("http://books.toscrape.com/index.html"),
```
kmike, line 23: why is this change needed? I think we want to support previous syntax, there is attrs convertor for that.

wRAR: Without this change the `url` attribute is a `str`.

tests/test_serialization.py
kmike: It's common to name such functions as asserts: assert_pages_equal. With `compare` it's not clear if it returns something, or raises an exception.

tests/conftest.py
kmike: This shouldn't be the case, because of attrs convertor

web_poet/serialization/functions.py
```diff
@@ -0,0 +1,152 @@
...
+from ..page_inputs.url import _Url
+from .api import SerializedData, deserialize, register_serialization, serialize
+
+
+def _serialize_dict(o: dict) -> SerializedData:
+    return {"json": json.dumps(o).encode()}
+
+
+def _deserialize_dict(t: Type[dict], data: SerializedData) -> dict:
+    return t(json.loads(data["json"].decode()))
```
kmike, line 22:
```suggestion
    return t(json.loads(data["json"]))
```

json.loads supports bytes directly

Gallaecio:
> we wanted to save the body as a plain HTML file

What would happen with a binary response?

wRAR:
> What would happen with a binary response?

Any response is saved byte-to-byte into a file, the file will always be named `foo.body.html` though. It should be possible to add some logic to use a different extension in certain cases (checking the Content-Type header maybe?)

Gallaecio:
> checking the Content-Type header maybe?

Ooh… Careful there, that can [get complicated](https://github.com/scrapy/scrapy/pull/5204) :slightly_smiling_face: .

wRAR:
After further discussions we decided that instead of inspecting Injectables to get and recursively serialize their attributes we will use `andi` to (recursively) find dependencies that are themselves Injectable, and we will serialize them by their type, not by the attribute name. We will need some list of "leaf" dependencies to pass to `andi` as the `externally_provided` argument, we may need to hardcode classes provided by web-poet and provide a way to extend it.

wRAR:
The current interface and implementation are different from the initial ones so the earlier comment is partially outdated. 

* We decided that it's not possible to serialize an existing page object because things needed to create it may be no longer available or not possible to find with just introspection. So now the serialization function takes a list of dependencies used to create the page object, and the code that creates it (e.g. in scrapy-poet) is now responsible for saving that list and passing it to the serialization function. The deserialization function still takes a class and returns an object, using `andi` internally to recreate it from the dependencies it deserialized.
* The API now provides top-level `(de)serialize()` functions described in the previous point and `(de)serialize_leaf()` ones that use `singledispatch` to work on specific dependency types and shouldn't be needed to be called directly.
* The serialized data structure is now more simple: its keys are dependency type names and its values are dicts that directly describe file names and contents:
```yaml
"HttpResponse":
    "body.html": <bytes>
    "other.json": <bytes>
"ResponseUrl":
    "txt": <bytes>
```
Then the type name can be joined with the file name to get the full file name on the disk (`HttpResponse-body.html`, `HttpResponse-other.json`, `ResponseUrl.txt`). I don't really like this structure because the code that reads and writes the files needs to know that "simple" deps just use `type.ext` and others use `type-component.ext` and because the code `(de)serialize_leaf()` for complex types like the example `HttpResponse` needs to merge/split the dicts returned/taken by the (de)serialization functions for the body and the `other` dict. It should be possible to return to a hierarchical structure from the earlier implementation that better maps to the file system and is better suitable for recursive serialization. The current structure was implemented because it's trivial to type-annotate, easier to read and because `(de)serialize_leaf()` calling other `(de)serialize_leaf()` should probably be rare.

web_poet/serialization/api.py
```diff
@@ -0,0 +1,122 @@
...
+    return {
+        dep.__class__.__name__: serialize_leaf(dep)
+        for dep in deps
+        if not is_injectable(dep.__class__)
+    }
+
+
+def deserialize(t: Type[InjectableT], data: SerializedData) -> InjectableT:
+    plan = andi.plan(
+        t, is_injectable=is_injectable, externally_provided=externally_provided
```
wRAR, line 111: It looks like we don't really need the statically-defined `externally_provided` list here as we can pass the list of types found in `data` instead. Not sure if this is a correct thing to do though.
(and see below)

web_poet/serialization/api.py
```diff
@@ -0,0 +1,122 @@
...
+    )
+
+    deps: Dict[type, Any] = {}
+
+    for dep_type_name, dep_data in data.items():
+        dep_type = get_dep_type(dep_type_name)
```
wRAR, line 117: `get_dep_type()` here uses the statically-defined `externally_provided` list to find the correct type by its unqualified name. We can switch to qualified names instead.

tox.ini
```diff
@@ -11,6 +11,7 @@ deps =
     pytest-asyncio
     requests
     aiohttp
+    andi
```
kmike, line 14: I think we should remove it from here, it should be installed via setup.py. If it's not, it's an issue, and tests will correctly fail.

web_poet/serialization/api.py
```diff
@@ -0,0 +1,106 @@
...
+def read_serialized_data(directory: Union[str, os.PathLike]) -> SerializedData:
+    result: SerializedData = {}
+    directory = Path(directory)
+    for entry in directory.iterdir():
+        if not entry.is_file():
+            continue
+        if "-" in entry.name:
+            prefix, name = entry.name.split("-", 1)
+        else:
+            prefix, name = entry.name.rsplit(".", 1)
```
kmike, line 31: I think it makes sense to move such logic to a separate function, and have some tests for it; even doctest would do.

web_poet/serialization/api.py
```diff
@@ -0,0 +1,106 @@
...
+    f_ser: SerializeFunction = serialize_leaf.dispatch(t)
+    return f_ser.f_deserialize(t, data)
+
+
+def _get_fqname(t: type) -> str:
+    return f"{t.__module__}.{t.__qualname__}"
+
+
+def serialize(deps: List[Any]) -> SerializedData:
+    # we skip injectable classes though the interface should probably not accept them at all
```
kmike, line 82: If it can prevent user errors, it makes sense to raise an exception.

web_poet/serialization/api.py
```diff
@@ -0,0 +1,106 @@
...
+    return {
+        _get_fqname(dep.__class__): serialize_leaf(dep)
+        for dep in deps
+        if not is_injectable(dep.__class__)
+    }
+
+
+def _load_type(type_name: str) -> Optional[type]:
```
kmike, line 90: Could you please add a test for it? Even a simple doctest is fine.

web_poet/serialization/api.py
```diff
@@ -0,0 +1,106 @@
...
+    module, name = type_name.rsplit(".", 1)
+    mod = import_module(module)
+    return getattr(mod, name, None)
+
+
+def deserialize(t: Type[InjectableT], data: SerializedData) -> InjectableT:
```
kmike, line 96: Can we use `cls` name instead of `t`? Or are there some issues with this?

kmike: This applies to other uses of "type" in this module.

web_poet/serialization/functions.py
```diff
@@ -0,0 +1,108 @@
...
+
+def _serialize_str(o: str) -> SerializedLeafData:
+    return {"txt": o.encode()}
+
+
+def _deserialize_str(t: Type[str], data: SerializedLeafData) -> str:
+    return t(data["txt"].decode())
+
+
+register_serialization(_serialize_str, _deserialize_str)
```
kmike, line 97: Could you please group "primitive" functions, vs functions for concrete dependencies defined by web-poet?

Also, I think it might be good to add some comment about it, on why we need ser/de for str, dict, bytes. My understanding is that we don't really need it (as these types are very unlikely to be leaf dependencies), but we provide it so that all subclasses of these primitive classes don't require registering of custom functions. Is it correct?

There could be an argument for not doing this, and just creating functions for each dependency. But I don't have a strong opinon.

web_poet/serialization/functions.py
```diff
@@ -0,0 +1,108 @@
...
+from .api import (
+    SerializedLeafData,
+    deserialize_leaf,
+    register_serialization,
+    serialize_leaf,
+)
+
+
+def _serialize_dict(o: dict) -> SerializedLeafData:
+    return {"json": json.dumps(o).encode()}
```
kmike, line 15:
```suggestion
    return {"json": json.dumps(o, ensure_ascii=False, sort_keys=True).encode()}
```

* ensure_ascii allows to save space, and also makes it more readable
* sort_keys ensures that there won't be spurious differences in key order, e.g. when we re-create something. It's not very likely in recent Python versions, as dict preserve the insert order, but it still might happen.

web_poet/serialization/functions.py
kmike: Also, just to double-check: is this all working ok if there are 2 subclasses of the same primitive type (e.g. two str subclasses), and both are used? I assume that's working fine.

tests/test_serialization.py
```diff
@@ -0,0 +1,144 @@
...
+def test_serialization_leaf() -> None:
+    data = {"a": "b", "c": 42}
+    serialized_data = serialize_leaf(data)
+    assert isinstance(serialized_data["json"], bytes)
+    assert json.loads(serialized_data["json"]) == data
+    deserialized_data = deserialize_leaf(dict, serialized_data)
+    assert data == deserialized_data
+
+
+def test_serialization_leaf_unsup() -> None:
```
kmike, line 43:
```suggestion
def test_serialization_leaf_unsupported() -> None:
```

tests/test_serialization.py
```diff
@@ -0,0 +1,144 @@
...
+    resp_noenc = HttpResponse(
+        url=ResponseUrl("http://books.toscrape.com/index.html"), body=body
+    )
+    assert resp_noenc._encoding is None
+    deserialized_resp_noenc = deserialize_leaf(HttpResponse, serialize_leaf(resp_noenc))
+    assert deserialized_resp_noenc._encoding is None
+
+
+def test_custom_functions() -> None:
+    class C:
```
kmike, line 100: is fully qualified name works ok in this case?

wRAR: `serialize_leaf` doesn't use that, class names are only used to refer to serialized leaves in the `SerializedData` dict.

web_poet/serialization/api.py
```diff
@@ -69,31 +104,49 @@ def register_serialization(
...
+    result: SerializedData = {}
+    for dep in deps:
+        cls = dep.__class__
+        if is_injectable(cls):
+            raise ValueError(f"Injectable type {cls} passed to serialize()")
+        result[_get_fqname(cls)] = serialize_leaf(dep)
+    return result
 
 
 def _load_type(type_name: str) -> Optional[type]:
```
kmike, line 131: What do you think about renaming it to `_load_type_or_none`?
Also, an option is to just make it raise an exception; no strong opinion on that.

web_poet/serialization/functions.py
wRAR: Functions for `str` and `bytes` are indeed provided for the sake of completion and are not used by the existing code. 

The intended use cases for these (and other similar) primitive type functions are both to cover subclasses (though `HttpResponseBody` is a subclass of `bytes` and we provide functions for it, but just to have a better file extension) and to be able to write `serialize_leaf()` in functions like `_(de)serialize_HttpResponse()`. Though it may be easier to just inline all that code in the second use case.

> is this all working ok if there are 2 subclasses of the same primitive type (e.g. two str subclasses), and both are used? I assume that's working fine.

What use case do you have in mind? Registering functions for both subclasses should work correctly for both `serialize_leaf()` and `deserialize_leaf()`, and just using the functions provided for `str` should also work correctly. If you mean using both in a dependency list, then it should never matter that they are subclasses of `str`.

kmike: > What use case do you have in mind?

Let's say we register `str`, and expect it to cover two subclasses: BrowserHtml(str) and MyBrowserHtml(str), without registering anything for these subclasses explicitly. My question was if this works correctly, but it seems it does, as the keys in dictionary are subclass names, and there is no clash.

Overall, I'm not 100% sure, but slightly prefer removing str, bytes, dict serializers, and supporting classes which we provide directly instead. I'm not sure what's the use case for making some arbitrary str or dict subclasses serializable, in context of page objects.

wRAR: I've removed those and inlined them in the HttpResponse implementation. I think it looks fine.

Note that now we handle HttpResponseBody manually instead of using its specific methods even though they still exists.

kmike:
The implementation looks good, great job @wRAR! I think we can merge when CI is green. I'd also like to explore using classes instead of 2 functions; let me send a PR, to see how it looks.

tests/test_serialization.py
```diff
@@ -0,0 +1,153 @@
...
+    serialize,
+    serialize_leaf,
+    write_serialized_data,
+)
+
+
+def _assert_webpages_equal(p1: WebPage, p2: WebPage) -> None:
+    assert p1.response.body == p2.response.body
+    assert p1.response.status == p2.response.status
+    assert p1.response.headers == p2.response.headers
```
BurnzZ, line 24: Similar to when asserting if the URL types are the same, I think we could also compare if they are of the same type (e.g. `HttpResponseBody` == `HttpResponseBody`.)

web_poet/serialization/api.py
```diff
@@ -0,0 +1,166 @@
...
+    >>> _load_type("foo.bar")
+    Traceback (most recent call last):
+     ...
+    ValueError: Unknown type foo.bar
+    """
+    module, name = type_name.rsplit(".", 1)
+    try:
+        mod = import_module(module)
+    except ModuleNotFoundError:
+        raise ValueError(f"Unknown type {type_name}")
```
BurnzZ, line 151:
```suggestion
        raise ValueError(f"Unable to import module: {module}")
```

web_poet/serialization/api.py
```diff
@@ -0,0 +1,166 @@
...
+    """Return the type by its fully qualified name.
+
+    >>> _load_type("decimal.Decimal")
+    <class 'decimal.Decimal'>
+    >>> _load_type("web_poet.pages.WebPage")
+    <class 'web_poet.pages.WebPage'>
+    >>> _load_type("decimal.foo")
+    Traceback (most recent call last):
+     ...
+    ValueError: Unknown type decimal.foo
```
BurnzZ, line 141:
```suggestion
    ValueError: Unable to import module: decimal.foo
```

web_poet/serialization/api.py
```diff
@@ -0,0 +1,166 @@
...
+    >>> _load_type("foo.bar")
+    Traceback (most recent call last):
+     ...
+    ValueError: Unknown type foo.bar
```
BurnzZ, line 145:
```suggestion
    ValueError: Unable to import module: foo.bar
```

web_poet/serialization/api.py
wRAR: This one should stay though?

BurnzZ: Ahh yes, you're right! 👍

wRAR:
I've explored different options related to typing, and workarounds for custom function attributes (making a protocol for the function) are hard to combine with the singledispatch-specific typing. I'm also not sure why the singledispatch-specific typing for `serialize_leaf` doesn't work for its `dispatch()` and `register()` methods, I think it should. Finally,  `plan.final_kwargs()`, which takes `Mapping[Callable, Any]`, doesn't want to take `Dict[type, Any]` so I've changed the var type to be more general.

kmike (approved):
👍 I think that's fine to merge this feature without docs, because it starts to make more sense only later, when we either add testing feature to web-poet, or at switch scrapy-poet cache implementation to web-poet's serialization.

BurnzZ (approved):

kmike:
Thanks @wRAR!

PR #108, opened 2022-11-25
Rework the tutorial

Gallaecio:
Reworked the tutorial to cover most web-poet topics and current best practices.


docs/intro/tutorial.rst
```diff
@@ -1,141 +1,434 @@
...
+web-poet uses itemadapter_ for item class support, which means that any kind of
+item class can be used. In this tutorial, you will use attrs_ to define your
+item class.
+
+.. _attrs: https://www.attrs.org/en/stable/
+.. _itemadapter: https://github.com/scrapy/itemadapter
+
+Copy the following code into ``tutorial-project/tutorial/items.py``:
+
+.. literalinclude:: /../tutorial-project/tutorial/items.py
```
kmike, line 90: What do you think about moving tutorial-project to the docs folder? The code looks a bit strange as a general-purpose example (imports in the middle of files, etc).

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,434 @@
...
+   :language: python
+   :lines: 1-6
+
+This code defines a ``Book`` item class, with a single, required, string
+``title`` attribute to store the book title.
+
+``Book`` is a minimal class designed specifically for this tutorial. In real
+web-poet projects, you will often define item classes with many more
```
kmike, line 98:
```suggestion
web-poet projects, you will usually define item classes with many more
```

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,434 @@
...
+
+In this tutorial you use ``web_poet.example.get_item``, which is a simple,
+incomplete implementation of the web-poet standard, built specifically for this
+tutorial, for demonstration purposes. In real projects, use instead an actual
+web-poet framework, like `scrapy-poet`_.
+
+``web_poet.example.get_item`` serves to illustrate the power of web-poet. Once
+you have defined your page object class, a web-poet framework only needs 2
+inputs from you: the URL from which you want to extract data, and the import
+path of the Python modules where you define your page object classes.
```
kmike, line 211: I think we should require 2 inputs:

* URL
* Item type.

Import path to modules should be something which is handled on setup time, not something which is passed to every get_item call. That's global shared state (modules currently imported in the Python process), and we should be explicit about it; it's not good that a call to get_item may implicitly affect other get_item calls.

As for asking to specifiy the item class, it could be error-prone otherwise. For example, the modules may contain page objects which extract different item types from the same pages. Let me elaborate on this.

I think it's quite common that you can't write a specific URL pattern for an item. So, handle_urls could be annotated just with a domain name, without trying to limit it further by path, etc, and that's perfectly fine. In most cases the crawling code knows what kind of content to expect after following the link, so it should be able to use right page objects on right URLs; it may depend no only on the URL, but also on CSS/XPath of the link extracted. 

While being more precise with handle_urls is fine, it may also backfire sometimes, because this logic can get split between crawling (link extraction code) and handle_urls. 

The main purpose for using URL patterns in handle_urls is to distinguish between different page objects which return the same item class, not to infer the item class returned for an URL.

docs/tutorial-project/tutorial/items.py
```diff
@@ -0,0 +1,15 @@
+from attrs import define
+
+
+@define(kw_only=True)
```
kmike, line 4: What do you think about removing `kw_only=True`, to simplify the example?

web_poet/example.py
```diff
@@ -0,0 +1,76 @@
...
+def _get_page(
+    url: str,
+    page_class: Type[ItemPage],
+    *,
+    page_params: Optional[Dict[Any, Any]] = None,
+) -> ItemPage:
+    # We make quite a few assumptions here that prevent this function from
+    # working in scenarios where those assumptions are wrong. These shortcuts
+    # have been taken to save development time, providing the minimum
+    # implementation that works well enough for the tutorial.
```
kmike, line 43: I think we should use andi even here, even if it's a bit more complicated. Hopefully, it wouldn't be too bad.

Gallaecio: https://github.com/scrapinghub/web-poet/pull/108/commits/94cf9a9886545ff20af204f99ad806aee13ddb12 switches to andi.

However, besides that, it is still rather tutorial-specific (only needed “providers” are implemented, and the HttpClient object only supports the interface subset used in the tutorial), and not too clean (e.g. we could organize providers in a way closer to scrapy-poet, and have a cleaner for look to build instances).

I’m not sure how far I should go with the implementation within the scope of these tutorial changes. Although the hard part, me getting to understand the basics of andi, is done now.

web_poet/example.py
```diff
@@ -0,0 +1,86 @@
...
+            rule.use: rule.for_patterns
+            for rule in default_registry.get_rules()
+            if (rule.to_return is None or rule.to_return == item_class)
+        }
+    )
+    return url_matcher.match(url)
+
+
+def _get_http_response(url: str) -> HttpResponse:
+    response = get(url)
```
kmike, line 30: nitpick: I think requests.get could be more readable

web_poet/example.py
```diff
@@ -0,0 +1,86 @@
...
+    ``@handle_urls`` to take effect.
+
+    This function is an example of a minimal, incomplete web-poet framework
+    implementation, intended for use in the web-poet tutorial.
+    """
+    page_class = _get_page_class(url, item_class)
+    if page_class is None:
+        raise ValueError(f"No page object class found for URL: {url}")
+    page = _get_page(url, page_class, page_params=page_params)
+    return run(page.to_item())
```
kmike, line 86: should we use ensure_awaitable here?

web_poet/example.py
```diff
@@ -0,0 +1,86 @@
...
+    async def get(self, url: str) -> HttpResponse:
+        return _get_http_response(url)
+
+
+def _get_page_class(url: str, item_class: Type) -> Type[ItemPage]:
+    url_matcher = URLMatcher(
+        {
+            rule.use: rule.for_patterns
+            for rule in default_registry.get_rules()
+            if (rule.to_return is None or rule.to_return == item_class)
```
kmike, line 23: why do you need to consider classes with to_return==None?

web_poet/example.py
```diff
@@ -0,0 +1,86 @@
+from asyncio import run
```
kmike, line 1: Could you please add a docstring to this module? I think we should state here that it's a toy example for the purpose of making it easier for the tutorial, and discourage people from using it in production or relying on the API in this module. What do you think?

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
+
+-   ``BookPage`` declares ``Book`` as its return type.
+
+    :class:`~web_poet.pages.WebPage`, like its parent class
+    :class:`~web_poet.pages.ItemPage`, is a :ref:`generic class <generics>`
+    that accepts a type parameter. Unlike most generic classes, however, the
+    specified type parameter is used for more than type hinting: it determines
+    the item class that is used to store the data that fields return.
+
+-   ``BookPage`` is decorated with :meth:`~web_poet.overrides.handle_urls`,
```
kmike, line 156: the link to handle_urls doesn't work; probably it should point to .rules module

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
+    (``Book``).
+
+#.  It inspects the inputs declared by ``BookPage``, and builds an instance of
+    ``BookPage`` with the required inputs.
+
+    ``BookPage`` is a :class:`~web_poet.pages.WebPage` subclass, and
+    :class:`~web_poet.pages.WebPage` declares an attribute named ``response``
+    of type :class:`~web_poet.page_inputs.HttpResponse`. Your web-poet
+    framework sees this, and creates an
+    :class:`~web_poet.page_inputs.HttpResponse` object from the input URL as a
```
kmike, line 242: HttpResponse link doesn't work, not sure why

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
+   :lines: 14-15, 17, 19-26, 28-35, 39-52
+   :emphasize-lines: 1-3, 9, 11-12, 18-33
 
-Final Result
-============
+In the code above:
 
-Our simple Python script might look like this:
+-   You declare a new input in ``CategorizedBookPage``, ``http``, of type
+    :class:`~web_poet.page_inputs.client.HttpClient`.
```
kmike, line 358: I wonder if this point could be more clear if we don't use WebPage, and use ItemPage directly. No strong opinon though.

web_poet/example.py
Gallaecio: I assumed, if no `to_return` class is specified, that they are valid options whatever the value of  `to_return` is. But I guess it makes sense to consider that, by passing a specific item class to `get_item`, only page object classes which report returning such an item class should be elegible.

kmike: > I assumed, if no to_return class is specified, that they are valid options whatever the value of to_return is. 

Let's say you request MyCustomProduct and MyCustomproductList items on example.com, and there is some a `handle_urls("example.com")` page object. Most likely, it knows nothing about either of the MyCustom items. And you can't possibly use it to return both items. So, if to_return is not specified, it means we shouldn't use this page object as a way to return an item.

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
-                "links": self.links,
-            }
+    tutorial-project/
+        run.py
+        tutorial/
+            __init__.py
+            items.py
+            pages/
+                __init__.py
+                books_toscrape_com.py
```
BurnzZ, line 66: What do you think about using GNU tree's formatting on this?

```
.
└── tutorial-project
    ├── run.py
    └── tutorial
        ├── __init__.py
        ├── items.py
        └── pages
            ├── __init__.py
            └── books_toscrape_com.py
```

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
+   :language: python
+   :lines: 1-11
+
+In the code above:
+
+-   You define a page object class named ``BookPage`` by subclassing
+    :class:`~web_poet.pages.WebPage`.
+
+    It is possible to create a page object class subclassing instead the
+    simpler :class:`~web_poet.pages.ItemPage` class, but
```
BurnzZ, line 133:
```suggestion
    simpler :class:`~web_poet.pages.ItemPage` class. However, the
```

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
+        :meth:`~web_poet.pages.WebPage.xpath`, that make it easier to write
+        web data extraction code.
+
+-   ``BookPage`` declares ``Book`` as its return type.
+
+    :class:`~web_poet.pages.WebPage`, like its parent class
+    :class:`~web_poet.pages.ItemPage`, is a :ref:`generic class <generics>`
+    that accepts a type parameter. Unlike most generic classes, however, the
+    specified type parameter is used for more than type hinting: it determines
+    the item class that is used to store the data that fields return.
```
BurnzZ, line 154: Let's also mention that the `.to_item()` method that `WebPage` inherits from `ItemPage` is responsible for returning the said `Book` item.

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
+    :ref:`Fields <fields>` are methods of page object classes, preferably async
+    methods, decorated with :meth:`~web_poet.fields.field`. Fields define the
+    logic to extract a specific piece of information from the input of your
+    page object class.
+
+    ``BookPage.title`` extracts the title of a book from a book details
+    webpage. Specifically, it extracts the text from the first ``h1`` element
+    on the input HTTP response.
+
+    Here ``title`` is not an arbitrary name, it was chosen specifically to
```
BurnzZ, line 176:
```suggestion
    Here, ``title`` is not an arbitrary name. It was chosen specifically to
```

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
+    ``BookPage[CategorizedBook]`` would not work.
+
+    So instead you use :class:`~web_poet.pages.Returns`, a special, generic
+    class that you can inherit to re-define the output type of your page object
+    subclasses.
+
+After you update your ``tutorial-project/run.py`` script to request a
+``CategorizedBook`` item:
+
+.. literalinclude:: /tutorial-project/run.py
```
BurnzZ, line 307: Ahhh that's super cool! TIL 💯

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
-=====================
-web-poet on a surface
-=====================
+========
+Tutorial
+========
 
-In this tutorial, we’ll assume that web-poet is already installed on your
-system.
+In this tutorial you will learn to use web-poet as you write web data
```
BurnzZ, line 8:
```suggestion
In this tutorial, you will learn to use web-poet as you write web data
```

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
+web-poet framework, like `scrapy-poet`_.
+
+``web_poet.example.get_item`` serves to illustrate the power of web-poet: once
+you have defined your page object class, a web-poet framework only needs 2
+inputs from you: the URL from which you want to extract data, and the desired
+output item class.
+
+Notice that you must also call :func:`~web_poet.rules.consume_modules` once
+before your first call to ``get_item``. ``consume_modules`` ensures that the
+specified Python modules are loaded. You pass ``consume_modules`` the modules
```
BurnzZ, line 221:
```suggestion
specified Python modules are loaded. You pass ``consume_modules`` the module paths
```

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
+where you define your page object classes. Upon load,
```
BurnzZ, line 222:
```suggestion
where you define your page object classes. After loading the passed module paths,
```

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
+:meth:`~web_poet.overrides.handle_urls` decorators register the page object
+classes that they decorate into the default registry, which ``get_item`` uses
```
BurnzZ, line 224:
```suggestion
classes that they decorate into ``web_poet.default_registry``, which ``get_item`` uses
```

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
+
+-   You define a new page object class: ``CategorizedBookPage``.
 
+-   ``CategorizedBookPage`` subclasses ``BookPage``, inheriting its ``title``
+    field, and defining a new one: ``category``.
 
-    response = requests.get("http://books.toscrape.com")
+    ``CategorizedBookPage`` does *not* define a ``category_rank`` field
+    yet, you will add it later on. For now, the default value defined in
+    ``CategorizedBook`` for ``category_rank``, ``None``, will be used.
```
BurnzZ, line 289:
```suggestion
    ``CategorizedBook`` for ``category_rank`` will be  ``None``.
```

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
 
-Creating Page Input
-===================
+-   ``CategorizedBookPage`` indicates that it returns a ``CaterogizedBook``
```
BurnzZ, line 291:
```suggestion
-   ``CategorizedBookPage`` indicates that it returns a ``CategorizedBook``
```

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
 
-Let's create a Python file where we'll put our first Page Object implementation.
-This Page Object will be responsible for extracting book links from the book
-list page on `books.toscrape.com <http://books.toscrape.com/>`_.
+.. _Python: https://docs.python.org/
 
-.. code-block:: python
 
-    from web_poet.pages import WebPage
+Create a project folder
```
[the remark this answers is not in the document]
Gallaecio: I’m applying the change, but for the record I think it is OK to use the word folder even when there is no GUI involved, and I think it can be less confusing to some users.

docs/intro/tutorial.rst
Gallaecio: I am unsure about this. It is covered in the following section, in the 3rd bullet point at the end, and I think that may be the best point to cover this.

docs/intro/tutorial.rst
Gallaecio: I’m not sure about this one.

docs/intro/tutorial.rst
Gallaecio: My first attempt at committing this change locally:

```$ git comimt -am "CaterogizedBook → CategorizedBook"```

:facepalm:

File: web_poet/example.py (hunk 9 above)
Comment (Gallaecio) [reply web_poet/example.py:8]: Added a warning message on import as well.

File: docs/intro/tutorial.rst (hunk 12 above)
Comment (Gallaecio) [reply docs/intro/tutorial.rst:358]: While I think you are right about its effect on this point, I think it is good for the tutorial to cover WebPage earlier, and showcase e.g. `self.css`. No strong opinion either, though.

File: docs/intro/tutorial.rst [hunk 25]
```diff
@@ -1,141 +1,446 @@
...
+web-poet does not limit how you structure your web-poet web data extraction
+code, beyond the limitations of Python itself.
+
+However, in this tutorial you will use a specific project folder structure
+designed with web-poet best practices in mind. Consider using the same project
+folder structure in all your web-poet projects.
+
+First create your project folder: ``tutorial-project``.
+
+Within ``tutorial-project``, create:
```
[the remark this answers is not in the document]
Gallaecio: Went with including the word folder. I actually like using `/` for this purpose, but I would then use it everywhere, and it felt too verbose to use it in conjugation with the world folder in some of the other references in this section.

tests/test_serialization.py
```diff
@@ -0,0 +1,167 @@
+from typing import Type
```
kmike, line 1: That's weird - why is this file appearing in diff?

web_poet/serialization/__init__.py
```diff
@@ -0,0 +1,13 @@
+from . import functions  # needed to run register functions
```
kmike, line 1: It seems there are a few more files in diff which shouldn't be there

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
-    from web_poet.pages import WebPage
+Create a project directory
+==========================
+
+web-poet does not limit how you structure your web-poet web data extraction
+code, beyond the limitations of Python itself.
+
+However, in this tutorial you will use a specific project directory structure
+designed with web-poet best practices in mind. Consider using the same project
+directory structure in all your web-poet projects.
```
kmike, line 27: I think the structure might depend on the framework used, e.g. it might be different in case of scrapy-poet. Should we say "similar" instead of "the same"?

kmike: The structure looks fine overall, but I wonder what run.py method should contain in user projects.

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
+
+
+Use your page object class
+==========================
+
+Now that you have a page object class defined, it is time to use it.
+
+First, install andi_ and requests_, which are required by ``web_poet.example``.
+
+.. _andi: https://github.com/scrapinghub/andi
```
kmike, line 188: andi is now in install_requires, right?

kmike (approved):
+1 to merge after resolving weird diff issues. This tutorial is way better than the one we have currently in master.

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
+
+.. literalinclude:: /tutorial-project/tutorial/items.py
+   :language: python
+   :lines: 1-6
+
+This code defines a ``Book`` item class, with a single required ``title``
+string attribute to store the book title.
+
+``Book`` is a minimal class designed specifically for this tutorial. In real
+web-poet projects, you will usually define item classes with many more
```
BurnzZ, line 98: Do you think if we can mention (as an example) and add a link to zyte-common-items here?

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
+In the code above:
+
+-   You define a page object class named ``BookPage`` by subclassing
+    :class:`~web_poet.pages.WebPage`.
+
+    It is possible to create a page object class subclassing instead the
+    simpler :class:`~web_poet.pages.ItemPage` class. However,
+    :class:`~web_poet.pages.WebPage`:
+
+    -   Indicates that your page object class requires an HTTP response as
```
BurnzZ, line 136:
```suggestion
    -   Indicates that your page object class requires an :class:`web_poet.page_inputs.http.HttpResponse` as
```

docs/intro/tutorial.rst
```diff
@@ -1,141 +1,446 @@
...
+
+``web_poet.example.get_item`` serves to illustrate the power of web-poet: once
+you have defined your page object class, a web-poet framework only needs 2
+inputs from you: the URL from which you want to extract data, and the desired
+output item class.
+
+Notice that you must also call :func:`~web_poet.rules.consume_modules` once
+before your first call to ``get_item``. ``consume_modules`` ensures that the
+specified Python modules are loaded. You pass ``consume_modules`` the import
+paths of the modules where you define your page object classes. After loading
```
BurnzZ, line 222:
```suggestion
paths of the modules where your page object classes are defined. After loading
```

tests/test_example.py
```diff
@@ -0,0 +1,43 @@
...
+        class Page(ItemPage[Item]):
+            @field
+            async def foo(self):
+                return "bar"
+
+        item = get_item("file:///dev/null", Item)
+        assert isinstance(item, Item)
+        assert item.foo == "bar"
+    finally:
+        default_registry._rules.pop()
```
BurnzZ, line 24: We'll need to update this `RulesRegistry._rules` has became a `dict`. https://github.com/scrapinghub/web-poet/pull/112/files#diff-22f478b54a07b372ea47ead887ca01488362ad501692903a9d303511f9d1fe24L118

web_poet/example.py
```diff
@@ -0,0 +1,100 @@
...
+    ),
+    UserWarning,
+    stacklevel=2,
+)
+
+
+class _HttpClient:
+    async def get(self, url: str) -> HttpResponse:
+        return _get_http_response(url)
+
```
BurnzZ, line 34: Let's have codecov ignore this file so it wouldn't affect the coverage values. 

`.codecov.yml`:
```
ignore:
  - "web_poet/example.py"
```

Ref: https://docs.codecov.com/docs/ignoring-paths

web_poet/example.py
```diff
@@ -0,0 +1,100 @@
...
+
+def _get_page_class(url: str, item_class: Type) -> Type[ItemPage]:
+    url_matcher = URLMatcher(
+        {
+            rule.use: rule.for_patterns
+            for rule in default_registry.get_rules()
+            if rule.to_return == item_class
```
BurnzZ, line 41: We have a new `.page_object_for_item()` method in https://github.com/scrapinghub/web-poet/pull/112/files#diff-22f478b54a07b372ea47ead887ca01488362ad501692903a9d303511f9d1fe24R383-R394

tests/test_serialization.py
Gallaecio: I have no idea what sort of merge mess I made, but the latest merge seems to have dealt with it.

docs/intro/tutorial.rst
Gallaecio: > I wonder what run.py method should contain in user projects.

Nothing, it is described as “a file specific to this tutorial” on its first mention.

BurnzZ (approved):
+1 to merge. :100:

PR #109, opened 2022-11-27
FieldsMixin: ignore AttributeError from delattr

Gallaecio:
I found that I needed this change to be able to [use a mixin for `ItemPage` subclasses that implements fields](https://github.com/zytedata/zyte-common-items/pull/19), since the mixin itself is not (*cannot be*, I think) an `ItemPage` subclass.

### DIFF
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -3,7 +3,6 @@ import random
 
 import attrs
 import pytest
-from attrs import define
 
 from tests.po_lib_to_return import (
     CustomProductPage,
@@ -455,7 +454,6 @@ async def test_field_processors_async() -> None:
 
 
 def test_field_mixin() -> None:
-    @define
     class A(ItemPage):
         @field
         def a(self):
@@ -466,7 +464,6 @@ def test_field_mixin() -> None:
         def mixin(self):
             return None
 
-    @define
     class B(Mixin, A):
         @field
         def b(self):
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -451,3 +451,15 @@ async def test_field_processors_async() -> None:
 
     page = Page()
     assert await page.name == "namex"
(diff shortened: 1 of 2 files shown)


web_poet/fields.py
```diff
@@ -44,7 +45,8 @@ def __init_subclass__(cls, **kwargs):
             fields = {**base_class_fields, **this_class_fields}
             setattr(cls, _FIELDS_INFO_ATTRIBUTE_READ, fields)
             if hasattr(cls, _FIELDS_INFO_ATTRIBUTE_WRITE):
-                delattr(cls, _FIELDS_INFO_ATTRIBUTE_WRITE)
+                with suppress(AttributeError):
+                    delattr(cls, _FIELDS_INFO_ATTRIBUTE_WRITE)
```
kmike, line 49: would it make sense to remove the hasattr call?

tests/test_fields.py
```diff
@@ -451,3 +451,15 @@ async def name(self):  # noqa: D102
...
+
+
+def test_field_mixin() -> None:
+    class Mixin:
+        @field
+        def a(self):
+            return "a"
+
+    class A(ItemPage, Mixin):
+        pass
```
kmike, line 463: Could you please change the test to check that A class can define fields as well? Probably it also makes sense to check that A can inherit from 2 base classes which define fields.

tests/test_fields.py
```diff
@@ -3,6 +3,7 @@
 
 import attrs
 import pytest
+from attrs import define
```
kmike, line 6: a nitpick: it seems none of the defines are needed

Gallaecio: I keep forgetting how that works :facepalm:

kmike:
Thanks @Gallaecio!

PR #110, opened 2022-12-13
Pytest runner for PO tests

wRAR:
This will need tests and docs.

The expected file structure:

```yaml
<some dir, maybe "tests" or "fixtures">:
    fully.qualified.PO:
        <URL fingerprint or other identifier>:
            po:
                <files created by serializing PO deps for a specific PO and URL>
            output.json
        <another URL fingerprint>:
            ...
    another.PO:
        ...
```
(see also the image)

The runner finds `output.json` files and discovers PO names and dirs from that. Each PO dir is a pytest "file" and each URL fingerprint inside (whose values the runner doesn't use in any way except as identifiers) is a pytest "item" (test case).
Screenshot_20221213_172655


wRAR:
Not sure if we need the nested `packaging` module, as I don't know if we will have some other code to put inside.

web_poet/serialization/api.py
```diff
@@ -129,18 +129,18 @@ def serialize(deps: List[Any]) -> SerializedData:
     return result
 
 
-def _load_type(type_name: str) -> type:
+def load_type(type_name: str) -> type:
```
kmike, line 132: What do you think about renaming it to load_class? See e.g. https://docs.python.org/3/library/types.html module - it is called "types", and there are CoroutineType, etc., but the functions use "class".

kmike: It seems "types" are used more in context of some definitions, which then used in isinstance/issubclass/typing checks. It seems when you talk about objects (e.g. class objects), "class" is more commonly used.

web_poet/testing/pytest.py
```diff
@@ -0,0 +1,89 @@
...
+
+import pytest
+
+from web_poet import ItemPage
+from web_poet.serialization import SerializedDataFileStorage, deserialize, load_type
+from web_poet.utils import ensure_awaitable
+
+
+class WebPoetFile(pytest.File):
+    """Represents a directory containing test case sibdirectories for one PO."""
```
kmike, line 14:
```suggestion
    """Represents a directory containing test case subdirectories for one page object."""
```

web_poet/testing/pytest.py
```diff
@@ -0,0 +1,89 @@
...
+    def po_name(self) -> str:
+        """The type name of the PO being tested."""
+        if not self.parent:
+            raise ValueError("WebPoetItem has no parent")
+        return self.parent.name
+
+    @property
+    def input_path(self) -> Path:
+        """The po subdirectory path."""
+        return self.path / "po"
```
kmike, line 42: I think "dependencies" or "deps" or "inputs" can be better names here.

web_poet/testing/pytest.py
```diff
@@ -0,0 +1,89 @@
...
+        assert output == expected_output
+
+
+_found_po_dirs: Set[Path] = set()
+
+
+def pytest_collect_file(
+    file_path: Path, path: Any, parent: pytest.Collector
+) -> Optional[pytest.Collector]:
+    if file_path.name == "output.json":
```
kmike, line 78: It seems "output.json" is used a few times, probably better to create a constant for it.

wRAR: It (and "po") will be also used by the code that generates the fixtures, e.g. in scrapy-poet, do you think we should put both in public constants somewhere, or maybe make some helpers?

kmike:
Overall structure makes sense to me 👍 
+1 to use "fixtures" name.

I'm a bit worried about path length limitations, it seems we might hit them here. 
It also seems that in most cases, the dependencies are going to be the ones which are defined by web-poet, and available as top level web_poet imports. What do you think about having a shortcut: for import paths without dots assume `web_poet.{path}`? E.g. `HttpResponse-body.html` would be transformed to `web_poet.HttpResponse`.

If we do so, it could be ok to have all files in a flat folder as well, though it's not clear which is better:

```
fixtures
    my_library.pages.BooksProductPage
        23476fadf2342fdadfadfafsdf
            HttpResponse-body.html
            HttpResponse-other.json
            ResponseUrl.txt
            output.json  
```

kmike:
Regarding URL fingerprints - do we want a generated test to overwrite a previous test if it's for the smae URL? If no, probably it makes sense to use something else as an identifier. It could be a random string, incrementing numbers; it may also include a part of url (e.g. domain), but I'm not sure about this. 

Something like that?

```
fixtures
    my_library.pages.BooksProductPage
        case-1
            inputs
                HttpResponse-body.html
                HttpResponse-other.json
                ResponseUrl.txt
            output.json 
```

It doesn't affect the code here, but might affect docs. We may even encourage people to rename these `case-1` folders manually to use readable names, when appropriate. We should also ensure that these names are used as a part of pytest id, so that it's clear which test is failing (probably it's already the case, but I haven't checked it).

wRAR:
I've added `save_fixture()`, that takes the input deps as a list, the output item as something that ItemAdapter can consume (or we could expect the resulting dict here instead), and the path to the test case directory (so naming it is a responsibility of the caller, but we could instead expect the PO directory and name the test case directory themselves based e.g. on the next available number). Let me know what do you think.

> What do you think about having a shortcut: for import paths without dots assume web_poet.{path}? E.g. HttpResponse-body.html would be transformed to web_poet.HttpResponse.

I think this is fine, though we will need to know that `HttpResponse` is something that can use this short name when being serialized.

> If we do so, it could be ok to have all files in a flat folder as well, though it's not clear which is better:

I think this is technically possible (we can add some ignore list to the code that reads inputs and in the future add e.g. meta.json there), but may be slightly confusing to humans looking at the directory. It also means coupling the code that reads serialized inputs and the code that works with tests, as the former will need to know that "output.json" is something used by the latter.

> We should also ensure that these names are used as a part of pytest id, so that it's clear which test is failing (probably it's already the case, but I haven't checked it).

I think it works now, the output is like this:

> fixtures/scrapy_tests.po.BTSBookPage::8545295f30386e6d0cf6f150de6c99ac9f5d2e36 <- fixtures/scrapy_tests.po.BTSBookPage/8545295f30386e6d0cf6f150de6c99ac9f5d2e36 PASSED [ 33%]

(looks like it prints the directory name in addition to the test IDs? we may want to disable this if possible, as the information there is the same)

kmike:
> I think this is fine, though we will need to know that HttpResponse is something that can use this short name when being serialized.

Something like this?

```py
import web_poet

_SHORTCUTS = {getattr(web_poet, name) for name in dir(web_poet) if name[0].isupper()}
# ...
if cls in shortcuts:
   return cls.__qualname__
else:
   return f"{cls.__module__}.{cls.__qualname__}"
```

By the way, there is also https://github.com/scrapinghub/web-poet/blob/a44c426cd2be1cd97ec16af44586b660faa0bcf9/web_poet/utils.py#L13; it seems we can have one implementation.

web_poet/testing/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+def save_fixture(
+    base_directory: Union[str, os.PathLike], inputs: Iterable[Any], item: Any
+) -> None:
+    inputs_dir = Path(base_directory, INPUT_DIR_NAME)
+    inputs_dir.mkdir(parents=True)
+    serialized_inputs = serialize(inputs)
+    storage = SerializedDataFileStorage(inputs_dir)
+    storage.write(serialized_inputs)
+    with Path(base_directory, OUTPUT_FILE_NAME).open("w") as f:
+        json.dump(ItemAdapter(item).asdict(), f)
```
kmike, line 23: let's use ensure_ascii=False. Not sure about sort_keys=True, probably makes sense not to use it.

kmike: I'd also use some indent option here, indent=2 or indent=4.

web_poet/testing/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+
+from itemadapter import ItemAdapter
+
+from web_poet.serialization import SerializedDataFileStorage, serialize
+
+INPUT_DIR_NAME = "inputs"
+OUTPUT_FILE_NAME = "output.json"
+
+
+def save_fixture(
```
kmike, line 14: All other __init__ files in web-poet are import-only - what do you think about moving it out of here?

web_poet/testing/pytest.py
```diff
@@ -0,0 +1,90 @@
...
+        assert output == expected_output
+
+
+_found_po_dirs: Set[Path] = set()
+
+
+def pytest_collect_file(
+    file_path: Path, path: Any, parent: pytest.Collector
+) -> Optional[pytest.Collector]:
+    if file_path.name == OUTPUT_FILE_NAME:
```
kmike, line 79: output.json is quite a common name; do you see any issues with that?

kmike:
> As for web_poet.utils._clspath, can we change its implementation or whould we just call it for the general case?

No strong opinion. It'd be good to have `f"{cls.__module__}.{cls.__qualname__}"` in a single place. It may still be 2 functions.

web_poet/testing/pytest.py
wRAR: Yeah, but as far as I can see pytest collection works on files and there seems to be no better way to identify a test case. We could add more checks here, e.g. for the inputs dir. Or we could use some prefix for output.json, a potential meta.json etc. and move deps into the same dir, so that ignoring them when readins serialized deps is easier and seeing them when looking at the dir content manually is hopefully easier too.

wRAR: > Sorry, I didn't quite catch this. Could you please elaborate?

This is related to the earlier discussion in this PR: we could put output.json and the serialized deps in the same directory and I suggested we could add some other files there which are not serialized deps. All those files could have some common prefix.

setup.py
```diff
@@ -17,6 +17,7 @@
     package_data={
         "web_poet": ["py.typed"],
     },
+    entry_points={"pytest11": ["web-poet = web_poet.testing.pytest"]},
```
BurnzZ, line 20: Should we also include `pytest` as a dependency?

web_poet/testing/utils.py
```diff
@@ -0,0 +1,23 @@
...
+import os
+from pathlib import Path
+from typing import Any, Iterable, Union
+
+from itemadapter import ItemAdapter
+
+from web_poet.serialization import SerializedDataFileStorage, serialize
+
+INPUT_DIR_NAME = "inputs"
+OUTPUT_FILE_NAME = "output.json"
```
BurnzZ, line 11: I think I might be missing it but what's the current proposed structure inside this file?

wRAR: It's a `json.dumps` of the to_item() result.

setup.py
wRAR: I don't know, I even thought whether we should split this into a separate pytest-web-poet package to avoid this (which is probably too much work).

wRAR:
> Regarding URL fingerprints - do we want a generated test to overwrite a previous test if it's for the smae URL? If no, probably it makes sense to use something else as an identifier. It could be a random string, incrementing numbers; it may also include a part of url (e.g. domain), but I'm not sure about this.

Now save_fixture takes an optional dir name and finds the next available `test-%d` if it's not passed.

wRAR:
> If we use increasing numbers (which seems to be quite friendly), we should sort the tests according to numeric value, not using regular string-based sorting. It can be solved later

I've added a method that can be improved later.

setup.py
wRAR: So, I think this is not needed, as if you want to use the plugin you should already have pytest installed, and if you don't need it you can import even `web_poet.testing` without installing pytest.

web_poet/testing/pytest.py
```diff
@@ -0,0 +1,100 @@
...
+                if item.input_path.is_dir() and item.output_path.is_file():
+                    result.append(item)
+        return self.sorted(result)
+
+
+class WebPoetItem(pytest.Item):
+    """Represents a directory containing one test case."""
+
+    @property
+    def po_name(self) -> str:
```
kmike, line 41: Could you please rename this method (and similar methods) to avoid using "po" shorcut? We often use "page" in web-poet, but other names might be more appropriate. The same applies to docstrings.

tests/test_testing.py
```diff
@@ -47,3 +63,21 @@ def test_pytest_plugin_fail(pytester, book_list_html_response) -> None:
...
     result = pytester.runpytest()
     result.assert_outcomes(failed=1)
+
+
+class DateItemPage(WebPage):
+    async def to_item(self) -> dict:  # noqa: D102
+        return {
+            "foo": "bar",
+            "scraped_on": datetime.datetime.now().isoformat(),
+        }
```
kmike, line 73: What do you think about using an actual item from zyte-common-items, to ensure we don't break it?

web_poet/testing/__init__.py
```diff
@@ -0,0 +1 @@
+from .utils import INPUT_DIR_NAME, META_FILE_NAME, OUTPUT_FILE_NAME, save_fixture
```
kmike, line 1: Why would users need to access these constants?

I was thinking that the programmatic API we expose for non-pytest users could be the following:

1. save_fixture, to save a fixture
2. load_fixture, to load a fixture :)
3. a function extracted from WebPoetItem.runtest, which actually runs the test

I'm not sure, (2) might not be necessary.

What do you think?

wRAR: I think I thought that `save_fixture` is a convenience function but you could want to write or manipulate the fixture directly. But maybe it's not really needed.

web_poet/testing/pytest.py
```diff
@@ -0,0 +1,119 @@
...
+                    name=entry.name,
+                    path=entry,
+                )
+                if item.input_path.is_dir() and item.output_path.is_file():
+                    result.append(item)
+        return self.sorted(result)
+
+
+class WebPoetItem(pytest.Item):
+    """Represents a directory containing one test case."""
```
kmike, line 40: It seems we use 3 different terms for the same concept:

* Item (I guess it comes from pytest)
* test case (not sure a name is good; in unittest TestCase is a collection of tests
* fixture (save_fixture, etc.)

None of this fits the definitions in https://docs.python.org/3/library/unittest.html. I think pytest uses "test" term for this.

Not sure what's the best way, but it'd be good to clean it up.

web_poet/testing/__init__.py
kmike: If we allow to manipulate fixtures directly, we have less flexibility in changing its structure; it seems to be better not to expose this option. Hm, though it goes against the idea of creating fixtures without any code, just in an editor.

wRAR: File names also aren't enough to know the structure, as there are also folder hierarchies that we could change.

kmike: Let's try to close it off for now, and discourage "manual" interventions. We can expose it later. In the meantime, we'll have time to iterate on folder and file structure if needed.

tests/test_testing.py
wRAR: zyte-common-items depends on web-poet, though it looks like it's not a problem for running tox

web_poet/testing/utils.py
Gallaecio: A `json.dumps` of `ItemAdapter(page_object.to_item()).asdict()`, to be exact. Line 40 below.

I wonder if the web-poet standard should require the output of `to_item()` to be an itemadapter-supported item. We could make that a test-only requirement, but for the sake of standardization I would go for a general requirement.

Gallaecio (commented):
Looks good to me.

I do want to have another pass at the documentation, I might have a few suggestions.

I take it additional requests are out of scope here, to be implemented later (I think we agreed to this elsewhere).

web_poet/testing/utils.py
kmike: +1, I think it makes sense to document it. 

It is not a limitation on the item type returned, because one can register custom adapters.

web_poet/serialization/api.py
```diff
@@ -110,42 +112,59 @@ def deserialize_leaf(cls: Type[T], data: SerializedLeafData) -> T:
     return f_ser.f_deserialize(cls, data)  # type: ignore[attr-defined]
 
 
-def _get_fqname(cls: type) -> str:
-    """Return the fully qualified name for a type.
+def _get_name_for_class(cls: type) -> str:
```
kmike, line 115: The coverage warnings here are weird. This line should be executed on import time, it can't be the case that the module is never imported.

web_poet/testing/pytest.py
```diff
@@ -0,0 +1,120 @@
+import asyncio
+import json
+import operator
+from pathlib import Path
+from typing import Any, Iterable, List, Optional, Set, Union
```
kmike, line 5: coverage warnings are annoying, they basically tell that the module is never imported, which doesn't make sense

web_poet/testing/pytest.py
```diff
@@ -0,0 +1,120 @@
...
+
+_found_po_dirs: Set[Path] = set()
+
+
+def pytest_collect_file(
+    file_path: Path, path: Any, parent: pytest.Collector
+) -> Optional[pytest.Collector]:
+    if file_path.name == OUTPUT_FILE_NAME:
+        testcase_dir = file_path.parent
+        if not (testcase_dir / INPUT_DIR_NAME).is_dir():
```
kmike, line 108: It seems this logic is repeated: testcase_dir / INPUT_DIR_NAME code is in WebPoetItem, and .is_dir() check is in WebPoetFile.collect. Would it be possible to have a single place which defines what the fixture is, and what is a valid fixture?

web_poet/testing/pytest.py
```diff
@@ -0,0 +1,120 @@
...
+            return None
+        po_dir = testcase_dir.parent
```
kmike, line 110: Let's try to rename all variables named "po_..." :)

web_poet/serialization/api.py
wRAR: After some change (I would say it's tox 4 but I don't think it would matter) the coverage output indeed changed from "all is covered except several lines in 3 files" to this, where all defs and imports etc. are not covered.

web_poet/testing/utils.py
```diff
@@ -0,0 +1,44 @@
...
+from pathlib import Path
+from typing import Any, Iterable, Optional, Union
+
+from itemadapter import ItemAdapter
+
+from web_poet.serialization import SerializedDataFileStorage, serialize
+
+INPUT_DIR_NAME = "inputs"
+OUTPUT_FILE_NAME = "output.json"
+META_FILE_NAME = "meta.json"
```
kmike, line 12: nitpick: it would probably make sense to move these constants closer to WebPoetItem

web_poet/testing/utils.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def _get_available_filename(template: str, directory: Union[str, os.PathLike]) -> str:
+    i = 1
+    while True:
+        result = Path(directory, template.format(i))
+        if not result.exists():
```
kmike, line 19: This might become slow-ish with large folders, but we can optimize it later. Probably, some solution with iterdir() could be better. Probably we don't need to worry about this :)

web_poet/testing/utils.py
```diff
@@ -0,0 +1,44 @@
...
+            return result.name
+        i += 1
```
kmike, line 21: I slighly prefer to continue from the last number, without trying to fill the gaps, but either way is fine.

web_poet/testing/utils.py
```diff
@@ -0,0 +1,44 @@
...
+    base_directory: Union[str, os.PathLike],
+    inputs: Iterable[Any],
+    item: Any,
+    meta: Optional[dict] = None,
+    fixture_name=None,
+) -> Path:
+    if not fixture_name:
+        fixture_name = _get_available_filename("test-{}", base_directory)  # noqa: P103
+    fixture_dir = Path(base_directory, fixture_name)
+    inputs_dir = Path(fixture_dir, INPUT_DIR_NAME)
```
kmike, line 34: A nitpick: I wonder if we can centralize this logic of the folder structure. For example, create another class similar to WebPoetItem to contain all this logic, and make WebPoetItem use it.

web_poet/testing/pytest.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def get_expected_output(self) -> dict:
+        """Return the saved output."""
+        return json.loads(self.output_path.read_bytes())
+
+    def runtest(self) -> None:  # noqa: D102
+        meta = self.get_meta()
+        frozen_time: Optional[str] = meta.get("frozen_time")
+        if frozen_time:
+            with freeze_time(frozen_time):
```
kmike, line 92: Do you know how are timezones handled in freezegun? 

For example, on a machine with UTC+1 timezone frozen_time was recorded and stored to meta. Then, to run the test we load the time on another machine, which is in UTC+2 timezone. The code in the page object uses UTC time, or unix timestamps. Are these timestamps going to be different on those 2 machines?

docs/page-objects/testing.rst
```diff
@@ -0,0 +1,96 @@
...
+        │   │   ├── HttpResponse-other.json
+        │   │   └── ResponseUrl.txt
+        │   ├── meta.json
+        │   └── output.json
+        └─── test-2
+            ├── inputs
+            │   ├── HttpResponse-body.html
+            │   ├── HttpResponse-other.json
+            │   └── ResponseUrl.txt
+            ├── meta.json
```
kmike, line 67: @wRAR as you noted elsewhere, it'd be good to document the time freezing option in meta.json.

kmike (approved):
Looks good overall, thanks @wRAR! We might need to do some light clean-up on docs later, but I think that's not a requirement to merge it.

web_poet/serialization/api.py
kmike: Interesting! If that's tox issue, probably it has something to do with their new default to use isolated builds.

web_poet/testing/pytest.py
wRAR: The string is paresed into a naive datetime object and then you can pass the timezone explicitly via https://github.com/spulec/freezegun#timezones

Without additional thinking (and recalling how do naive objects work) I'm not sure what this means for us, I'll check it.

BurnzZ (approved):
+1 to merge as well! I'm keen on seeing the succeeding PRs when we try to handle additional requests. 😄 We can refine the existing docstrings and tutorials on that PR.

web_poet/testing/pytest.py
```diff
@@ -0,0 +1,120 @@
...
+            output = self.get_output()
+        expected_output = self.get_expected_output()
+        assert output == expected_output
+
+
+_found_po_dirs: Set[Path] = set()
+
+
+def pytest_collect_file(
+    file_path: Path, path: Any, parent: pytest.Collector
```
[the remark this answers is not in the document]
wRAR: Yeah, the `path` argument is deprecated, not sure if I should support passing it. I guess only pytest itself calls this method and with the 7.0+ version it will pass the `file_path` argument, but we probably need to support older pytest too? Not sure about this.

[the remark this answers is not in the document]
wRAR: According to https://github.com/pytest-dev/pytest/discussions/10261 it's not enough to handle both arguments, old pytest requires the old signature.

docs/page-objects/testing.rst
```diff
@@ -72,6 +72,10 @@ Page Object directory from an iterable of dependencies, an output item and an
 optional metadata dictionary. It can optionally take a name for the fixture
 directory. By default it uses incrementing names "test-1", "test-2" etc.
 
+.. note::
+    ``output.json`` contains a result of
+    ``json.dumps(ItemAdapter(page_object.to_item()).asdict()``.
```
kmike, line 77:
```suggestion
    ``json.dumps(ItemAdapter(page_object.to_item()).asdict())``.
```

This is also not technically true, as we use some JSON options like ensure_ascii=True. What do you think about replacing it with something like

```
``ItemAdapter(page_object.to_item()).asdict()`` saved as JSON
```

wRAR:
I've added a Fixture class which represents a directory structure, so it's not useful for e.g. `save_fixture`, not sure if we need something else.

OTOH it can be used instead of a hypothetical `load_fixture` if you pass two arguments to it. Actually we only need just one argument, I'll update it.

docs/page-objects/testing.rst
```diff
@@ -87,10 +91,20 @@ provided by it to generate fixtures in a convenient way.
...
 The provided ``pytest`` plugin is automatically registered when ``web-poet`` is
 installed, and running ``pytest`` in a directory containing fixtures will
 discover them and run tests for them.
+
+.. _web-poet-testing-frozen_time:
+
+Handling time fields
+====================
+
+Page Objects can return the scraping date or even time in the result and it
```
kmike, line 106: There might be other cases when time is important, e.g. page object might use a current timestamp to build some url. What do you think about something along these lines? 

> Sometimes output of a page object might depend on the current time. For example, the item may contain the scraping datetime, or a current timestamp may be used to build some URLs. ...

docs/page-objects/testing.rst
```diff
@@ -87,10 +91,20 @@ provided by it to generate fixtures in a convenient way.
...
+would break the test when it runs later. To avoid this the metadata dictionary
+can contain a ``frozen_time`` field set to the time value used when generating
```
kmike, line 108: What is the format of this field? Is it an ISO8061 string, or something else?

Are there any gotchas related to time zones users need to be aware of?

wRAR: `freezegun` uses `dateutil` so it seems that it can be any reasonable format. I haven't checked the timezone stuff yet.

web_poet/testing/utils.py
```diff
@@ -21,6 +30,69 @@ def _get_available_filename(template: str, directory: Union[str, os.PathLike]) -
...
+        if frozen_time:
+            with freeze_time(frozen_time):
+                output = self.get_output()
+        else:
+            output = self.get_output()
+        expected_output = self.get_expected_output()
+        assert output == expected_output
+
+
 def save_fixture(
```
kmike, line 100: An idea (not 100% sure): make this a classmethod of Fixture. This way you might be able to reuse some of the methods like output_path. It can also return a Fixture instance.

setup.py
wRAR, line 20: I've found that the coverage report gets broken by this line, not sure why yet.

wRAR: https://stackoverflow.com/questions/62221654/how-to-get-coverage-reporting-when-testing-a-pytest-plugin

https://pytest-cov.readthedocs.io/en/latest/plugins.html

wRAR: I've used the SO idea of running `coverage` directly because the idea from the second link is both complicated and leads to several additional problems, some of which have open issues in pytest-cov. The benefits of pytest-cov are listed at https://pytest-cov.readthedocs.io/en/latest/readme.html , hopefully we can live withouth them.

web_poet/testing/pytest.py
wRAR: The datetime string, whatever it is, is converted to UTC if the timezone is present, or treated as UTC if it's not.

The code that runs under `freeze_time()` thinks that the current timezone is UTC, unless you pass a different timezone to `freeze_time()`, hopefully this shouldn't be a problem.

So when we generate the value we can either convert it to UTC or include the timezone, and for users who do that themselves we should say to do the same.

docs/page-objects/testing.rst
wRAR: (any reasonable format except a UNIX timestamp, apparently)

docs/page-objects/testing.rst
```diff
@@ -0,0 +1,116 @@
...
+.. _web-poet-testing-scrapy-poet:
+
+scrapy-poet integration
+=======================
+
+Projects that use the `scrapy-poet`_ library can use the `Scrapy command`_
+provided by it to generate fixtures in a convenient way.
+
+.. _scrapy-poet: https://github.com/scrapinghub/scrapy-poet
+.. _Scrapy command: <insert doc link>
```
kmike, line 92: This might prevent merging; I think it could make sense to merge this PR first, before scrapy-poet one. What do you think?

wRAR: Indeed. Should we omit this section completely for now?

web_poet/testing/pytest.py
kmike: Wouldn't assuming UTC by default be problematic?

So, let's say there is a Page Object:

```py
class MyPage(WebPage):    
    @field 
    def scraped_timestamp(self):
        # Unix timestamp. It doesn't depend on the TZ of the machine.
        return time.time()  

    @field 
    def scraped_datetime(self):
        # This one depends on the timezone.
        return str(datetime.datetime.now())
```

User runs generation code locally, and saves frozen_time without a timezone. MyPage assumes computer timezone when generating the item. But when freezegun is used, UTC is used instead. So, scraped_timestamp stays the same, but scraped_datetime returns a different value, and the comparison fails.

What do you think about requiring to set the timezone in the frozen_time, and not assuming UTC? Any other options?

I'm not sure converting frozen_time time to UTC before saving works. It seems that if we don't store timezone, and use either local TZ or UTC, then we can't guarantee scraped_datetime output would be the same.

docs/page-objects/testing.rst
kmike: Yes, that's an option.

web_poet/testing/utils.py
```diff
@@ -0,0 +1,128 @@
+import asyncio
```
kmike, line 1: What do you think about renaming this module to web_poet.testing.fixture?

web_poet/testing/pytest.py
```diff
@@ -0,0 +1,132 @@
...
+            session: Optional[pytest.Session] = None,
+            nodeid: Optional[str] = None,
+            **kw,
+        ) -> None:
+            super().__init__(name, parent, config, session, nodeid, **kw)
+            self.fixture = Fixture(self.path)
+
+    else:
+
+        def __init__(  # type: ignore[misc]
```
kmike, line 56: Do you know why is it marked as not covered, while there is an env with older pytest?

wRAR: Oh, because it runs in "check", not in "test" and so coverage for it is not uploaded. Not sure how to rewrite this in the best way, as the "test" steps run `tox -e py`.

web_poet/testing/pytest.py
wRAR: Yes, if the result depends on the timezone then we need to store both time and timezone. It looks like it will need to be a separate field, as `freeze_time` doesn't use the TZ specification from the time string to set a fake TZ and takes a separate argument for that. Alternatively we may be able to parse the string to extract the timezone in web-poet before passing it to `freeze_time`.

wRAR: I've tried to implement this but there are multiple open issues in freezegun related to its handling of timezones (I think I'm running into https://github.com/spulec/freezegun/issues/348 with my current code) and I worry that even if we manage to write code that works for our tests it can break on some user PO code. Also if the user generates the fixture manually and not with the scrapy-poet code (which can call freezegun in some complicated way if needed) we may need to give some instructions to them.

There is a newer alternative to freezegun, https://github.com/adamchainz/time-machine, I haven't tried it but it's a binary module. There may also be some workaround by e.g. requiring frozen_time to be in UTC (it shouldn't break user code, just reduce the cases that can be tested).

kmike: So, our options:

1. Require that tests must be generated and run on machines with the same TZ. It looks pretty bad.
2. Work-around freezegun issues, or help fixing them. 
3. Handle timezones ourselves, e.g. using time.tzset(). Could be a simple solution. But could be not, we might start fighting against edge cases, increasing maintenance burden. Also, it doesn't work on Windows.
4. Try to use time-machine. Binary doesn't bother me too much. Cons: binary, doesn't work in Pypy.

> There may also be some workaround by e.g. requiring frozen_time to be in UTC (it shouldn't break user code, just reduce the cases that can be tested).

How would it work, could you please elaborate?

I also wonder if there is some other workaround possible, e.g. something about freezing the time before generating the output.

My votes:

1. If there is some creative work-around, let's do it. I think that's totally fine to be really particular about the format of frozen_time if this helps. For example, asking users to convert their local time to UTC is fine, as soon as scrapy-poet does the right thing. We may even go as far as not documenting the format of the time, and just saying it should be an output of some function which we provide.
2. Try time-machine. It quotes some other benefits, e.g. it says that freeze-gun can be super-slow in large projects, and we are going to have large projects.

wRAR: > I think that's totally fine to be really particular about the format of frozen_time if this helps. For example, asking users to convert their local time to UTC is fine, as soon as scrapy-poet does the right thing. 

The format is not the problem: as long as it's UTC or has the timezone information in it, the current code should return correct datetime values, they will just be in UTC, not in some other timezone. It's making the code run as if the timezone is different and still return correct values is problematic.

wRAR: Additional findings:

1. Both freezegun and time-machine freeze the fixture generation command if the asyncio reactor is used and time ticking is not enabled. It's enabled by default in time-machine and the doc says this: "code tends to make the (reasonable) assumption that time progresses whilst running, and should normally be tested as such. Testing with time frozen can make it easy to write complete assertions, but it's quite artificial. Write assertions against time ranges, rather than against exact values." This is a valid consideration but we currently compare exact values.
2. time-machine mocks the timezone using `time.tzset()` which doesn't work on Windows. It also requires `backports.zoneinfo` for this on Python < 3.8, but that shouldn't be a problem. This means that some tests will fail on Windows (ones for code that outputs data dependent on the current timezone) and we won't be able to ignore the affected fields unless we make some system that identifies such fields based on metadata or some logic. We can only recommend generating such tests using UTC.
3. I haven't finished the implementation but it looks possible to write all kinds of timezone-related tests with time-machine as opposed to freezegun and it won't require working around issues. It will require creating zoneinfo.ZoneInfo objects though so it will slightly complicate the code.
4. Technically, we don't need to use the same implementation in test generation and test running, but as scrapy-poet depends on web-poet it makes sense to use the same one.

kmike: Do you think we can find some workaround for Windows users, even if it's ugly? Like asking hem to set timezone manually? :) Also, are date fields in zyte-common-items timezone-dependent?

wRAR: > Do you think we can find some workaround for Windows users, even if it's ugly? Like asking hem to set timezone manually?

Setting the timezone manually should work, no idea about lighter workarounds though.

> are date fields in zyte-common-items timezone-dependent?

`Metadata.dateDownloaded` is in UTC.

tests/test_testing.py
```diff
@@ -0,0 +1,135 @@
...
+    if date.tzinfo is None:
+        # convert to the aware object so that date_local_str always includes the offset
+        date = date.astimezone()
+    date_str = date.astimezone(dateutil.tz.UTC).strftime("%Y-%M-%dT%H:%M:%SZ")
+    date_local_str = date.strftime("%Y-%M-%dT%H:%M:%S%z")
+    return Product(
+        url="http://example.com",
+        name="foo",
+        metadata=MetadataLocalTime(
+            dateDownloaded=date_str, dateDownloadedLocal=date_local_str  # type: ignore[call-arg]
```
wRAR, line 89: mypy bug, probably https://github.com/python/mypy/pull/6107 (we use old mypy)

docs/page-objects/testing.rst
```diff
@@ -0,0 +1,105 @@
...
+
+Sometimes output of a page object might depend on the current time. For
+example, the item may contain the scraping datetime, or a current timestamp may
+be used to build some URLs. When a test runs at a different time it will break.
+To avoid this the metadata dictionary can contain a ``frozen_time`` field set
+to the time value used when generating the test. This will instruct the test
+runner to use the same time value so that field comparisons are still correct.
+
+The value can be any string understood by `dateutil`_. If it doesn't include
+timezone information, UTC will be assumed.
```
kmike, line 103: I think we might need to add a bit more information here. In particular:

* Which time format should users use, and which options should they avoid
* What should Windows users do

web_poet/testing/fixture.py
```diff
@@ -0,0 +1,169 @@
...
+        return ItemAdapter(item).asdict()
+
+    def get_expected_output(self) -> dict:
+        """Return the saved output."""
+        return json.loads(self.output_path.read_bytes())
+
+    @staticmethod
+    def _parse_frozen_time(meta_value: str) -> datetime.datetime:
+        """Parse and possibly fix the frozen_time metadata string."""
+        parsed_value = dateutil.parser.parse(meta_value)
```
kmike, line 101: Is dateutil only used to allow flexible time formats? If so, can we drop this dependency, and make time formats more restricted?

wRAR: time-machine depends on it anyway.

kmike (approved):
+1 to merge after expanding docs about date and timezone handling.

PR #112, opened 2023-01-06
WIP: move some registry functionalities from scrapy-poet

BurnzZ:
Taken from https://github.com/scrapinghub/scrapy-poet/pull/103.

# TODO: 

- [ ] Fix flaky tests
- [ ] Optimize
- [ ] Docs
- [ ] Changelog

### DIFF
--- a/web_poet/rules.py
+++ b/web_poet/rules.py
@@ -4,14 +4,15 @@ import importlib
 import importlib.util
 import pkgutil
 import warnings
-from collections import deque
+from collections import defaultdict, deque
 from operator import attrgetter
-from typing import Any, Dict, Iterable, List, Optional, Type, TypeVar, Union
+from typing import Any, Dict, Iterable, List, Mapping, Optional, Type, TypeVar, Union
 
 import attrs
-from url_matcher import Patterns
+from url_matcher import Patterns, URLMatcher
 
 from web_poet._typing import get_item_cls
+from web_poet.page_inputs.url import _Url
 from web_poet.pages import ItemPage
 from web_poet.utils import _create_deprecated_class, as_list, str_to_pattern
 
@@ -115,9 +116,52 @@ class RulesRegistry:
     """
 
     def __init__(self, *, rules: Optional[Iterable[ApplyRule]] = None):
-        self._rules: List[ApplyRule] = []
+        self._rules: Dict[int, ApplyRule] = {}
+        self.overrides_matcher: Dict[Type[ItemPage], URLMatcher] = defaultdict(
+            URLMatcher
+        )
+        self.item_matcher: Dict[Optional[Type], URLMatcher] = defaultdict(URLMatcher)
+
         if rules is not None:
-            self._rules = list(rules)
+            for rule in rules:
+                self.add_rule(rule)
+
+    def add_rule(self, rule: ApplyRule) -> None:
+        """Registers an :class:`web_poet.rules.ApplyRule` instance."""
+        rule_id = hash(rule)
+
+        # A common case when a page object subclasses another one with the same
+        # URL pattern.
+        matched = self.item_matcher.get(rule.to_return)
+        if matched:
+            pattern_dupes = [
+                pattern
+                for pattern in matched.patterns.values()
+                if pattern == rule.for_patterns
+            ]
+            if pattern_dupes:
+                rules = [
+                    r
+                    for p in pattern_dupes
+                    for r in self.search(for_patterns=p, to_return=rule.to_return)
+                ]
+                warnings.warn(
+                    f"Similar URL patterns {pattern_dupes} were declared earlier "
+                    f"that use to_return={rule.to_return}. The first, highest-priority "
+                    f"rule added to SCRAPY_POET_REGISTRY will be used when matching "
+                    f"against URLs. Consider updating the priority of these rules: "
+                    f"{rules}."
+                )
+
+        if rule.instead_of:
+            self.overrides_matcher[rule.instead_of].add_or_update(
+                rule_id, rule.for_patterns
+            )
+        if rule.to_return:
+            self.item_matcher[rule.to_return].add_or_update(rule_id, rule.for_patterns)
+
+        # TODO: test removing the rule
+        self._rules[rule_id] = rule
 
     @classmethod
     def from_override_rules(
@@ -199,7 +243,7 @@ class RulesRegistry:
                 to_return=to_return or get_item_cls(cls),
                 meta=kwargs,
             )
-            self._rules.append(rule)
+            self.add_rule(rule)
             return cls
 
         return wrapper
@@ -214,7 +258,7 @@ class RulesRegistry:
             beforehand to recursively import all submodules which contains the
             ``@handle_urls`` decorators from external Page Objects.
         """
-        return self._rules[:]
+        return list(self._rules.values())
 
     def get_overrides(self) -> List[ApplyRule]:
         """Deprecated, use :meth:`~.RulesRegistry.get_rules` instead."""
@@ -258,6 +302,32 @@ class RulesRegistry:
         msg = "The 'search_overrides' method is deprecated. Use 'search' instead."
         warnings.warn(msg, DeprecationWarning, stacklevel=2)
         return self.search(**kwargs)
+
+    def _rules_for_url(
+        self, url: Union[_Url, str], url_matcher: Dict[Any, URLMatcher]
+    ) -> Mapping[Type, Type[ItemPage]]:
+        result: Dict[Type, Type[ItemPage]] = {}
+
+        for target, matcher in url_matcher.items():
+            rule_id = matcher.match(url)
+            if rule_id is not None:
+                result[target] = self._rules[rule_id].use
+
+        return result
+
+    def overrides_for(self, url: Union[_Url, str]) -> Mapping[Type, Type[ItemPage]]:
+        """Finds all of the page objects associated with the given URL and
+        returns a Mapping where the 'key' represents the page object that is
+        **overridden** by the page object in 'value'.
+        """
+        return self._rules_for_url(url, self.overrides_matcher)
+
+    def page_object_for(self, url: Union[_Url, str]) -> Mapping[Type, Type[ItemPage]]:
+        """Finds all of the page objects associated with the given URL and
+        returns a Mapping where the 'key' represents the item class that is
+        **produced** by the page object in 'value'.
+        """
+        return self._rules_for_url(url, self.item_matcher)
 
 
 def _walk_module(module: str) -> Iterable:
--- a/tests/test_rules.py
+++ b/tests/test_rules.py
@@ -1,3 +1,5 @@
+import warnings
+
 import attrs
 import pytest
 from url_matcher import Patterns
@@ -9,8 +11,11 @@ from tests.po_lib import (
     POTopLevelOverriden2,
 )
 from tests.po_lib.a_module import POModule, POModuleOverriden
-from tests.po_lib.nested_package import PONestedPkg
-from tests.po_lib.nested_package.a_nested_module import PONestedModule
+from tests.po_lib.nested_package import PONestedPkg, PONestedPkgOverriden
+from tests.po_lib.nested_package.a_nested_module import (
+    PONestedModule,
+    PONestedModuleOverriden,
+)
 from tests.po_lib_sub import POLibSub
 from tests.po_lib_to_return import (
     CustomProductPage,
@@ -20,7 +25,10 @@ from tests.po_lib_to_return import (
     LessProductPage,
     MoreProductPage,
     Product,
+    ProductFewerFields,
+    ProductMoreFields,
     ProductPage,
+    ProductSeparate,
     ProductSimilar,
     SeparateProductPage,
     SimilarProductPage,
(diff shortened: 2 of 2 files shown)


web_poet/rules.py
```diff
@@ -259,6 +317,34 @@ def search_overrides(self, **kwargs) -> List[ApplyRule]:
...
+        url = str(url)
+
+        for target, matcher in matchers.items():
+            rule_id = matcher.match(url)
+            if rule_id is not None:
+                result[target] = self._rules[rule_id].use
+
+        return result
+
+    def overrides_for(self, url: Union[_Url, str]) -> Mapping[Type, Type[ItemPage]]:
```
kmike, line 334: If that's a page object=>page object mapping, can we use `Mapping[Type[ItemPage], Type[ItemPage]]`?

web_poet/rules.py
```diff
@@ -259,6 +317,34 @@ def search_overrides(self, **kwargs) -> List[ApplyRule]:
...
+        """Finds all of the page objects associated with the given URL and
+        returns a Mapping where the 'key' represents the page object that is
+        **overridden** by the page object in 'value'.
+        """
+        return self._rules_for_url(url, self._overrides_matchers)
+
+    def page_object_for(self, url: Union[_Url, str]) -> Mapping[Type, Type[ItemPage]]:
+        """Finds all of the page objects associated with the given URL and
+        returns a Mapping where the 'key' represents the item class that is
+        **produced** by the page object in 'value'.
```
kmike, line 344: What's the use case for returning all item classes? If I'm not mistaken, in scrapy-poet the most common case is getting 1 item type. Currently if you need to know which page object handles MyProduct data type for example.com, we'll run matching for example.com for all registered item types, not just for MyProduct.

Basically, _rules_for_url iterates over all targets and runs matching for all of them, but it looks pointless, because we know the target already (the item class); it's a key in the matchers dict, so we can look it up easily.

BurnzZ: Apart from being symmetric with `overrides_for()`, it offers discoverability on the types of items that can be extracted from a given URL. This leads to several possible functionalities. For example, in **scrapy-poet**, there could be a wildcard which indicates "give me all the items you can extract from this URL based on its pattern". This is useful in scenarios where product discovery leads to various items with different URL patterns. 

Should we drop the current `page_object_for()` for now?

However, you're right about the inefficiencies for `page_object_for()` and I've created a new more efficient `page_object_for_item()` for it.

web_poet/rules.py
```diff
@@ -115,9 +116,66 @@ class ExampleComProductPage(WebPage[Product]):
...
+        # breaker woulud be the "Rule identifier", this means we can base it on
+        # when a given rule was added to the registry, i.e. a counter.
+        self._rule_counter = 0
+
         if rules is not None:
-            self._rules = list(rules)
+            for rule in rules:
+                self.add_rule(rule)
+
+    def add_rule(self, rule: ApplyRule) -> None:
```
BurnzZ, line 140: There's a potential for returning the `rule_id: int` but this also means that a new `def get_rule(self, rule_id: int)` needs to be defined as well.

Although I don't see an immediate use case to expose this for now apart from potentially storing the `rule_id` somewhere to have a quick registry look-up later on.

Let me know if we should do expose this as early as now.

web_poet/rules.py
kmike: hey! This makes sense; there could be some use cases which require getting everything. But I'd probably drop the method for now, until we face them.

BurnzZ: Done in https://github.com/scrapinghub/web-poet/pull/112/commits/a5e8480f9d20bf65f6ecc8b0d6e56e666abb1223

web_poet/rules.py
BurnzZ: Done in https://github.com/scrapinghub/web-poet/pull/112/commits/c53249cfb8372dd520e26cb5b40f72dfa2aa0080

web_poet/rules.py
```diff
@@ -115,9 +116,66 @@ class ExampleComProductPage(WebPage[Product]):
...
+        if rule.instead_of:
+            self._overrides_matchers[rule.instead_of].add_or_update(
+                rule_id, rule.for_patterns
+            )
+        if rule.to_return:
+            self._item_matchers[rule.to_return].add_or_update(
+                rule_id, rule.for_patterns
+            )
+
+        # TODO: test removing the rule
```
Gallaecio, line 177: To be done in this pull request, or later on?

Gallaecio (approved):

web_poet/rules.py
```diff
@@ -259,6 +316,50 @@ def search_overrides(self, **kwargs) -> List[ApplyRule]:
...
+        See example: :ref:`rules-page_object_for_item-example`.
+        """
+
+        url = str(url)
+
+        matcher = self._item_matchers.get(item_cls)
+        if matcher:
+            rule_id = matcher.match(url)
+            if rule_id is not None:
+                return self._rules[rule_id].use
```
kmike, line 359: It'd be nice to clean up a code a bit: 

* some code is duplicated between page_object_for_item and _rules_for_url
* page_object_for_item and overrides_for are not "symmetric" in implementation: inner code is extracted to a method in overrides_for, but inlined in page_object_for_item.

BurnzZ: Refactored. https://github.com/scrapinghub/web-poet/pull/112/commits/299f1652521a8e8da956133250cc2501dc552918

web_poet/rules.py
```diff
@@ -236,10 +295,32 @@ def search(self, **kwargs) -> List[ApplyRule]:
...
+        if matcher:
+            if rule_ids:
+                # If both params are used then narrow down the rules.
+                rule_ids.union(matcher.patterns.keys())
+            else:
+                rule_ids.update(matcher.patterns.keys())
+
+        rules = [self._rules[id_] for id_ in rule_ids]
+
+        if len(kwargs) == 1 and rules:
```
kmike, line 316: It seems len(kwargs) can be 2, and we still might be able to return the result (kwargs.keys() is {to_return, instead_of}).

What do you think about using something like `kwargs.keys() <= {"to_return", "instead_of"}` ?

BurnzZ: Good call. I've refactored it to support this. More importantly, we can also now have quicker look-ups when `None` is passed when searching for either `instead_of` or `to_return`.

web_poet/rules.py
```diff
@@ -115,9 +116,63 @@ class ExampleComProductPage(WebPage[Product]):
     """
 
     def __init__(self, *, rules: Optional[Iterable[ApplyRule]] = None):
-        self._rules: List[ApplyRule] = []
+        self._rules: Dict[int, ApplyRule] = {}
+        self._overrides_matchers: Dict[
+            Optional[Type[ItemPage]], URLMatcher
+        ] = defaultdict(URLMatcher)
+        self._item_matchers: Dict[Optional[Type], URLMatcher] = defaultdict(URLMatcher)
```
kmike, line 123: Why is Optional used here, are there cases when keys are going to be None?
Also, if it's a defaultdict, should't we use defaultdict as a type, not Dict?

web_poet/rules.py
```diff
@@ -115,9 +116,63 @@ class ExampleComProductPage(WebPage[Product]):
...
         if rules is not None:
-            self._rules = list(rules)
+            for rule in rules:
+                self.add_rule(rule)
+
+    def add_rule(self, rule: ApplyRule) -> None:
+        """Registers an :class:`web_poet.rules.ApplyRule` instance."""
+
+        # Ignore the type since ``ApplyRule`` could have an empty ``.to_return``
+        matched = self._item_matchers.get(rule.to_return)  # type: ignore[arg-type]
```
kmike, line 144: It can be also written like `if rule.to_return is not None`, avoiding the type issue. It's the same amout of code, because you'd be able to drop the comment :) 

When we drop Python 3.7 support, we should be able to use walrus operator to simplify the code.

web_poet/rules.py
```diff
@@ -259,6 +339,48 @@ def search_overrides(self, **kwargs) -> List[ApplyRule]:
...
+        See example: :ref:`rules-overrides_for-example`.
+        """
+        result: Dict[Type[ItemPage], Type[ItemPage]] = {}
+        for target, matcher in self._overrides_matchers.items():
+            if target is None:
+                continue
+            po = self._match_url_for_page_object(url, matcher)
+            if po:
+                result[target] = po
+        return result
```
kmike, line 369: What do you think about something like this?

```suggestion
        result: Dict[Type[ItemPage], Type[ItemPage]] = {}
        for replaced_page, matcher in self._overrides_matchers.items():
            if replaced_page is None:
                continue
            page = self._match_url_for_page_object(url, matcher)
            if page:
                result[replaced_page] = page
        return result
```

web_poet/rules.py
BurnzZ: In https://github.com/scrapinghub/web-poet/pull/112/commits/dc6181024758aa0c1038a70b03a6f529f479c327, it allows faster look-up on cases like `.search(to_return=None)`. Combos well with `instead_of` like `.search(to_return=None, instead_of=None)`.

Updated to use `DefaultDict`.

web_poet/rules.py
BurnzZ: Ahh we could actually remove the `type: ignore` here since `self._item_matcher` supports `None` as the key now.

BurnzZ: > Any why do we support None as the key?

It enables faster `.search()` look-up when `to_return` is passed as `None`, like `.search(to_return=None)`. [More examples in the tests](https://github.com/scrapinghub/web-poet/pull/112/files#diff-236257b347f2d255e260081e2980f584367de9fc7999ba32c08e47e4f7edcad5R291-R306).

As opposed to the previous implementation, if we want to do `.search(to_return=None)`, it'd be *O(n)*. But since `self._item_matchers` now supports `None` as the key, it would be *O(1)*.

The same is true of *overrides* and `self._overrides_matchers`.

> Are there cases when the key is None?

An example would be:

```python
@handle_url("example.com", instead_of=SomePage)
@attrs.define
def MyPage(WebPage):
    ...
```

kmike: I'm not sure I follow. self._overrides_matchers and self._item_matchers never contain None as a key, right? So why do we annotate them as defaultdicts which can contain None?

BurnzZ: Both `self._overrides_matchers` and `self._item_matchers` could actually contain `None` as the key.

Let's suppose we have the following as the source of `ApplyRule`s:

```python
@handle_urls("example-a.com")
def NoOverrideNoItem(WebPage): ...

@handle_urls("example-b.com")
def WithItemNoOverride(WebPage[Product]): ...

@handle_urls("example-c.com", instead_of=SomePage)
def NoItemWithOverride(WebPage): ...

@handle_urls("example-d.com", instead_of=SomePage)
def WithItemWithOverride(WebPage[Product]): ...
```

Then we'd have the following state:

```python
self._overrides_matchers = {
    None: URLMatcher(...),      # has 2 rules
    SomePage: URLMatcher(...),  # has 2 rules
}

self._item_matchers = {
    None: URLMatcher(...),      # has 2 rules
    Product: URLMatcher(...),   # has 2 rules
}
```

If we don't support `None` as the key for either of them, then it'd be:

```python
self._overrides_matchers = {
    SomePage: URLMatcher(...),  # has 2 rules
}

self._item_matchers = {
    Product: URLMatcher(...),   # has 2 rules
}
```

However, when we need to do any of the following:
- `.search(to_return=None)`
- `.search(instead_of=None)`
- `.search(to_return=None, instead_of=None)`

It'd be *O(n)* since `self._item_matchers` and `self._overrides_matchers` couldn't match anything and `.search()` would now go through the `List[ApplyRule]` and check the corresponding `None` attributes.

Supporting `None` as the key allows the operations above to be *O(1)* since we could immediately know which `ApplyRule`s could have either `to_return=None` or `instead_of=None`.

BurnzZ: `ApplyRule` also has the following:
- `instead_of: Optional[Type[ItemPage]] = attrs.field(default=None, kw_only=True)`
- `to_return: Optional[Type[Any]] = attrs.field(default=None, kw_only=True)`

kmike: I missed it, thanks for the explanation @BurnzZ!

docs/page-objects/rules.rst
```diff
@@ -390,6 +390,58 @@ in the :class:`~.ApplyRule` to be written into ``web_poet.default_registry``.
...
+
+Suppose you want to retrieve the page object class that is able to create the
+item class that you want from a given webpage, you can use
+:meth:`~.RulesRegistry.page_object_for_item`. For example:
+
+.. code-block:: python
+
+    from web_poet import default_registry
+
+    item_cls = default_registry.page_object_for_item(
```
kmike, line 439: shouldn't it be page_cls?

kmike (approved):
👍

docs/page-objects/rules.rst
BurnzZ: Do you mean renaming it something like `page_object_cls_for_item`? or maybe `page_cls_for_item`?

kmike: The varible is named `item_cls`, but it contains page class, not item class.

BurnzZ: I totally missed that. Thanks! Updated. https://github.com/scrapinghub/web-poet/pull/112/commits/1ab1a43abf2a3dc4c490216ebb6bc301bec6522e

kmike: You actually raised a good point - should we rename this method to page_cls_for_item, as it doesn't return an instantiated object?

BurnzZ: Yeah, I think that's much better. https://github.com/scrapinghub/web-poet/pull/112/commits/e86bf379f8ca1ddd9c6479499a1988e1b47b794a

kmike:
Thanks @BurnzZ!

PR #26, opened 2022-02-25
add core exceptions

BurnzZ:
Built on top of https://github.com/scrapinghub/web-poet/pull/22.

---

- [ ] Docs
- [ ] Changelog 
- [ ] Move existing `RequestBackendError` to this new module

PR #113, opened 2023-01-09
Implement invalid input handling

Gallaecio:
To do:
- [x] Implementation
    - [x] Solution to the async-from-sync issue
    - [x] Caching
    - [x] Support for `@field` method calls
    - [x] Exception subclasses
    - [x] Validation return item use
- [x] Documentation


web_poet/fields.py
```diff
@@ -95,22 +96,39 @@ def __get__(self, instance, owner=None):
...
+                        # TODO: Implement
+                        #
+                        # Running an async function within a sync function
+                        # looks complicated
+                        # https://github.com/django/asgiref/blob/36022636d3390c1c1cbb952336a34d652a6bfb26/asgiref/sync.py#L107-L314
+                        #
+                        # I am not sure if there is a different approach that
+                        # we can take, but I can’t help how much easier things
+                        # would be if we requred the field methods and the
+                        # to_item methods to be async.
```
Gallaecio, line 130: This pull request has basically no implementation yet, because the first step I see would be to determine how to best approach this aspect of the implementation.

The current proposal allows implementing an async `validate_input` method, and when implemented, it is called when accessing a field. However, fields can be both sync and async, and when they are sync, calling the async `validate_input` method from them [seems cumbersome](https://github.com/django/asgiref/blob/36022636d3390c1c1cbb952336a34d652a6bfb26/asgiref/sync.py#L107-L314).

I am considering the following choices at the moment:

- Require `validate_input` to be sync, and only support sync fields from it. While not ideal from a user perspective, it does simplify our implementation, and it may encourage to keep validation computationally cheap.
- Going the [cumbersome approach](https://github.com/django/asgiref/blob/36022636d3390c1c1cbb952336a34d652a6bfb26/asgiref/sync.py#L107-L314), i.e. running the (async) `validate_input` method in a separate thread with its own loop when called from a synchronous field method.
- Making `validate_input` only work when all fields are async, and raising an exception when the method is defined and any field method is not async. This could be problematic in inheritance scenarios, e.g. if you inherit from a class that defines a sync field method, you need to redefine it in your subclass.
- Going full async, i.e. stop allowing sync versions of field and to_item methods altogether as part of the web-poet specification. This would be *very* backward-incompatible, but looking at the existing code to deal with sync/async mixes, and to the “cumbersome approach”, the idea is growing on me.

kmike: I think only supporting sync validate_input is fine. 

It would disallow some use cases, e.g. validating the input in case the RequestUrl is a dependency, and actual response is retrieved using HttpClient. But in such cases there is another place where the validation can happen - the same method / field where an async request is made.

If we figure out that we do want async validate_input support, and implement one of the solutions, then we'll be able to support the sync version easily, by calling ensure_async.

web_poet/exceptions/core.py
```diff
@@ -6,6 +6,19 @@
 """
 
 
+class InvalidInput(ValueError):
```
kmike, line 9: How should we handle cases where a layout is invalid, and retry won't help, but fallback (e.g. to ML) can? Is it going to be a separate exception? What's the class hierarchy - are they separate, or is one of them a subclass of another?

Gallaecio: I was not planning to make it part of this pull request, but it does make sense to do so.

What about something in these lines? (I did not think too hard about names)

- `BadInput(ValueError)`
  - `WrongInputType`: the type of input data does not seem meant for this page object
    A fallback mechanism could make sense in a crawl targeting additional page objects, although from my experience ignoring such inputs is usually better (e.g. if a product URL points/redirects to a product list URL, you may not want to crawl the products on that product list).
  - `UnknownLayout`: the input seems meant for this page object, but extraction expectations (for key fields?) are failing, probably due to upstream layout changes or an A/B tests.
    A fallback mechanism to a machine-learning alternative would make sense.
  - `NotFound`: the resource is not present, e.g. HTTP 404 or HTTP 410.
    An HTTP 404 or HTTP 410 response may *not* raise this exception if it contains the expected data nonetheless, or if it contains data matching the other `BadInput` subclasses, or if there is reason to believe the response is a ban, meaning `Retry` should be raised instead.
    A fallback mechanism would not make sense.

Even if no fallback of any kind is implemented for some or any of these bad input exceptions, raising different exceptions may be helpful for stats.

In line with this, I wonder if we should implement a `RetriesExceeded` exception, to be thrown by web-poet frameworks after too many failed retries, and whether that exception should be a child of `BadInput` or not.

kmike: See also: https://github.com/scrapinghub/web-poet/pull/26/files

I'm not sure how common is NotFound; what do you think about focussing on the first two now? 

Regarding names, there are at least two ways to approach it. First is to reflect action in the name: if fallback is needed, raise some kind of Fallback.. exception. Second is to be descriptive, and have the fallback logic implicit. 

It seems it'd be good to be explicit about fallbacks, to ensure users don't trigger it by mistake. For example, there might an item which extends standard Product item. If a layout is unknown, it could make sense to trigger a fallback, to fill at least the standard fields. In a second project there might be an item which is completely custom. Fallback doesn't make sense in this case, even if layout is unknown.

Gallaecio: What about `Abort(ValueError)` and `DelegateFallback(ValueError)`?

I wonder if a slightly shorter `UseFallback` or `FallBack` would work instead of `DelegateFallback`. Instead of `Abort` we could use `Fail` or `GiveUp` (I considered `End` and `Stop`, but besides verbs they are also names, so it seems less ideal).

I also wonder if `DelegateFallback` should be a subclass of `Abort`, to make it easier to handle both the same way on cases where a fallback is not supported.

BurnzZ: I'm okay with `UseFallback`. ~On the other hand, I think `AbortPageObject` would be much cleaner regarding what to abort.~ But rethinking about it, we're using `Retry` and not `RetryPageObject`.

In terms of exception hierarchies, should we have a base exception like the following to neatly tie them together?

```
Exception
├── RequestDownloaderVarError
└── ValueError
    └── PageObjectActions
        ├── Retry
        ├── UseFallback
        └── AbortPageObject
```

Gallaecio: Besides making it `PageObjectAction` (singular), :+1:

tests/test_input_validation.py
```diff
@@ -0,0 +1,221 @@
...
+async def test_valid_input_async_field():
+    class Page(BaseValidInputPage):
+        @field
+        async def a(self):
+            return "a"
+
+    assert await Page().a == "a"
+
+
+# Inalid input
```
wRAR, line 66: typo

web_poet/exceptions/core.py
wRAR: Yeah, if the reason to have different exceptions is to handle them in different ways it makes more sense to have "fallback" and "abort" instead of explicit "the layout is unknown" (I assume there are more reasons to fallback than this one).

kmike: Hey! One more thought, to complicate things, as usual :) Let's say we have a page object which extract product information, using zyte-common-items Product schema. We have the following error cases:

1. The page is not a product page, so Product can't be extracted.
2. The page is a product page, and page object doesn't know how to handle the layout. We expect page object to be able to handle all layouts on this website.
3. The page is a product page, and page object doesn't know how to handle the layout. We expect that there could be some layouts that the page object doesn't handle.
4. The page is blank or incorrect, we expect that a retry could help.

It seems this is what we should do in each of the cases:

1. We should return an empty Product, with metadata.probability set to 0. 
2. Not clear. Probably, we should raise a fallback error, and also send this page to developers for investigation.
3. Raise a fallback error.
4. Raise a Retry error.

It seems there is no scenario when page object should be raising an Abort exception, unless I'm missing something. Does it make sense? 

I'm not sure how to support the first case. Some options:

1. Page object should raise Abort exception. Returning an empty item with zero probability is handles elsewhere.
2. We allow to return an item from the validate_input method. If an item is returned from validate_input, then it's cached, and used in to_item and in all the fields.

Gallaecio: My thinking for `Abort` was that it would result in the errback being called.

I see how that would be problematic, though. For example, maybe you want to request both Product and ProductList, and even if one is wrong, you want your callback to receive both, and be able to tell which one was the right one, and for it to have the expected data.

However, I don’t see the point in returning an item from `validate_input`. How would that be different from not raising Abort and returning the item with probability=0 from `to_item`?

Right now I am thinking option 1 would be best, with a scrapy-poet implementation that calls the callback but gives `None` or some sentinel value for the requested item when an `Abort` exception is raised for it.

And maybe we can implement in zyte-common-items a spider middleware to transform that `None`/`sentinel` into an item with the response URL and 0 probability, although I am not sure in what scenarios that would be preferable.

kmike: > However, I don’t see the point in returning an item from validate_input. How would that be different from not raising Abort and returning the item with probability=0 from to_item?

The difference is that in this case we can make fields work properly.

Gallaecio: Work properly meaning raising `Abort`?

kmike:
```py
@attrs.define
    response: HttpResponse

    @field
    def price(self):
        return self.response.css("...")

    @field
    def metadata(self):
        return Metadata(url=self.response.url, probability=1.0)
    
    # ...
    
    def validate_input(self):
        if len(self.response.body) < 100:
            # looks like a download or rendering issue
            raise Retry()  

        if not self.response.css(".product").get():
            # not a product page, a right thing is to return probability=0 without trying to extract data
            return Product(url=self.response.url, metadata=Metadata(probability=0))
            
        if not self.respose.css(".product .supported-layout").get():
            # unsupported layout, use a fallback
            raise Fallback()

# later
page = create_page(MyPage, ...)
item = await page.to_item()

assert item.price == page.price
assert item.metadata.probability == page.metadata.probability
```

Gallaecio: Oh, I see, the fields of the output item of `validate_item` would get written into the corresponding fields, besides also becoming the output of `to_item`. (“used in to_item and in all the fields” as you described in option 2)

It definitely would not be a good idea to have the output of `to_item` be written back into fields, so I think the return-item approach may be the best idea.

I do wonder if it would be cleaner to go for a slightly different API: raising an exception with the item passed as a parameter, rather than returning the item (e.g. `raise Abort(item=…)`, or `raise SetItem(…)`). But I have no strong opinion on it, and they would work basically the same way in practice.

kmike: Returning an item looks ok to me from the API point of view, because of these reasons:

1) it's similar to how scrapy's downloader middlewares work - they can either raise an exception, or return a new request or response
2) it's less code than having to wrap an item in an exception
3) you can have it type checked properly; I'm not sure how to express in mypy "this method should raise an Abort exception which argument must be of an ItemT type"

Gallaecio:
@kmike So, `Retry` and returning and item both work. What do we do about `UseFallback ? Should we, as part of this pull request, make it so that it works, or at least define how it should work? Should the next matching page object be used (i.e. the next one with a lower priority)?

kmike:
> Should the next matching page object be used (i.e. the next one with a lower priority)?

Oh, that's an interesting option, I haven't thought about it. It seems to make sense :)

I don't think it's required to make UseFallback work in this PR. Wouldn't it be a job of library like scrapy-poet anyways? Documenting the expected behavior is the way to go.

Regarding the expected behavior: using priority is one way to do it, but not the only one. I don't think we should specify the concrete algorithm here.

docs/page-objects/input-validation.rst
```diff
@@ -0,0 +1,80 @@
...
+
+-   Raise :exc:`~web_poet.exceptions.Retry`, indicating that the input
+    looks like the result of a temporary issue, and that trying to fetch
+    similar input again may result in valid input.
+
+    See also :ref:`retries-additional-requests`.
+
+-   Raise :exc:`~web_poet.exceptions.UseFallback`, indicating that the
+    page object does not support the input, and that an alternative page object
+    should be tried instead.
```
kmike, line 26: It doesn't have to be a page object though. For example, it may be an API call.

docs/page-objects/retries.rst
```diff
@@ -1,88 +0,0 @@
-.. _retries:
```
kmike, line 1: Have you moved this all to the input-validation section because we recommend raising Retry from validate_input method, preferring it to to_item? That's a nice touch.

docs/page-objects/input-validation.rst
```diff
@@ -0,0 +1,80 @@
...
+    page object for products of Shopify-based websites, ``ShopifyProductPage``,
+    and a more specific page object for commerce.example,
+    ``EcommerceExampleProductPage``. If ``EcommerceExampleProductPage`` cannot
+    parse a product page, but it looks like it might be a valid product page,
+    you would raise :exc:`~web_poet.exceptions.UseFallback` to try to
+    parse the same product page with ``ShopifyProductPage``, in case it works.
+
+    .. note:: web-poet does not yet allow defining which page object to use as
+              fallback, so raising this exception has no effect with the
+              current version of web-poet.
```
kmike, line 39: I think it's better to remove "yet" from this note, or word it in a different way, until we settle on the fallback mechanism.

docs/page-objects/input-validation.rst
```diff
@@ -0,0 +1,80 @@
...
+    page object for products of Shopify-based websites, ``ShopifyProductPage``,
+    and a more specific page object for commerce.example,
+    ``EcommerceExampleProductPage``. If ``EcommerceExampleProductPage`` cannot
+    parse a product page, but it looks like it might be a valid product page,
+    you would raise :exc:`~web_poet.exceptions.UseFallback` to try to
+    parse the same product page with ``ShopifyProductPage``, in case it works.
+
+    .. note:: web-poet does not yet allow defining the alternative parsing
+              implementation to use as fallback, so raising this exception has
+              no effect with the current version of web-poet.
```
kmike, line 39: I think it's more that the frameworks we have (scrapy-poet and web_poet.example) don't have an implementation; I'm not sure it's web-poet's job to handle this exception.

kmike: There is an effect already - the extraction stops, and exception is propagated; it seems this would be the same in future web-poet versions as well.

Gallaecio: https://github.com/scrapinghub/web-poet/pull/113/commits/8dcc57c8b847614d31579787948928e3d12615f2 ?

web_poet/fields.py
```diff
@@ -94,24 +94,27 @@ def __get__(self, instance, owner=None):
             return self.unbound_method(instance)
 
         @staticmethod
-        def _process(value):
-            for processor in out:
-                value = processor(value)
+        def _process(value, page_object):
+            if hasattr(page_object, "validate_input"):
```
kmike, line 98: Is this check needed? It seems it's run twice, as `_validate_input` checks the same.

web_poet/pages.py
```diff
@@ -63,10 +63,27 @@ def __init_subclass__(cls, skip_nonitem_fields: bool = False, **kwargs):
...
+    @cached_method
+    def _validate_input(self) -> None:
+        """Run self.validate_input if defined."""
+        if not hasattr(self, "validate_input"):
+            return
+        if hasattr(self, "_validating_input"):
+            # We are in a recursive call, i.e. _validate_input is being called
+            # from _validate_input itself (likely through a @field method).
+            return
+        self._validating_input = True
```
kmike, line 82: What do you think about making it protected (`__validating_input`)?

web_poet/fields.py
```diff
@@ -94,24 +94,27 @@ def __get__(self, instance, owner=None):
...
+                validation_item = page_object._validate_input()
+                if validation_item is not None:
+                    return getattr(validation_item, method.__name__)
```
kmike, line 101: What do you think about adding a section to fields.rst documentation, which mentions how fields work with validate_item? For example, we can explan the following:

1. Field value may be changed to the one returned by validate_item
2. It is ok to use fields in validate_item, though they must be sync

kmike (approved):
The PR looks good, thanks @Gallaecio!

web_poet/pages.py
Gallaecio: The thing is we need it to be called in subclasses as well, and if we protect it things break in subclasses.

Are you suggesting it so that users do not accidentally use `_validate_input` for their own purposes, or for something else? If the former, maybe we can use some noise on the method, i.e. `_valudate_input_<some random characters>`.

Gallaecio:
@kmike Both https://github.com/scrapinghub/web-poet/pull/113/commits/aaa64ecdc6f59bbc7c63c7a3e65b06fa1a8d8783 and https://github.com/scrapinghub/web-poet/pull/113/commits/91799bae6984e18a0476065f08f22bf5d62b9e73 ended up being more elaborated than what I initially expected, so you might want to have another look.

web_poet/pages.py
kmike: Hey! I was thinking about renaming the flag, not the method.

.flake8
```diff
@@ -44,7 +44,7 @@ per-file-ignores =
     web_poet/serialization/__init__.py:F401,F403
     web_poet/testing/__init__.py:F401,F403
     web_poet/testing/pytest.py:D102
-    tests/po_lib_to_return/__init__.py:D102
+    tests/*:D102
```
kmike, line 47: 👍

docs/page-objects/input-validation.rst
```diff
@@ -0,0 +1,95 @@
...
+only synchronous fields are supported. For example:
+
+.. code-block:: python
+
+   class Page(WebPage[Item]):
+       def validate_input(self):
+           if not self.name:
+               raise UseFallback()
+
+       @field
```
kmike, line 76:
```suggestion
       @field(cached=True)
```

We might actually want to document this gotcha as well - if a field is used in validate_input, it's better to use cached=True, because otherwise if validation is happy, the field will be evaluated twice.

docs/page-objects/input-validation.rst
```diff
@@ -0,0 +1,95 @@
+.. _input-validation:
+
+================
+Input validation
+================
+
+Sometimes the data that your page object receives as input may be invalid.
+
+You can define a ``validate_input`` method in a page object class to check its
```
BurnzZ, line 9: I think it could be worth mentioning that `validate_input()` is called within `.to_item()` method. This reminds users who completely override the `.to_item()` method without calling super to also call `validate_input()` .

kmike: There also might be some examples in the docs to update, where we define to_item method.

kmike: There could also be cases where you don't want to call super() in to_item, but still want to have the validation.

Gallaecio: I have addressed comments 1 and 3.

@kmike I am unsure about modifying existing examples to include `super().to_item` or input validation code, as it would complicate those examples. I think covering how to apply validation from a custom `to_item` as you suggested later may be enough.

docs/page-objects/additional-requests.rst
```diff
@@ -773,3 +773,53 @@ Here's an example:
...
+
+    import attrs
+    from tenacity import retry, stop_after_attempt
+    from web_poet import HttpClient, WebPage
+
+    @attrs.define
+    class MyPage(WebPage):
+        http: HttpClient
+
+        @retry(stop=stop_after_attempt(3))
```
VMRuiz, line 809: We should encapsulate 3rd party lib functionality into wrappers in case we need to migrate the functionality to other library on the future.

Gallaecio: How people implement retries for this scenario is not part of the web-poet spec, this is just an example. People are free to use something other than tenacity.

docs/page-objects/input-validation.rst
Gallaecio: @kmike Sorry to get back to you about this, after we just discussed it a few hours ago off-GitHub, but could you elaborate on which examples you would update and how? Maybe put 1 example?

docs/page-objects/additional-requests.rst
```diff
@@ -773,3 +773,53 @@ Here's an example:
...
+            if not response.css(".expected"):
+                raise ValueError
+            return response.css(".data").get()
+
+        async def to_item(self) -> dict:
+            try:
+                data = await self.get_data()
+            except ValueError:
+                return {}
+            return {"data": data}
```
kmike, line 821: That's an example of a page object where validate_input won't work properly, if user decides to extend the page object with some validation. User might need to do something like this:

```suggestion
        async def to_item(self) -> dict:        
            if item := self._validate_input():
                return item
            try:
                data = await self.get_data()
            except ValueError:
                return {}
            return {"data": data}
```

which doesn't look very nice, and uses a private method as well.

One may also try `item = super().to_item()`, e.g. to add more information in a subclass. But it could be wrong. For example, validate_input of this class (or of a base class) might return an item. We say that this item is going to be used instead of to_item result. And indeed, `super().to_item()` is going to return this item, even if validate_input is implemented in a subclass. But then to_item method may start adding more things to the item, which means that the item is going to be different.

docs/page-objects/input-validation.rst
kmike: See https://github.com/scrapinghub/web-poet/pull/113/files#r1143860051. I just realized that `super().to_item()` may not be the right suggestion.

docs/page-objects/additional-requests.rst
Gallaecio: We can definitely make `_validate_input` public. Would that be enough?

Alternatively, we could reconsider switching the “override item” logic from a return value to another exception, so that here we can go from 2 lines to 1 (i.e. save the return line).

Gallaecio: [Done](https://github.com/scrapinghub/web-poet/pull/113/commits/9819d57049b33cc7014cf8cb21bd4027a8c75d29).

The decorator is used in ItemPage.to_item, so existing tests cover the async scenario.

For the sync scenario, we had `test_invalid_input_sync_to_item` already, but it was passing. It was passing because the custom `to_item` implementation happened to be returning the same object as it would with input validation properly enabled, because input validation was still applied at the field level. I added an extra `is_valid` field to the `Item` class in tests that defaults to `True`, so that it is only `False` when that field is set in an `invalid_input` method. After that, the test started failing until you added the `@validate_input` decorator.

Issue #115, opened 2023-01-20
Allowing subfields to be extracted


BurnzZ:
Stemming from https://github.com/scrapinghub/scrapy-poet/pull/111 where we'd want to implement the API in **web-poet** itself regarding extracting data from a subset of fields.

# API

The main directives that we want to support are:

- `include_fields`
- `exclude_fields`

Using both directives together should not be allowed. 

## via page object instance

```python
item = partial_item_from_page_obj(product_page, include_fields=["x", "y"])
print(item)  # ProductItem(x=1, y=2, z=None)
print(type(product_page))  # ProductPage
```

This API assumes we already have an instance of the page object with the appropriate response data in it. Moreover, the item class can be inferred from the page object definition:

```python
class ProductPage(WebPage[ProductItem]):
    ...
```

Arguably, we could also support page object classes as long as the URL is provided for the response data to be downloaded by the configured [downloader](https://web-poet.readthedocs.io/en/stable/frameworks/additional-requests.html).

## via item class

Conversely, we could also support directly asking for the item class instead of the page object as long as we have access to the `ApplyRule` to infer their relationships. Unlike the page object, a single item class could have relationships to multiple page objects, depending on the URL.

But this means that the response should still be downloaded and the [downloader](https://web-poet.readthedocs.io/en/stable/frameworks/additional-requests.html) is configured.

```python
item = partial_item_from_item_cls(
    ProductItem, include_fields=["x", "y"], url="https://example.com"
)
```

There are several combinations of scenarios for this type of API.

# Page object setup

## A. The page object has all fields using the `@field` decorator

This is quite straightforward to support since we can easily do:

```python
from web_poet.fields import get_fields_dict
from web_poet.utils import ensure_awaitable


fields = get_fields_dict(page_obj)
item_dict = item_from_fields_sync(page_obj, item_cls=dict, skip_nonitem_fields=False)
item = page_obj.item_cls(
    **{
        name: await ensure_awaitable(item_dict[name])
        for name in item_dict
        if name in field_names
    }
)
```

We basically derive all the fields from the page object and call them one-by-one.

## B. The page object doesn't use the `@field` decorator but solely utilizes the `.to_item()` method

Alternatively, the page object can be defined as:

```python
class ProductPage(WebPage[ProductItem]):
    def to_item(self) -> ProductItem:
        return ProductItem(x=1, y=2, z=3)
```

[...]

PR #118, opened 2023-01-27
Introduce `SelectFields`

BurnzZ:
This is related to #115 which implements [approach 3](https://github.com/scrapinghub/web-poet/issues/115#issuecomment-1408526596).

`swap_item_cls` is omitted for now to reduce its complexity. We can add it later.

### TODO:

*    [ ]  docs
*    [ ]  changelog

### DIFF
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -743,6 +743,14 @@ async def test_select_fields() -> None:
     assert page.fields_to_extract == ["x", "y"]
     assert page.call_counter == {"x": 1, "y": 1}
 
+    # If a non-SelectFields instance was passed to the `select_fields` init
+    # parameter, it would simply ignore it and use the default field directives.
+    page = BigPage(select_fields="not the instance it's expecting")  # type: ignore[arg-type]
+    item = await page.to_item()
+    assert item == BigItem(x=1, y=2, z=None)
+    assert page.fields_to_extract == ["x", "y"]
+    assert page.call_counter == {"x": 1, "y": 1}
+
     # The remaining tests below checks the different behaviors when encountering a
     # field which doesn't existing in the PO
     fields = {"x": True, "not_existing": True}
@@ -783,19 +791,14 @@ async def test_select_fields() -> None:
         assert page.fields_to_extract == ["x", "y"]
     assert page.call_counter == {"x": 1, "y": 1}
 
-
-@pytest.mark.asyncio
-async def test_select_fields_on_unknown_field_bad_value() -> None:
     # When SelectFields receive an invalid 'on_unknown_field' value, it should
-    # error out as well.
-    invalid_val = "invalid val"
-    expected_value_error_msg = (
-        f"web_poet.SelectFields only accepts 'ignore', 'warn', and 'raise' "
-        f"values. Received unrecognized '{invalid_val}' value which it treats as "
-        f"'ignore'."
+    # ignore it. Moreover, mypy should also raise an error (see
+    # tests_typing/test_fields)
+    page = BigPage(
+        # ignore mypy error since it's expecting a valid value inside the Literal.
+        SelectFields(fields=fields, on_unknown_field="bad val")  # type: ignore[arg-type]
     )
-    with pytest.raises(ValueError, match=expected_value_error_msg):
-        await BigPage(
-            # ignore mypy error since it's expecting a valid value inside the Literal.
-            SelectFields(fields=["y", "not_existing"], on_unknown_field=invalid_val)  # type: ignore[arg-type]
-        ).to_item()
+    item = await page.to_item()
+    assert item == BigItem(x=1, y=2, z=None)
+    assert page.fields_to_extract == ["x", "y"]
+    assert page.call_counter == {"x": 1, "y": 1}
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -652,98 +652,136 @@ class BigPage(ItemPage[BigItem]):
         self.call_counter["x"] += 1
         return 1
 
-    @field
+    @field(disabled=False)
     def y(self):
         self.call_counter["y"] += 1
         return 2
 
-    @field
+    @field(disabled=True)
     def z(self):
         self.call_counter["z"] += 1
         return 3
 
 
 @pytest.mark.asyncio
-async def test_select_fields_include() -> None:
-    # Basic case
-    page = BigPage(SelectFields(include=["x", "y"]))
-    item = await page.to_item()
-    assert item == BigItem(x=1, y=2, z=None)
-    assert page.fields_to_extract == {"x", "y"}
-    assert page.call_counter == {"x": 1, "y": 1}
-
-    # Repeated fields are ignored
-    page = BigPage(SelectFields(include=["x", "x"]))
-    item = await page.to_item()
-    assert item == BigItem(x=1, y=None, z=None)
-    assert page.fields_to_extract == {"x"}
-    assert page.call_counter == {"x": 1}
-
-    # Passing None value results in all fields to be extracted.
-    # Note that this is different when [] is passed. See test below.
-    page = BigPage(SelectFields(include=None))
-    item = await page.to_item()
-    assert item == BigItem(x=1, y=2, z=3)
-    assert page.fields_to_extract == {"x", "y", "z"}
-    assert page.call_counter == {"x": 1, "y": 1, "z": 1}
-
+async def test_select_fields() -> None:
     # Required fields from the item cls which are not included raise an TypeError
     expected_type_error_msg = (
         r"__init__\(\) missing 1 required positional argument: 'x'"
     )
-    page = BigPage(SelectFields(include=[]))
-    with pytest.raises(TypeError, match=expected_type_error_msg):
-        item = await page.to_item()
-    assert page.fields_to_extract == set()
-    assert page.call_counter == {}
 
-    page = BigPage(SelectFields(include=["y", "z"]))
+    # When SelectFields isn't set, it should simply extract the non-disabled
+    # fields.
+    page = BigPage()
+    item = await page.to_item()
+    assert item == BigItem(x=1, y=2, z=None)
+    assert page.fields_to_extract == ["x", "y"]
+    assert page.call_counter == {"x": 1, "y": 1}
+
+    # If no field selection directive is given but SelectFields is set, it would
+    # use the default fields that are not disabled.
+    page = BigPage(SelectFields(fields=None))
+    item = await page.to_item()
+    assert item == BigItem(x=1, y=2, z=None)
+    assert page.fields_to_extract == ["x", "y"]
+    assert page.call_counter == {"x": 1, "y": 1}
+
+    # Same case as above but given an empty dict.
+    page = BigPage(SelectFields(fields={}))
+    item = await page.to_item()
+    assert item == BigItem(x=1, y=2, z=None)
+    assert page.fields_to_extract == ["x", "y"]
+    assert page.call_counter == {"x": 1, "y": 1}
+
+    # Select all fields
+    page = BigPage(SelectFields(fields={"*": True}))
+    item = await page.to_item()
+    assert item == BigItem(x=1, y=2, z=3)
+    assert page.fields_to_extract == ["x", "y", "z"]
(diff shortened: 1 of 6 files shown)


web_poet/pages.py
```diff
@@ -63,10 +70,98 @@ def __init_subclass__(cls, skip_nonitem_fields: bool = False, **kwargs):
...
     async def to_item(self) -> ItemT:
         """Extract an item from a web page"""
+
+        partial_item = await self._partial_item()
+        if partial_item:
+            return partial_item
+
         return await item_from_fields(
             self, item_cls=self.item_cls, skip_nonitem_fields=self._skip_nonitem_fields
         )
```
BurnzZ, line 81: `_get_select_fields()` is called at least twice on the page but it's value is not being cached.

I think this should be okay since the user might want to re-use the same page instance but with different `SelectFields`:

```
page = SomePage(SelectFields(fields={"*": False, "x": True}))
item1 = page.to_item()

page.select_fields = SelectFields(fields={"*": False, "x": True, "y": True})
item2 = page.to_item()
```

web_poet/pages.py
```diff
@@ -63,10 +70,98 @@ def __init_subclass__(cls, skip_nonitem_fields: bool = False, **kwargs):
...
 
+    def _get_select_fields(self) -> Optional[SelectFields]:
+        select_fields = getattr(self, "select_fields", None)
```
BurnzZ, line 83: `_get_select_fields()` currently require (and assumes) that the `select_field` instance attribute contains `SelectFields`. We could reflect this in our docs as a requirement.

I think this is the simplest one for now.

---

An alternative would be to support dynamic variable names which could have names other than `select_field`. This means we need to do introspection to find any instances of `SelectFields` while iterating through its instance attributes. This also means that we need to handle edge cases when two or more `SelectFields` are present in a given page. We need a mechanism on how to choose one of them and reflect them in our docs.

This seems to be a more complex and messy approach.

---

Another alternative would be to set `select_fields: Optional[SelectFields]` as a dependency in `ItemPage`. Although I'm not quite sure about this at the moment.

web_poet/pages.py
```diff
@@ -63,10 +70,98 @@ def __init_subclass__(cls, skip_nonitem_fields: bool = False, **kwargs):
...
+
+        if not isinstance(select_fields, SelectFields):
+            return None
+
+        return select_fields
+
+    @property
+    def fields_to_extract(self) -> Iterable[str]:
+        """Returns an Iterable of field names which should populate the designated
+        ``item_cls``.
```
BurnzZ, line 95: I'm proposing to expose the `fields_to_extract` property method since it's a handy utility for frameworks like **scrapy-poet** to see which fields to populate.

web_poet/pages.py
```diff
@@ -64,9 +71,80 @@ def __init_subclass__(cls, skip_nonitem_fields: bool = False, **kwargs):
...
+        fields = select_fields.fields
+
+        if fields is None or len(fields) == 0:
+            return list(get_fields_dict(self))
+
+        page_obj_fields = get_fields_dict(self, include_disabled=True)
+
+        unknown_fields = set(fields) - set(page_obj_fields.keys()).union({"*"})
+        if unknown_fields:
+            self._handle_unknown_field(unknown_fields, select_fields.on_unknown_field)
```
BurnzZ, line 112: The handling of unknown fields seems to be more optimal to be placed here as opposed to something like `item_from_fields()` which could've already extracted some fields (which could be expensive, i.e., having additional request) before erroring out.

web_poet/pages.py
```diff
@@ -64,9 +71,80 @@ def __init_subclass__(cls, skip_nonitem_fields: bool = False, **kwargs):
...
+        msg = (
+            f"The {name} fields isn't available in {get_fq_class_name(self.__class__)}."
+        )
+
+        if action == "raise":
+            raise AttributeError(msg)
+        elif action == "warn":
+            warn(msg)
+        else:
+            raise ValueError(
```
BurnzZ, line 142: I'm thinking if we still need this runtime check somehow since mypy should be good enough to detect invalid values due to the `Literal` typing.

web_poet/fields.py
```diff
@@ -172,13 +182,18 @@ def get_fields_dict(
...
 # FIXME: type is ignored as a workaround for https://github.com/python/mypy/issues/3737
 # inference works properly if a non-default item_cls is passed; for dict
 # it's not working (return type is Any)
 async def item_from_fields(
-    obj, item_cls: Type[T] = dict, *, skip_nonitem_fields: bool = False  # type: ignore[assignment]
+    obj,
+    item_cls: Type[T] = dict,  # type: ignore[assignment]
+    *,
+    skip_nonitem_fields: bool = False,
+    field_names: Optional[Iterable[str]] = None,
```
BurnzZ, line 196: If an unknown field has been unintentionally passed, it'd error out with:

```
# omitted stack trace

File ~/dev/z/web-poet/web_poet/fields.py:235, in <dictcomp>(.0)
    233     if skip_nonitem_fields:
    234         field_names = _without_unsupported_field_names(item_cls, field_names)
--> 235 return item_cls(**{name: getattr(obj, name) for name in field_names})

AttributeError: 'SomePage' object has no attribute 'unknown_field'
```

I think this is descriptive enough and we don't need additional handling of unknown fields here. Handling it in the `.fields_to_extract` property method should be enough.

web_poet/pages.py
BurnzZ: Removed it.

web_poet/pages.py
```diff
@@ -64,9 +71,74 @@ def __init_subclass__(cls, skip_nonitem_fields: bool = False, **kwargs):
...
         """Extract an item from a web page"""
         return await item_from_fields(
-            self, item_cls=self.item_cls, skip_nonitem_fields=self._skip_nonitem_fields
+            self,
+            item_cls=self.item_cls,
+            skip_nonitem_fields=self._skip_nonitem_fields,
+            field_names=self.fields_to_extract,
         )
 
+    def _get_select_fields(self) -> Optional[SelectFields]:
```
BurnzZ, line 80: \`\_get\_select\_fields()\` is not being cached.

I think this should be okay since the user might want to re-use the same page instance but with different \`SelectFields\`:

\`\`\`python  
page = SomePage(SelectFields(fields={"\*": False, "x": True}))  
item1 = page.to\_item()

page.select\_fields = SelectFields(fields={"\*": False, "x": True, "y": True})  
item2 = page.to\_item()  
\`\`\`

Any thoughts about this usage?

tests/test_fields.py
```diff
@@ -630,3 +633,172 @@ def x(self) -> int:
...
+    assert page.call_counter == {"x": 1, "y": 1, "z": 1}
+
+    # Excluding a required field throws an error
+    page = BigPage(SelectFields(fields={"x": False}))
+    with pytest.raises(TypeError, match=expected_type_error_msg):
+        item = await page.to_item()
+    assert page.fields_to_extract == ["y"]
+    assert page.call_counter == {"y": 1}
+
+    # boolean-like values are not supported. They are simply ignored and the
```
kmike, line 738: I think we should raise an error, not ignore wrong values.

tests/test_fields.py
```diff
@@ -630,3 +633,172 @@ def x(self) -> int:
...
+    # page would revert back to the default field directives.
+    page = BigPage(SelectFields(fields={"x": 0, "y": 0, "z": 1}))  # type: ignore[dict-item]
+    item = await page.to_item()
+    assert item == BigItem(x=1, y=2, z=None)
+    assert page.fields_to_extract == ["x", "y"]
+    assert page.call_counter == {"x": 1, "y": 1}
+
+    # If a non-SelectFields instance was passed to the `select_fields` init
+    # parameter, it would simply ignore it and use the default field directives.
```
kmike, line 747: It makes more sense to me to raise an error in this case.

web_poet/fields.py
```diff
@@ -200,19 +220,43 @@ async def item_from_fields(
...
+    #: mapping of field names to boolean values to where ``True`` would indicate
+    #: it being included in ``.to_item()`` calls.
+    fields: Optional[MutableMapping[str, bool]] = None
+
+    #: Controls what happens when encountering an unknown field. For example,
+    #: an unknown field was passed and the page object doesn't recognize it.
+    #:
+    #: Setting it to ``"raise"`` would raise an :class:`AttributeError`,
+    #: ``"warn"`` produces a :class:`UserWarning`, while ``"ignore"`` does nothing.
+    on_unknown_field: UnknownFieldActions = "raise"
```
kmike, line 262: Hey! Could you please elaborate on the use cases for on_unknown_field?

Also, what does unknown field mean? Does it mean that the field is not in the item, or does it mean that the field is not defined on the page object?

BurnzZ: > does it mean that the field is not defined on the page object?

It's mostly regarding about this but because of the parity between the page object and its item class, the field should generally not be in the item as well.

The unknown field can be derived by something like `set(fields_to_include) - set(fields_to_exclude)`. This means unknown fields that are to be excluded won't be considered as actual unknown fields.

I'm also thinking of removing this parameter and simply raise any unknown fields by default. What do you think?

kmike: Aha, I see. 

It seems the case with field not in the item and a field not in the page object are very different. 

If a field is not in the item, we should probably raise an exception. But there is a case where skip_nonitem_fields=True is used; in this case it actually could make sense to have fields in SelectFields which are not in the item.

If a field is not defined in the page object, but is present in the item, then final item.field returned by to_item is going to have a default value, None in most cases. It could also be the case that the attribute is populated in to_item method without using web-poet fields. It's not an error. It seems that we should allow to pick any fields which are present in the item, even if they're not fields of the page object.

We should have a clear answer though what happens when some of the item attributes are extracted using `@field`s, and some of them are added later, in the to_item method.

There is also a complicated case where there is a field defined, but then the value assigned to the final item is changed in to_item method. Probably it's fine not to support this case, but we should at least document it.

Overall the problem looks more complicated than I initially thought :)

web_poet/pages.py
Gallaecio: Should we choose a name in line with the corresponding attribute of `@field` to exclude or include fields from `to_item` output? (e.g. `to_item_fields`)

web_poet/pages.py
Gallaecio: I am OK with not caching.

However, I think we should avoid covering this scenario in the docs, and I am not sure if we should even cover it in tests. Changing the value of `select_fields` does not feel like a very elegant API to me.

web_poet/fields.py
Gallaecio: When I suggested such an option, I was thinking of a scenario like this one:

B team writes `BPage`, subclassing `APage` from A team, and returning `BItem`, which does *not* subclass `AItem`. B team might not want new fields in `APage` to break `BPage`. But they might want a warning or error to look into it eventually in case we want to do something about it.

But it can indeed be more complex than that. When a field is removed, B team may want a different resolution. For example, they might want a warning for new fields, but an exception for missing fields.

web_poet/pages.py
BurnzZ: Apparently, there's a small problem when adding `ItemPage` as optional:

```python
@attrs.define
class ItemPage(Injectable, Returns[ItemT]):
     select_fields: typing.Optional[SelectFields] = None
     _skip_nonitem_fields: typing.ClassVar[bool]

@attrs.define
class WebPage(ItemPage[ItemT], ResponseShortcutsMixin):
    response: HttpResponse
```

**attrs** raises an exception:

```
ValueError: No mandatory attributes allowed after an attribute with a default value or factory.  Attribute in question: Attribute(name='response', default=NOTHING, validator=None, repr=True, eq=True, eq_key=None, order=True, order_key=None, hash=None, init=True, metadata=mappingproxy({}), type=<class 'web_poet.page_inputs.http.HttpResponse'>, converter=None, kw_only=False, inherited=False, on_setattr=None, alias=None)
```

It also breaks the API for `WebPage(HttpResponse("", b""))` since users would now have to `WebPage(SelectFields(fields={"*": True}), HttpResponse("", b""))`

However, there's a way around it by using `@attrs.define(kw_only=True)`. I think this is much better though since it makes the API more compatible in the future. I'll implement this in the another PR.

BurnzZ:
Closing in favor of #138

PR #120, opened 2023-01-30
introduce @field(disabled=True)

BurnzZ:
Similar to https://github.com/scrapinghub/web-poet/pull/63 but this requires the field to be controlled with `@field(disabled=True)`. It also requires that the disabled field be available in the item class.

``` python
class ArticlePage(WebPage[Article]):

    @field
    def url(self):
        ...

    # other standard fields
    # ...

    @field(disabled=True)
    def articleBodyRaw(self):
        ...

page = ArticlePage(response)
item = await page.to_item()
# ArticlePage(url="https://example.com", ..., articleBodyRaw=None)
```

This enables some fields to not be included when `.to_item()` is called. Related to https://github.com/scrapinghub/web-poet/pull/118 (_the API would change with respect to this_), the fields set with `@field(disabled=True)` can be selected later on.

I also think that this PR and https://github.com/scrapinghub/web-poet/pull/118 should be released together. Otherwise, disabled fields can't really be utilized at all.

### To be decided:
- [ ] should we go with the `disabled` terminology? or perhaps something else.

### TODO:
- Docs will be handled in https://github.com/scrapinghub/web-poet/pull/118 alongside how to select disabled fields.

### DIFF
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -1,5 +1,6 @@
 import asyncio
 import random
+from typing import Optional
 
 import attrs
 import pytest
@@ -26,7 +27,7 @@ from web_poet import (
     item_from_fields,
     item_from_fields_sync,
 )
-from web_poet.fields import get_fields_dict
+from web_poet.fields import FieldInfo, get_fields_dict
 
 
 @attrs.define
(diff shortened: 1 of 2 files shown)


web_poet/fields.py
```diff
@@ -16,6 +16,10 @@
 _FIELDS_INFO_ATTRIBUTE_WRITE = "_web_poet_fields_info_temp"
 
 
+def _fields_template():
+    return {"enabled": {}, "disabled": {}}
```
BurnzZ, line 20: An alternative approach is to declare something like:

```
_FIELDS_INFO_ATTRIBUTE_READ = "_web_poet_fields_info"
_FIELDS_INFO_ATTRIBUTE_WRITE = "_web_poet_fields_info_temp"
_FIELDS_INFO_ATTRIBUTE_READ_DISABLED = "_web_poet_fields_info_disabled"
_FIELDS_INFO_ATTRIBUTE_WRITE_DISABLED = "_web_poet_fields_info_temp_disabled"
```

But I think it could be unwieldly when more field-flags are introduced later on.

tests/test_fields.py
```diff
@@ -522,3 +523,110 @@ def b(self):
...
+    class Page(ItemPage[Item]):
+        @field
+        def x(self) -> int:
+            return 1
+
+        @field(disabled=False)
+        def y(self) -> int:
+            return 2
+
+        @field(disabled=True)
```
Gallaecio, line 545: Maybe `enabled` would be a better name.

I would prefer something that suggests how the field can still be used as a property but is not included in the `to_item` output, but I cannot think of a good name (`export`?  `itemize`? `itemizable`? `to_item`? `add_to_item`? `include_in_item`?).

BurnzZ: My initial implementation actually used `enabled`, but it seems `get_fields_dict(include_disabled=True)` sounds much better.

Hmmm, `to_item` and `include_in_item` does sound better in terms of what it does. Currently, either `enabled` or `disabled` isn't that clear about what it does.

+1 on something like `@field(to_item=False)`.

kmike: I was initially thinking about something like "exclude_by_default=True" or "excluded_by_default=True", or "extract_by_default=False"; alternatives like "to_item=False" also look fine. 

A possible issue with to_item=False is that you may think such field doesn't have to be defined in the item, but it does; I think it's still an error if item doesn't have it.

BurnzZ: > I was initially thinking about something like "exclude_by_default=True" or "excluded_by_default=True", or "extract_by_default=False"

I think these have the same issue with `include` and `exclude` where it's not clear what's happening, i.e., where it's being excluded from.

> A possible issue with to_item=False is that you may think such field doesn't have to be defined in the item, but it does; I think it's still an error if item doesn't have it.

You're right. What do you think about having something like `omit_in_to_item=True` or `to_item_omit=True`?

web_poet/fields.py
```diff
@@ -125,12 +147,20 @@ def processed(*args, **kwargs):
         return _field
 
 
-def get_fields_dict(cls_or_instance) -> Dict[str, FieldInfo]:
+def get_fields_dict(
+    cls_or_instance, include_disabled: bool = False
+) -> Dict[str, FieldInfo]:
     """Return a dictionary with information about the fields defined
     for the class: keys are field names, and values are
     :class:`web_poet.fields.FieldInfo` instances.
```
BurnzZ, line 155: TODO: update this docstring when `disabled` has been renamed to a better one.

AdrianAtZyte:
Shall we close this?

PR #121, opened 2023-01-31
bump black and isort versions

BurnzZ:
Similar to scrapinghub/scrapy-poet@140239a which fixes the broken "linters" tox CI.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,13 +4,13 @@ repos:
         language_version: python3
         exclude: ^docs/tutorial-project/
     repo: https://github.com/ambv/black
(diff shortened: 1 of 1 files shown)


kmike:
Thanks @BurnzZ!

FTR, I haven't bumped isort to the latest version because it dropped support for Python 3.7. It's not a big deal, as we can run isort in 3.8+ environments, but it might be an issue for some user who makes changes locally.

Gallaecio:
> but it might be an issue for some user who makes changes locally.

[I did not take that into account for Scrapy](https://github.com/scrapy/scrapy/pull/5806) :sweat:

PR #123, opened 2023-01-31
Indent content of HttpResponse-other.json

kmike:
It seems that without indenting "headers" field make HttpResponse-other.json quite unreadable in practice.

### DIFF
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -66,7 +66,12 @@ def test_serialization(book_list_html_response) -> None:
     url = ResponseUrl(url_str)
 
     serialized_deps = serialize([book_list_html_response, url])
-    other_json = f'{{"_encoding": "utf-8", "headers": [], "status": null, "url": "{url_str}"}}'.encode()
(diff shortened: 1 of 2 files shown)


wRAR (approved):

tests/test_serialization.py
```diff
@@ -66,7 +66,12 @@ class MyWebPage(WebPage):
     url = ResponseUrl(url_str)
 
     serialized_deps = serialize([book_list_html_response, url])
-    other_json = f'{{"_encoding": "utf-8", "headers": [], "status": null, "url": "{url_str}"}}'.encode()
+    other_json = f"""{{
+  "_encoding": "utf-8",
+  "headers": [],
+  "status": null,
+  "url": "{url_str}"
+}}""".encode()
```
kmike, line 74: I don't love this test, it was tricky to get it right, and the diffs in case of failure are obscure, because they include complete httpResponseBody content. Not sure how to improve on it, maybe we can leave it as-is for now.

PR #128, opened 2023-02-07
Add saving/loading of HttpClient responses

wRAR:
This adds two flags to HttpClient: one tells it to save all responses into an attribute, another one tells it to return saved responses if they are available and raise an exception if they are not. It also adds serialization support for HttpClient that saves/restores these responses and sets the second flag on the deserialized HttpClient. I don't currently see how can we set the first flag when we generate a test, maybe it shouldn't be a flag of `__init__` but some other check.

This requires having some fingerprinting code for HttpRequest (the one in this PR is a placeholder). Not sure how comprehensive should it be. The Scrapy one is quite complicated taking into account that URLs can have fragments, not all headers are important etc. The code in scrapy-poet that gets a fingerprint for caching uses the Scrapy code as far as I remember, but here we can't depend on Scrapy. I also wonder if this fingerprinting needs to be user-configurable.

### DIFF
--- a/docs/page-objects/testing.rst
+++ b/docs/page-objects/testing.rst
@@ -63,6 +63,13 @@ it, that contains data for Page Object inputs and output::
         │   └── output.json
         └─── test-2
             ├── inputs
+            │   ├── HttpClient-0-HttpRequest.info.json
+            │   ├── HttpClient-0-HttpResponse.body.html
+            │   ├── HttpClient-0-HttpResponse.info.json
+            │   ├── HttpClient-1-HttpRequest.body.txt
+            │   ├── HttpClient-1-HttpRequest.info.json
+            │   ├── HttpClient-1-HttpResponse.body.html
+            │   ├── HttpClient-1-HttpResponse.info.json
             │   ├── HttpResponse-body.html
             │   ├── HttpResponse-info.json
             │   └── ResponseUrl.txt
--- a/docs/page-objects/testing.rst
+++ b/docs/page-objects/testing.rst
@@ -202,3 +202,24 @@ Please also check the official Git LFS documentation for more information.
 
 .. _Git LFS: https://git-lfs.com/
 .. _implementations: https://github.com/git-lfs/git-lfs/wiki/Implementations
+
+Additional requests support
+===========================
+
+If the page object uses the :class:`~.HttpClient` dependency to make
+:ref:`additional requests <additional-requests>`, the generated fixtures will
+contain these requests and their responses. When the test runs,
+:class:`~.HttpClient` will return the saved responses without doing actual
+requests.
+
+Currently requests are compared by their URL, method, headers and body, so if a
+page object makes requests that differ between runs, the test won't be able to
+find a saved response and will fail.
+
+Test coverage
+=============
+
+The coverage for page object code is reported correctly if tools such as
+`coverage`_ are used when running web-poet tests.
+
+.. _coverage: https://coverage.readthedocs.io/
--- a/docs/page-objects/testing.rst
+++ b/docs/page-objects/testing.rst
@@ -57,14 +57,14 @@ it, that contains data for Page Object inputs and output::
         ├── test-1
         │   ├── inputs
         │   │   ├── HttpResponse-body.html
-        │   │   ├── HttpResponse-other.json
+        │   │   ├── HttpResponse-info.json
         │   │   └── ResponseUrl.txt
         │   ├── meta.json
         │   └── output.json
         └─── test-2
             ├── inputs
             │   ├── HttpResponse-body.html
-            │   ├── HttpResponse-other.json
+            │   ├── HttpResponse-info.json
             │   └── ResponseUrl.txt
             ├── meta.json
             └── output.json
--- a/tests/test_testing.py
+++ b/tests/test_testing.py
@@ -177,17 +177,19 @@ def _get_fp_for_url(url: str) -> str:
 
 
 def test_httpclient(pytester, book_list_html_response) -> None:
-    client = HttpClient()
     body1 = HttpResponseBody(b"body1")
     url1 = "http://books.toscrape.com/1.html"
     response1 = HttpResponse(url=url1, body=body1, encoding="utf-8")
     fp1 = _get_fp_for_url(url1)
-    client.saved_responses[fp1] = response1
     body2 = HttpResponseBody(b"body2")
     url2 = "http://books.toscrape.com/2.html"
     response2 = HttpResponse(url=url2, body=body2, encoding="utf-8")
     fp2 = _get_fp_for_url(url2)
-    client.saved_responses[fp2] = response2
+    responses = {
+        fp1: response1,
+        fp2: response2,
+    }
+    client = HttpClient(responses=responses)
 
     base_dir = pytester.path / "fixtures" / get_fq_class_name(ClientPage)
     item = {
--- a/tests/test_testing.py
+++ b/tests/test_testing.py
@@ -30,7 +30,7 @@ def test_save_fixture(book_list_html_response, tmp_path) -> None:
         assert (input_dir / "HttpResponse-body.html").read_bytes() == bytes(
             book_list_html_response.body
         )
-        assert (input_dir / "HttpResponse-other.json").exists()
+        assert (input_dir / "HttpResponse-info.json").exists()
         assert (directory / OUTPUT_FILE_NAME).exists()
         assert json.loads((directory / OUTPUT_FILE_NAME).read_bytes()) == item
         if expected_meta:
@@ -195,8 +195,8 @@ def test_httpclient(pytester, book_list_html_response) -> None:
     assert (input_dir / "HttpResponse-body.html").read_bytes() == bytes(
         book_list_html_response.body
     )
-    assert (input_dir / "HttpClient-0-HttpRequest.other.json").exists()
-    assert (input_dir / "HttpClient-0-HttpResponse.other.json").exists()
+    assert (input_dir / "HttpClient-0-HttpRequest.info.json").exists()
+    assert (input_dir / "HttpClient-0-HttpResponse.info.json").exists()
     assert (input_dir / "HttpClient-0-HttpResponse.body.html").read_bytes() == bytes(
         body1
     )
--- a/tests/test_testing.py
+++ b/tests/test_testing.py
@@ -12,6 +12,7 @@ from itemadapter import ItemAdapter
 from zyte_common_items import Item, Metadata, Product
 
 from web_poet import HttpClient, HttpRequest, HttpResponse, WebPage
+from web_poet.page_inputs.client import SavedResponseData
 from web_poet.testing import Fixture
 from web_poet.testing.fixture import INPUT_DIR_NAME, META_FILE_NAME, OUTPUT_FILE_NAME
 from web_poet.utils import get_fq_class_name
@@ -264,8 +265,8 @@ def test_httpclient(pytester, book_list_html_response) -> None:
     request2 = HttpRequest(url2, method="POST", body=b"post")
     response2 = HttpResponse(url=url2, body=b"body2", encoding="utf-8")
     responses = [
-        (request1, response1),
-        (request2, response2),
+        SavedResponseData(request1, response1),
+        SavedResponseData(request2, response2),
     ]
     client = HttpClient(responses=responses)
 
@@ -292,7 +293,7 @@ def test_httpclient_no_response(pytester, book_list_html_response) -> None:
     request = HttpRequest(url)
     response = HttpResponse(url=url, body=b"body1", encoding="utf-8")
     responses = [
-        (request, response),
+        SavedResponseData(request, response),
     ]
     client = HttpClient(responses=responses)
 
--- a/tests/test_testing.py
+++ b/tests/test_testing.py
@@ -11,7 +11,7 @@ import time_machine
 from itemadapter import ItemAdapter
 from zyte_common_items import Item, Metadata, Product
 
-from web_poet import HttpResponse, WebPage
+from web_poet import HttpClient, HttpRequest, HttpResponse, HttpResponseBody, WebPage
 from web_poet.testing import Fixture
 from web_poet.testing.fixture import INPUT_DIR_NAME, META_FILE_NAME, OUTPUT_FILE_NAME
 from web_poet.utils import get_fq_class_name
@@ -158,3 +158,48 @@ def test_pytest_frozen_time_tz_windows_pass(pytester, book_list_html_response) -
         2022, 3, 4, 20, 21, 22, tzinfo=dateutil.tz.tzlocal()
     )
     _assert_frozen_item(frozen_time, pytester, book_list_html_response)
+
+
+@attrs.define
+class ClientPage(WebPage):
+    client: HttpClient
+
+    async def to_item(self) -> dict:  # noqa: D102
+        resp1 = await self.client.get("http://books.toscrape.com/1.html")
+        resp2 = await self.client.get("http://books.toscrape.com/2.html")
+        return {"foo": "bar", "additional": [resp1.body.decode(), resp2.body.decode()]}
+
+
+def _get_fp_for_url(url: str) -> str:
+    req = HttpRequest(url=url)
+    return req.fingerprint()
+
+
+def test_httpclient_serialize(pytester, book_list_html_response) -> None:
+    client = HttpClient()
+    body1 = HttpResponseBody(b"body1")
+    url1 = "http://books.toscrape.com/1.html"
+    response1 = HttpResponse(url=url1, body=body1, encoding="utf-8")
+    fp1 = _get_fp_for_url(url1)
+    client.saved_responses[fp1] = response1
+    body2 = HttpResponseBody(b"body2")
+    url2 = "http://books.toscrape.com/2.html"
+    response2 = HttpResponse(url=url2, body=body2, encoding="utf-8")
+    fp2 = _get_fp_for_url(url2)
+    client.saved_responses[fp2] = response2
+
+    base_dir = pytester.path / "fixtures" / get_fq_class_name(ClientPage)
+    item = {
+        "foo": "bar",
+        "additional": ["body1", "body2"],
+    }
+    Fixture.save(base_dir, inputs=[book_list_html_response, client], item=item)
+    input_dir = base_dir / "test-1" / INPUT_DIR_NAME
+    assert (input_dir / "HttpResponse-body.html").read_bytes() == bytes(
+        book_list_html_response.body
+    )
+    assert (input_dir / f"HttpClient-{fp1}-other.json").exists()
+    assert (input_dir / f"HttpClient-{fp1}-body.html").read_bytes() == bytes(body1)
+    assert (input_dir / f"HttpClient-{fp2}-body.html").read_bytes() == bytes(body2)
+    result = pytester.runpytest()
+    result.assert_outcomes(passed=1)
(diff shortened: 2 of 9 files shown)


web_poet/page_inputs/client.py
```diff
@@ -165,8 +174,16 @@ async def execute(
         There is no need to include ``100-3xx`` status codes in ``allow_status``,
         because :class:`~.HttpResponseError` is not raised for them.
         """
+        response_key = request.fingerprint()
+        if self.return_only_saved_responses:
+            saved_response = self.saved_responses.get(response_key)
+            if saved_response:
+                return saved_response
+            raise ValueError(f"No saved response for {request}")
```
kmike, line 182: I think it may deserve an exception in exceptions.py

web_poet/page_inputs/http.py
```diff
@@ -181,6 +182,10 @@ def urljoin(self, url: Union[str, _RequestUrl, _ResponseUrl]) -> _RequestUrl:
         If *url* is relative, it is made absolute relative to :attr:`url`."""
         return _RequestUrl(urljoin(str(self.url), str(url)))
 
+    def fingerprint(self) -> str:
+        """Return the fingerprint of this request."""
+        return sha1(str(self.url).encode()).hexdigest()  # TODO
```
kmike, line 187: it should include method, request body and headers

web_poet/page_inputs/client.py
```diff
@@ -39,8 +39,17 @@ class HttpClient:
     :ref:`advanced-downloader-impl` documentation.
     """
 
-    def __init__(self, request_downloader: Callable = None):
+    def __init__(
+        self,
+        request_downloader: Callable = None,
+        *,
+        save_responses: bool = False,
+        return_only_saved_responses: bool = False,
```
kmike, line 47: 👍 for the argument names

web_poet/serialization/functions.py
```diff
@@ -60,3 +71,32 @@ def _deserialize__Url(cls: Type[_Url], data: SerializedLeafData) -> _Url:
...
+    cls: Type[HttpClient], data: SerializedLeafData
+) -> HttpClient:
+    serialized_responses: Dict[str, SerializedLeafData] = {}
+    for k, v in data.items():
+        response_key, subkey = k.rsplit("-", 1)
+        serialized_responses.setdefault(response_key, {})[subkey] = v
+
+    result = cls(return_only_saved_responses=True)
+    for response_key, serialized_response in serialized_responses.items():
+        result.saved_responses[response_key] = deserialize_leaf(
```
kmike, line 96: It seems to make sense to add saved_responses parameter to HttpClient.__init__, with an initial list of saved responses.

web_poet/page_inputs/http.py
kmike: I see this is a placeholder :)

web_poet/page_inputs/client.py
```diff
@@ -39,8 +39,17 @@ class HttpClient:
     :ref:`advanced-downloader-impl` documentation.
     """
 
-    def __init__(self, request_downloader: Callable = None):
+    def __init__(
+        self,
+        request_downloader: Callable = None,
+        *,
+        save_responses: bool = False,
```
kmike, line 46: > I don't currently see how can we set the first flag when we generate a test, maybe it shouldn't be a flag of __init__ but some other check.

So, HttpClient is provided here: https://github.com/scrapinghub/scrapy-poet/blob/cb79d3b870d3cb2cd3f9770e828481a0a51401b3/scrapy_poet/page_input_providers.py#L246

To change an argument for HttpClient, the provider must know if we're in the test generation phase or in the test running / regular execution phase. It seems this can be done through crawler - either through a custom setting (something like `_SCRAPY_POET_SAVEFIXTURE=True`), or by setting some crawler attribute (probably worse, but not sure).

web_poet/page_inputs/http.py
kmike: > I also wonder if this fingerprinting needs to be user-configurable.

It seems that can be desirable to make it configurable.
For example, we can add `request_fingerprinter` argument to `HttpClient.__init__`. 

The issue though is that we must use the same fingerprinter both during test generation and test running, which means test running is no longer independent from test generation. You can't e.g. reuse Scrapy's fingerprinter in test generation, and then run the tests.

Because of this, it looks like it's somewhat complex to make it configurable in a nice way. So, I'd focus on a reasonable default function for now. 

The goal of this fingerprinter is slightly different from e.g. Scrapy's, as it's not used for deduplication of requests, it's used for identification of requests. So, it seems it's fine for these fingerpints to be more strict.

We'll have issue, e.g. with random hashes, or with timestamps used as parts of URL, but it seems we'll be able to add customization options later; it doesn't look like a requirement for this PR.

kmike: To make it easier to customize later, I'd make fingerprinting a method of HttpClient, or a standalone function, not a method of HttpRequest.

web_poet/page_inputs/http.py
```diff
@@ -305,3 +302,14 @@ def _auto_detect_fun(self, body: bytes) -> Optional[str]:
...
                 continue
             return resolve_encoding(enc)
+
+
+def request_fingerprint(req: HttpRequest) -> str:
+    """Return the fingerprint of the request."""
+    fp = sha1()
+    fp.update(req.method.encode() + b"\n")
+    fp.update(canonicalize_url(str(req.url)).encode())
+    fp.update(str(req.url).encode() + b"\n")
```
kmike, line 312: why do you need to include url in the fingerprint twice?

web_poet/page_inputs/http.py
```diff
@@ -305,3 +302,14 @@ def _auto_detect_fun(self, body: bytes) -> Optional[str]:
...
+    for name, value in sorted(req.headers.items()):
```
kmike, line 313: What's the behavior of .items() when a header has multiple values for the same name?

wRAR: ["View contains all items, multiple items can have the same key."](https://multidict.aio-libs.org/en/stable/multidict.html#multidict.MultiDict.items)

web_poet/page_inputs/http.py
```diff
@@ -300,3 +302,14 @@ def _auto_detect_fun(self, body: bytes) -> Optional[str]:
             except UnicodeError:
                 continue
             return resolve_encoding(enc)
+
+
+def request_fingerprint(req: HttpRequest) -> str:
```
kmike, line 307: Let's add some basic tests for this function - e.g. that fingerprints of different requests are different, and fingerprints of similar (or equal) requests are the same.

web_poet/page_inputs/client.py
```diff
@@ -39,8 +40,20 @@ class HttpClient:
...
+        return_only_saved_responses: bool = False,
+        responses: Optional[Iterable[Tuple[HttpRequest, HttpResponse]]] = None,
+    ):
         self._request_downloader = request_downloader or _perform_request
+        self.save_responses = save_responses
+        self.return_only_saved_responses = return_only_saved_responses
+        self._saved_responses: Dict[str, Tuple[HttpRequest, HttpResponse]] = {
+            request_fingerprint(req): (req, resp) for req, resp in responses or {}
+        }
```
wRAR, line 56: I don't like this (and the related code in other places) that much, but the alternative is to calculate fingerprints on each comparison which doesn't look like a good idea to me, and we can't throw out requests because we need to serialize them. Maybe there is a better way to store this.

web_poet/page_inputs/http.py
```diff
@@ -300,3 +302,14 @@ def _auto_detect_fun(self, body: bytes) -> Optional[str]:
...
+    """Return the fingerprint of the request."""
+    fp = sha1()
+    fp.update(req.method.encode() + b"\n")
+    fp.update(canonicalize_url(str(req.url)).encode())
```
kmike, line 311: I think we need to add a separator here, so that `request_fingerprint(url+"X", header_name) != request_fingerprint(url, "X"+header_name)` if it makes sense :)

web_poet/serialization/functions.py
```diff
@@ -1,9 +1,55 @@
...
+def _deserialize_HttpRequest(
+    cls: Type[HttpRequest], data: SerializedLeafData
+) -> HttpRequest:
+    body = HttpRequestBody(data.get("body.txt", b""))
+    other_data = json.loads(data["other.json"])
+    return cls(
+        body=body,
+        url=RequestUrl(other_data["url"]),
+        method=other_data["method"],
+        headers=HttpRequestHeaders(other_data["headers"]),
```
kmike, line 48: I'd expect all these type conversions to be unnecessary, it should be possible to just pass the url and headers. Is it some weird typing issue?

web_poet/page_inputs/client.py
kmike: There are alternatives like using a cache for the fingerprinting function, but it looks fine as-is to me.

tests/test_testing.py
```diff
@@ -158,3 +158,70 @@ def test_pytest_frozen_time_tz_windows_pass(pytester, book_list_html_response) -
...
+    item = {
+        "foo": "bar",
+        "additional": ["body1", "body2"],
+    }
+    Fixture.save(base_dir, inputs=[book_list_html_response, client], item=item)
+    input_dir = base_dir / "test-1" / INPUT_DIR_NAME
+    assert (input_dir / "HttpResponse-body.html").read_bytes() == bytes(
+        book_list_html_response.body
+    )
+    assert (input_dir / "HttpClient-0-HttpRequest.other.json").exists()
```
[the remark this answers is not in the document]
wRAR: Yeah, this caught my attention too.

wRAR:
Should we also save the raised HTTP exceptions? Or is it not expected for a PO to normally get them?

kmike:
> Should we also save the raised HTTP exceptions? Or is it not expected for a PO to normally get them?

That's a very good question. I think we should. For example, page object makes a request, and the status code is not in allow_status. There is a valid try/except in the page object, like in https://web-poet.readthedocs.io/en/stable/page-objects/additional-requests.html#handling-exceptions-in-page-objects. If we don't store such exceptions, then tests won't work.

I'm fine with addressing it in a separate PR though.

tests/test_testing.py
```diff
@@ -158,3 +158,70 @@ def test_pytest_frozen_time_tz_windows_pass(pytester, book_list_html_response) -
...
+
+def test_httpclient(pytester, book_list_html_response) -> None:
+    body1 = HttpResponseBody(b"body1")
+    url1 = "http://books.toscrape.com/1.html"
+    request1 = HttpRequest(url1)
+    response1 = HttpResponse(url=url1, body=body1, encoding="utf-8")
+    body2 = HttpResponseBody(b"body2")
+    url2 = "http://books.toscrape.com/2.html"
+    request2 = HttpRequest(url2)
+    response2 = HttpResponse(url=url2, body=body2, encoding="utf-8")
```
kmike, line 181: The code can be simplified if you drop HttpResponseBody / HttpRequest wrapping, and use bytes/strings - here, and later in the test as well.

web_poet/exceptions/http.py
```diff
@@ -73,3 +73,20 @@ def __init__(
         if msg is None:
             msg = f"Unexpected HTTP Response received: {self.response}"
         super().__init__(msg, request=request)
+
+
+class NoSavedHttpResponse(HttpError):
```
kmike, line 78: Hm, I'm not sure it's good to have it as a subclass of HttpError. For example, if a page object handles HttpErrors (which is perfectly valid), then we might not get a test failure, as the page object might "handle" this NoSavedHttpResponse.

Should it be a subclass of AssertionError?

web_poet/serialization/functions.py
```diff
@@ -1,22 +1,59 @@
...
+        "method": o.method,
+        "headers": list(o.headers.items()),
+    }
+    result: SerializedLeafData = {
+        "info.json": json.dumps(
+            info, ensure_ascii=False, sort_keys=True, indent=2
+        ).encode(),
+    }
+    if o.body:
+        result["body.txt"] = bytes(o.body)
```
kmike, line 26: I think we should add tests for the missing lines.

kmike:
@wRAR I left some comments, but overall the approach looks solid 👍

docs/page-objects/testing.rst
```diff
@@ -202,3 +202,24 @@ Please also check the official Git LFS documentation for more information.
...
+page object makes requests that differ between runs, the test won't be able to
+find a saved response and will fail.
+
+Test coverage
+=============
+
+The coverage for page object code is reported correctly if tools such as
+`coverage`_ are used when running web-poet tests.
+
+.. _coverage: https://coverage.readthedocs.io/
```
Gallaecio, line 225: Do we need to cover this in the docs? To me it feels like a given; if it was not the case before, this could be covered in the changelog.

web_poet/page_inputs/client.py
```diff
@@ -165,8 +178,17 @@ async def execute(
...
+        if self.return_only_saved_responses:
+            for fp, (_, saved_response) in self._saved_responses.items():
+                if request_fingerprint(request) == fp:
+                    return saved_response
+            raise NoSavedHttpResponse(request=request)
+
         response = await self._request_downloader(request)
         self._handle_status(response, request, allow_status=allow_status)
+        if self.save_responses:
+            # TODO: copy()?
```
Gallaecio, line 190: I am also not sure copying is needed. I would think not.

web_poet/page_inputs/http.py
```diff
@@ -300,3 +302,14 @@ def _auto_detect_fun(self, body: bytes) -> Optional[str]:
...
+
+
+def request_fingerprint(req: HttpRequest) -> str:
+    """Return the fingerprint of the request."""
+    fp = sha1()
+    fp.update(req.method.encode() + b"\n")
+    fp.update(canonicalize_url(str(req.url)).encode() + b"\n")
+    for name, value in sorted(req.headers.items()):
+        fp.update(f"{name.title()}:{value}\n".encode())
+    fp.update(req.body)
```
Gallaecio, line 314: This seems problematic. Imagine a body that starts with what looks like a header.

Maybe a way to go is to add an extra empty line, like the HTTP protocol does, to separate headers from body.

```suggestion
        fp.update(f"{name.title()}:{value}\n".encode())
    fp.update(b"\n" + req.body)
```

docs/page-objects/testing.rst
wRAR: @kmike suggested adding it :)

kmike: We had this question asked, so maybe it's not given for the users. 
I'm fine both with keeping and dropping it.

web_poet/exceptions/http.py
wRAR: Yeah, I wasn't sure if this counts as HttpError or not. 

I don't know about AssertionError, will it have some consequences, e.g. special handling by pytest?

docs/page-objects/testing.rst
Gallaecio: > We had this question asked.

I guess that is motivation enough. :+1:

tests/test_testing.py
```diff
@@ -158,3 +158,70 @@ def test_pytest_frozen_time_tz_windows_pass(pytester, book_list_html_response) -
...
+        (request2, response2),
+    ]
+    client = HttpClient(responses=responses)
+
+    base_dir = pytester.path / "fixtures" / get_fq_class_name(ClientPage)
+    item = {
+        "foo": "bar",
+        "additional": ["body1", "body2"],
+    }
+    Fixture.save(base_dir, inputs=[book_list_html_response, client], item=item)
```
kmike, line 193: I think you can simplify it a bit using _save_fixture utility function added in #129.

web_poet/exceptions/http.py
wRAR: Should it also be moved to exceptions/core.py?

Gallaecio (approved):

web_poet/exceptions/http.py
kmike: We also have testing/exceptions.py, though I'm not sure it's good that they're not in exceptions/testing.py.

kmike: What do you think about moving testing/exceptions.py to exceptions/testing.py, and then moving NoSavedHttpResponse there?

kmike: These exceptions are not the ones which users should be catching though; that's why I've added them to testing module, not to top-level web_poet.exceptions. `exceptions/_testing.py` is also an option.

tests/test_testing.py
kmike: Why is using _save_fixture not possible here?

kmike: ah, I see :)

docs/page-objects/testing.rst
```diff
@@ -217,3 +217,24 @@ Please also check the official Git LFS documentation for more information.
...
+
+If the page object uses the :class:`~.HttpClient` dependency to make
+:ref:`additional requests <additional-requests>`, the generated fixtures will
+contain these requests and their responses. When the test runs,
+:class:`~.HttpClient` will return the saved responses without doing actual
+requests.
+
+Currently requests are compared by their URL, method, headers and body, so if a
+page object makes requests that differ between runs, the test won't be able to
+find a saved response and will fail.
```
BurnzZ, line 232: It might also be noteworthy to mention that these are saved with the following filenames:

*   `HttpClient-n-HttpRequest.info.json`
*   `HttpClient-n-HttpResponse.body.html`

Or maybe show them in the GNU tree output in the fixtures section above.

web_poet/exceptions/http.py
wRAR: Yeah, so it makes sense to leave `testing/exceptions.py` as is, then what should we do with `NoSavedHttpResponse`?

kmike: Options:

1. Move it to `testing/exceptions.py`.
2. Keep it as-is.
3. Move to core.py.

Either way, I think it shouldn't inherit from HttpError, as it'd lead to issues. I'm fine with any of these locations, as soon as it's not a subclass of HttpError.

We shouldn't expose this exception in `__init__.py`, because this exception is not supposed to be caught by web-poet users' code.

wRAR: Moving it to `testing/exceptions.py` causes some circular import between `web_poet.testing.fixture` and `web_poet`.

On the other hand, keeping it in `http` or moving to `core` means we need to change the current `from .core import *; from .http import *` code if we don't want to import this one there :)

kmike: > On the other hand, keeping it in http or moving to core means we need to change the current from .core import *; from .http import * code if we don't want to import this one there :)

Right; using `__all__` is also an option.

web_poet/page_inputs/client.py
```diff
@@ -22,6 +25,18 @@
 _StatusList = Union[str, int, List[Union[str, int]]]
 
 
+@dataclass
+class SavedResponseData:
```
wRAR, line 29: I made this a real class when I still thought we will need to add an exception field here, now we can make it a simple tuple again but I like named fields.

kmike: Would it make sense to call it _SavedResponseData?

wRAR: We import it in other parts of web-poet, though I don't think other projects would need to import it, so I think yes?

kmike: I see. Yes, it seems making it private is better, as it's not something users should use.

kmike (approved):
👍

BurnzZ (approved):

PR #129, opened 2023-02-13
Improve test reporting

kmike:
* By default, each field is a separate test
* There is an option to compare complete items
* Custom error reporting is implemented (no more 20 page tracebacks)

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -43,6 +43,7 @@ per-file-ignores =
     web_poet/page_inputs/__init__.py:F401,F403
     web_poet/serialization/__init__.py:F401,F403
     web_poet/testing/__init__.py:F401,F403
+    web_poet/testing/pytest.py:D102
     tests/po_lib_to_return/__init__.py:D102
 
     # the suggestion makes the code worse
--- a/web_poet/testing/exceptions.py
+++ b/web_poet/testing/exceptions.py
@@ -1,0 +1,14 @@
+class FieldMissing(AssertionError):
+    pass
+
+
+class FieldValueError(AssertionError):
+    pass
+
+
+class ItemValueError(AssertionError):
+    pass
+
+
+class FieldsUnexpected(AssertionError):
+    pass
--- a/web_poet/testing/pytest.py
+++ b/web_poet/testing/pytest.py
@@ -1,10 +1,16 @@
 import operator
 from pathlib import Path
-from typing import Any, Iterable, List, Optional, Set, Union
+from typing import Iterable, List, Optional, Set, Union
 
 import pytest
 
+from web_poet.testing.exceptions import (
+    FieldsUnexpected,
+    FieldValueError,
+    ItemValueError,
+)
 from web_poet.testing.fixture import OUTPUT_FILE_NAME, Fixture
+from web_poet.testing.utils import comparison_error_message
 
 # https://github.com/pytest-dev/pytest/discussions/10261
 _version_tuple = getattr(pytest, "version_tuple", None)
@@ -15,16 +21,16 @@ class WebPoetFile(pytest.File):
     """Represents a directory containing test subdirectories for one Page Object."""
 
     @staticmethod
-    def sorted(items: List["WebPoetItem"]) -> List["WebPoetItem"]:
+    def sorted(items: List["WebPoetCollector"]) -> List["WebPoetCollector"]:
         """Sort the test list by the test name."""
         return sorted(items, key=operator.attrgetter("name"))
 
     def collect(self) -> Iterable[Union[pytest.Item, pytest.Collector]]:  # noqa: D102
-        result: List[WebPoetItem] = []
+        result: List[WebPoetCollector] = []
         path = self.path if _new_pytest else Path(self.fspath)
         for entry in path.iterdir():
             if entry.is_dir():
-                item: WebPoetItem = _get_item(
+                item: WebPoetCollector = _get_collector(
                     self,
                     name=entry.name,
                     path=entry,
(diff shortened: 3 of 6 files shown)


.flake8
```diff
@@ -43,6 +43,7 @@ per-file-ignores =
     web_poet/page_inputs/__init__.py:F401,F403
     web_poet/serialization/__init__.py:F401,F403
     web_poet/testing/__init__.py:F401,F403
+    web_poet/testing/pytest.py:D102
```
kmike, line 46: most public methods defined in this module are pytest hooks

kmike:
FTR, Ubuntu tests seems to pass, but we seem to face https://github.com/codecov/codecov-action/issues/557, so jobs fail at trying to upload the coverage information.

kmike:
Windows tests fail with some time freezing error, which doesn't seem related to this PR, which is weird! (cc @wRAR)

.github/workflows/tests-ubuntu.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       uses: codecov/codecov-action@v3
       if: ${{ success() }}
       with:
-        fail_ci_if_error: true
+        fail_ci_if_error: false
```
kmike, line 45: I disabled it to fix the failing jobs, which were failing at codecov step. And indeed, the jobs passed after this change. But after this change the codecov step was successful (it seems there are some random failures), so I don't think they passed because of the change :)

web_poet/testing/pytest.py
```diff
@@ -34,38 +46,106 @@ def collect(self) -> Iterable[Union[pytest.Item, pytest.Collector]]:  # noqa: D1
...
+        if isinstance(excinfo.value, FieldsUnexpected):
+            fields = excinfo.value.args[0]
+            return f"The item contains unexpected fields: \n{self._format_extra_fields(fields)}"
+        else:
+            return super().repr_failure(excinfo, style)
+
+    def _format_extra_fields(self, extra_fields):
+        lines = []
+        for field, value in extra_fields.items():
+            lines.append(f" * {field} == {value!r}")
```
BurnzZ, line 122:
```suggestion
            lines.append(f" * {field} = {value!r}")
```

web_poet/testing/exceptions.py
```diff
@@ -0,0 +1,14 @@
+class FieldMissing(AssertionError):
```
BurnzZ, line 1: This one curently returns extra stracktraces unlike the other ones.

BurnzZ (commented):
I tried this out quickly on a Scrapy project and it does makes testing more clear. :tada:

wRAR:
Regarding the Windows test failures, the error of `test_pytest_frozen_time_tz_windows_fail` makes sense: we create a date with the offset of -7.5 seconds instead of hours, which is the underlying problem here. We now see this error just because the expected number of failures in this test wasn't updated like it was in other tests (`{'failed': 27} != {'failed': 1}`), but thanks to that we now know that the expected failure was not because of running the test in a wrong timezone but because of this ParserError.

So first I would replace -7.5 with -7.5*3600 and would check that the printed test failures are correct, and then would replace `{"failed": 1}` with `{"failed": 27}`.

web_poet/testing/pytest.py
kmike: Should it be without whitespace around `=` then?

web_poet/testing/exceptions.py
kmike: Fixed in https://github.com/scrapinghub/web-poet/pull/129/commits/3bc857ad905419d573cbfbc0afdd6435ca2c2668

web_poet/testing/pytest.py
```diff
@@ -34,38 +47,109 @@ def collect(self) -> Iterable[Union[pytest.Item, pytest.Collector]]:  # noqa: D1
...
+            got, expected = excinfo.value.args
+            return comparison_error_message(
+                config=self.config,
+                op="==",
+                expected=expected,
+                got=got,
+                prefix="The output doesn't match.",
+            )
+        else:
+            return super().repr_failure(excinfo, style)
```
kmike, line 103: Lines like this are the only ones which are missing the coverage. I think that's fine, because this code doesn't run normally; it's a safety measure, to get better output in case something strange happens (e.g. a bug in our plugin).

web_poet/testing/exceptions.py
```diff
@@ -0,0 +1,14 @@
...
+class FieldValueError(AssertionError):
+    pass
+
+
+class ItemValueError(AssertionError):
+    pass
+
+
+class FieldsUnexpected(AssertionError):
+    pass
```
Gallaecio, line 14: I wonder if we should rename `FieldMissing` and `FieldsUnexpected` to `MissingFieldError` and `UnexpectedFieldsError` for consistency.

Alternatively, rename all to something without an `Error` suffix but still consistent with each other (i.e. ending with a noun), e.g. `MissingField`, `UnexpectedFieldValue`, `UnexpectedItemValue`, `UnexpectedFields`.

Gallaecio (approved):

web_poet/testing/exceptions.py
kmike: My logic was to use Field.. prefixes for field errors and Item prefix for item error, and omit "Error" because they'are all in .exceptions module.

kmike: though I haven't omitted "Error", so probably the logic was different :)

kmike: Anyways, happy to proceed with any naming.

kmike: Would this be better? 

* FieldMissing
* FieldValueIncorrect
* ItemValueIncorrect
* FieldsUnexpected

kmike: The reason to use Field vs Item prefixes was that there is a very clear separation between those exception types: Item.. can only be raised when `--web-poet-test-per-item` is passed, and Field.. can only be raised if it's not passed.

Gallaecio: Just to be clear, I don’t have a strong opinion against the proposed names. https://github.com/scrapinghub/web-poet/pull/129#discussion_r1106367476 would also work for me.

wRAR (approved):

kmike:
Merging it, thanks @wRAR @BurnzZ @Gallaecio!

PR #130, opened 2023-02-15
support non-ascii output in fixtures

kmike:
savefixture command currently produces output like

>     "currencyRaw": "\u20ac",

instead of 

>     "currencyRaw": "€",

Is there a particular reason ensure_ascii is set to True?


wRAR:
Just a mistake.

wRAR (approved):

PR #132, opened 2023-02-17
python -m web_poet.testing rerun

kmike:
A command to print the current page object output, given the inputs stored in a fixture. It enables this very simplistic workflow: 

1. change a page object so that its output is correct (e.g. implement extraction for a field);
2. `python -m web_poet.testing rerun fixtures/pages.mywebsite.MyProductPage/test-1`;
3. copy relevant parts of the output, paste to the output.json file.

I think that even if we implement something which supports the TDD workflow better, a simple command to re-run the page object can still be useful.

TODO:

* [x] tests
* [x] docs

### DIFF
--- a/tests/test_testing.py
+++ b/tests/test_testing.py
@@ -349,14 +349,32 @@ def test_httpclient_exception(pytester, book_list_html_response) -> None:
     result.assert_outcomes(passed=3)
 
 
+class MyItemPage3(WebPage):
+    async def to_item(self) -> dict:  # noqa: D102
+        return {"foo": "bar", "egg": "spam", "hello": "world"}
+
+
 def test_cli_rerun(pytester, book_list_html_response, capsys) -> None:
     fixture = _save_fixture(
         pytester,
-        page_cls=MyItemPage,
+        page_cls=MyItemPage3,
         page_inputs=[book_list_html_response],
-        expected={"foo": "bar2"},
+        expected={"foo": "bar2", "egg": "spam", "hello": "world"},
     )
     cli_main(["rerun", str(fixture.path)])
     captured = capsys.readouterr()
     assert not captured.err
-    assert json.loads(captured.out) == {"foo": "bar"}
+    assert json.loads(captured.out) == {"foo": "bar", "egg": "spam", "hello": "world"}
+
+
+def test_cli_rerun_fields(pytester, book_list_html_response, capsys) -> None:
+    fixture = _save_fixture(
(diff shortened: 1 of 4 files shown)


tests/test_testing.py
```diff
@@ -368,3 +369,34 @@ def test_httpclient_exception(pytester, book_list_html_response) -> None:
...
+
+
+def test_cli_rerun(pytester, book_list_html_response, capsys) -> None:
+    fixture = _save_fixture(
+        pytester,
+        page_cls=MyItemPage3,
+        page_inputs=[book_list_html_response],
+        expected={"foo": "bar2", "egg": "spam", "hello": "world"},
+    )
+    cli_main(["rerun", str(fixture.path)])
```
kmike, line 386: Initially I wanted to use subprocess, but had hard time making the tests/test_testing.MyItemPage importable. This way it works.

web_poet/testing/__main__.py
```diff
@@ -0,0 +1,43 @@
...
+
+from web_poet.testing import Fixture
+
+
+def rerun(args):
+    fixture = Fixture(Path(args.fixture_path))
+    item = fixture.get_output()
+    if args.fields:
+        fields = args.fields.split(",")
+        item = {field: item[field] for field in fields}
```
BurnzZ, line 12: I think we can improve the error message here when users input a non-existing field. Or maybe ignore/supress the erorr?

kmike:
Also not sure about `rerun` name, but haven't came up with a better name :( 

The good thing is that it's not a "scrapy savefixture rerun" command, but an unrelated `python -m web_poet.testing rerun`, so there is a bit less opportunity for confusion.

BurnzZ (approved):

web_poet/testing/__main__.py
kmike: I was hoping nobody will notice it, and it'd be ok to just have a KeyError in the output :P

kmike: Done in https://github.com/scrapinghub/web-poet/pull/132/commits/c2b18065bd7837f0c3ea27257084a8b511cbe241.

kmike:
Merging it, thanks @Gallaecio @BurnzZ!

kmike:
Ouch, tests start to fail :)

kmike:
Fixed in https://github.com/scrapinghub/web-poet/pull/137, it was unrelated.

PR #133, opened 2023-02-17
Fix GH-131

### LINKED ISSUE #131: All per-field tests fail if one of the fields raise an exception
Currently tests work like this:

1. to_item is called
2. a test is generated and executed for each field

But if to_item raises an exception, then no fields are available; a test is still generated for each field, and they all fail, each with a huge traceback.

To make things worse, in case of an exception to_item output is not memoized, so to_item is called again for each field, making it all slow.

I think that calling to_item (as opposed to accessing individual fields) is still the way to go, because it gives the final output, and that's what we should be testing.

So, a proposal on how to solve it:

1. Add one more generted test, which cheks that to_item doesn't raise an exception. It'd be passing in most cases. If it fails, it'd contain a full traceback. 
2. If this "no exceptions" test fails, per-field tests should be skipped. This way we won't show a huge traceback per each field. 
3. It should be possible to still run a per-field skipped test explicitly. The use case: user works on a particular field, and an exception happens in this field. When this test is run, we sould show the full traceback.

We should aslo make sure that we don't call to_item multuple times if it fails, but it can be a separate ticket.

kmike:
This fixes #131, following the approach outlined there.

### DIFF
--- a/docs/page-objects/testing.rst
+++ b/docs/page-objects/testing.rst
@@ -107,7 +107,7 @@ The provided ``pytest`` plugin is automatically registered when ``web-poet`` is
 installed, and running ``python -m pytest`` in a directory containing fixtures
 will discover them and run tests for them.
 
-By default, the plugin generates
+By default, the plugin generates:
 
 * a test which checks that ``to_item()`` doesn't raise an exception
   (i.e. it can be executed),
--- a/docs/page-objects/testing.rst
+++ b/docs/page-objects/testing.rst
@@ -107,20 +107,28 @@ The provided ``pytest`` plugin is automatically registered when ``web-poet`` is
 installed, and running ``python -m pytest`` in a directory containing fixtures
 will discover them and run tests for them.
 
-By default, the plugin generates a test per each output attribute of the item,
-and an additional test to check that there are no extra attributes in the output.
+By default, the plugin generates
+
+* a test which checks that ``to_item()`` doesn't raise an exception
+  (i.e. it can be executed),
+* a test per each output attribute of the item,
+* an additional test to check that there are no extra attributes in the output.
+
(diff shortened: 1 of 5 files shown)


docs/page-objects/testing.rst
```diff
@@ -107,20 +107,28 @@ The provided ``pytest`` plugin is automatically registered when ``web-poet`` is
 installed, and running ``python -m pytest`` in a directory containing fixtures
 will discover them and run tests for them.
 
-By default, the plugin generates a test per each output attribute of the item,
-and an additional test to check that there are no extra attributes in the output.
+By default, the plugin generates
```
Gallaecio, line 110:
```suggestion
By default, the plugin generates:
```

web_poet/testing/pytest.py
```diff
@@ -124,12 +132,29 @@ def _format_extra_fields(self, extra_fields):
...
     def __init__(self, *, field_name: str, **kwargs) -> None:
         super().__init__(**kwargs)
         self.field_name = field_name
 
     def runtest(self) -> None:
+        if self.fixture.to_item_raised():
+            raise pytest.skip(
+                f"Skipping a test for item.{self.field_name} "
+                f"because to_item raised an exception"
+            )
```
Gallaecio, line 157: I am unsure about this one. Shouldn’t checks for fields that do not raise an exception still pass?

kmike: We can't get the final values for other fields if one of the fields fail. In this case to_item raises an error, so there are no values for other fields. 

I think that using page.field output directly is not the way to go, because we want to support page objects which don't use field decorator.

Gallaecio: > we want to support page objects which don't use field decorator

:+1:

Gallaecio (approved):

PR #134, opened 2023-02-17
Add serialization for dict

wRAR:
This is needed to serialize PageParams.

### DIFF
--- a/web_poet/serialization/functions.py
+++ b/web_poet/serialization/functions.py
@@ -140,3 +140,16 @@ def _deserialize_HttpClient(
 
 
 register_serialization(_serialize_HttpClient, _deserialize_HttpClient)
+
+
+def _serialize_dict(o: dict) -> SerializedLeafData:
+    return {
+        "json": json.dumps(o, ensure_ascii=False, sort_keys=True, indent=2).encode()
+    }
+
+
+def _deserialize_dict(cls: Type[dict], data: SerializedLeafData) -> dict:
+    return cls(json.loads(data["json"]))
+
+
+register_serialization(_serialize_dict, _deserialize_dict)
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -3,7 +3,7 @@ from typing import Type
 import attrs
 import pytest
 
-from web_poet import HttpResponse, HttpResponseBody, ResponseUrl, WebPage
+from web_poet import HttpResponse, HttpResponseBody, PageParams, ResponseUrl, WebPage
 from web_poet.page_inputs.url import _Url
 from web_poet.serialization import (
     SerializedDataFileStorage,
@@ -61,11 +61,13 @@ def test_serialization(book_list_html_response) -> None:
     @attrs.define
     class MyWebPage(WebPage):
         url: ResponseUrl
+        params: PageParams
 
     url_str = "http://books.toscrape.com/index.html"
     url = ResponseUrl(url_str)
+    page_params = PageParams(foo="bar")
 
-    serialized_deps = serialize([book_list_html_response, url])
+    serialized_deps = serialize([book_list_html_response, url, page_params])
     info_json = f"""{{
   "_encoding": "utf-8",
   "headers": [],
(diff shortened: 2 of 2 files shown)


web_poet/serialization/functions.py
```diff
@@ -140,3 +140,16 @@ def _deserialize_HttpClient(
 
 
 register_serialization(_serialize_HttpClient, _deserialize_HttpClient)
+
+
+def _serialize_dict(o: dict) -> SerializedLeafData:
```
kmike, line 145: I think earier we went away from providing serialize functions for native classes like dict or str. Would it make sense to have it specific to PageParams?

wRAR: We went away from providing them "just in case", but here it's useful for PageParams and I don't think a specialized function is better?

kmike: I guess my motivation was this: we don't want to accidentally introduce raw dict as a dependency, so that's good if it's explicitly unsupported, and that we don't suggest that it might be supported. That's why if it is the same amount of code, I prefer using PageParams explicitly, not having support for dict. 

But no strong opinon :)

kmike (approved):
👍

wRAR:
The problem with empty fields is that if you used `ItemAdapter.ADAPTER_CLASSES.appendleft(ZyteItemAdapter)` when generating the test you must do the same when running it, e.g. by adding it to conftest.py.

kmike:
> The problem with empty fields is that if you used ItemAdapter.ADAPTER_CLASSES.appendleft(ZyteItemAdapter) when generating the test you must do the same when running it, e.g. by adding it to conftest.py.

So, what's the solution? Is it a change to the docs, or should we do something else?

web_poet/serialization/api.py
```diff
@@ -182,5 +182,12 @@ def deserialize(cls: Type[InjectableT], data: SerializedData) -> InjectableT:
...
-    plan = andi.plan(cls, is_injectable=is_injectable, externally_provided=deps.keys())
+    externally_provided = deps.keys()
+    plan = andi.plan(
+        cls, is_injectable=is_injectable, externally_provided=externally_provided
+    )
+    for fn_or_cls, kwargs_spec in plan[:-1]:
+        if fn_or_cls in externally_provided:
+            continue
+        deps[fn_or_cls] = fn_or_cls(**kwargs_spec.kwargs(deps))
     return cls(**plan.final_kwargs(deps))
```
kmike, line 193: Just curious, can you do it like this? I think it'd be similar to what's in andi docs:

```suggestion
    for fn_or_cls, kwargs_spec in plan:
        if fn_or_cls in externally_provided:
            continue
        deps[fn_or_cls] = fn_or_cls(**kwargs_spec.kwargs(deps))
    return deps[cls]
```

kmike:
Thanks @wRAR! I'm merging it, because the fixes which are in this PR look complete.

There is still a question about ItemAdapter and empty item fields, but it doesn't have to be solved in the same PR.

wRAR:
> So, what's the solution? Is it a change to the docs, or should we do something else?

I think it should be a change for the docs that tell you to do enable `ZyteItemAdapter`, like https://zyte-common-items.readthedocs.io/en/latest/setup.html#configuration and similar, though I don't know what is a good place to talk about this (probably not web-poet and not zyte-common-items, though maybe we could mention it in web-poet?).

web_poet/serialization/api.py
wRAR: Yeah, looks like it works :)

kmike:
> I think it should be a change for the docs that tell you to do enable ZyteItemAdapter, like https://zyte-common-items.readthedocs.io/en/latest/setup.html#configuration and similar, though I don't know what is a good place to talk about this (probably not web-poet and not zyte-common-items, though maybe we could mention it in web-poet?).

@wRAR Could you please open a ticket in web-poet issue tracker, to have this discussion in a single place?
There might be more to it. For example, it's quite nice when savefixture generates an empty item with field values set to null, instead of dropping these fields. It allows to easily see what fields one needs to implement, and allows to set field values more easily.

PR #138, opened 2023-02-21
Introduce `SelectFields` — alternative

BurnzZ:
Build on top of https://github.com/scrapinghub/web-poet/pull/120.

This is related to #115 which implements [approach 3](https://github.com/scrapinghub/web-poet/issues/115#issuecomment-1408526596).

This is an alternative to https://github.com/scrapinghub/web-poet/pull/118 wherein:

- `SelectFields` is now an _optional_ dependency to `ItemPage`. 
    - ~However, this requires the dependencies of `ItemPage` and `WebPage` to be keyword-only.~ **EDIT:** solved in https://github.com/scrapinghub/web-poet/commit/fb7b7bd70ccd6344f309baf831fc3024b09b8596
- A new `item_from_select_fields()` utility function has been created to fully handle [Scenario A, B, and C](https://github.com/scrapinghub/web-poet/issues/115#issue-1550785899).

**TODO**

- [x] ensure parity and tests with `skip_nonitem_fields` after https://github.com/scrapinghub/web-poet/issues/141 is addressed
- [ ] docs
- [ ] changelog

Out of scope for this PR:

- exploring `on_unknown_field` in `SelectFields`

### DIFF
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -1,6 +1,7 @@
 import asyncio
 import random
-from typing import Optional
+from collections import defaultdict
+from typing import DefaultDict, Optional
 
 import attrs
 import pytest
@@ -23,9 +24,12 @@ from web_poet import (
     HttpResponse,
     Injectable,
     ItemPage,
+    SelectFields,
+    WebPage,
     field,
     item_from_fields,
     item_from_fields_sync,
+    item_from_select_fields,
 )
 from web_poet.fields import FieldInfo, get_fields_dict
 
@@ -439,7 +443,6 @@ def test_field_with_other_decorators() -> None:
 
 @pytest.mark.asyncio
 async def test_field_with_handle_urls() -> None:
-
     page = ProductPage()
     assert page.name == "name"
     assert page.price == 12.99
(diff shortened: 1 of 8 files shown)


BurnzZ:
The use case for this PR is ready for review, specifically the `tests/test_fields.py`.

tests/test_pages.py
```diff
@@ -41,7 +41,7 @@ def to_item(self) -> dict:  # type: ignore
                 "title": self.css("title::text").get("").strip(),
             }
 
-    page_object = MyWebPage(book_list_html_response)
+    page_object = MyWebPage(response=book_list_html_response)
```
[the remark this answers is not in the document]
kmike: Making it kw_only makes total sense to me, we should have done it from the start.

Does it mean though that user-defined dependencies also need to be kw-only now? It could mean that we need to update a examples in the docs; it might also break code.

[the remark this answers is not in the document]
BurnzZ: I've got a way around it by using an `attrs.field` converter in ItemPage for SelectPage. In this way, we don't have to break any code for now. fb7b7bd

tests/test_fields.py
```diff
@@ -630,3 +633,406 @@ def x(self) -> int:
...
+        r"__init__\(\) missing 1 required positional argument: 'x'"
+    )
+    response = HttpResponse("https://example.com", b"")
+
+    # When SelectFields isn't set
+    page = BigUnreliablePage(response=response)
+    assert page.fields_to_extract == ["x"]
+    assert await page.to_item() == BigItem(x=1, y=2, z=3)
+    assert page.call_counter == {"x": 1, "z": 1}
+    assert await item_from_select_fields(page) == BigItem(x=1, y=None, z=None)
```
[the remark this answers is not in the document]
Gallaecio: > If it were to simply call `.to_item()` and drop "z", that _might_ be inefficient since "z" could use additional requests.

> The best route to take here is the users should create page objects with efficient `.to_item()` methods

Since the efficiency of dropping `z` after `to_item()` is a user problem, one that users should address, should we still drop it?

```python
assert await item_from_select_fields(page) == BigItem(x=1, y=2, z=None)
```

*Maybe* also log a warning in scenarios where we drop a field that as not `None`?

Gallaecio (approved):

tests/test_fields.py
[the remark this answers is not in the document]
BurnzZ: As discussed last Thursday, we'll need to update `item_from_select_fields()` to always call `.to_item()` and then dropping the fields. 68d2cc2 In this way, the page obj developer needs to make sure that `.to_item()` is efficient (i.e. calling `super().to_item()` or `item_from_fields()` to utilize the selected fields.) Moreover, we'll use `fields_to_ignore` instead of `fields_to_extract`.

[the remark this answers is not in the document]
BurnzZ: As discussed last Thursday, we'll need to update \`item\_from\_select\_fields()\` to always call \`.to\_item()\` and then dropping the fields. d09bf66 In this way, the page obj developer needs to make sure that \`.to\_item()\` is efficient (i.e. calling \`super().to\_item()\` or \`item\_from\_fields()\` to utilize the selected fields.) Moreover, we'll use \`fields\_to\_ignore\` instead of \`fields\_to\_extract\`.

web_poet/fields.py
```diff
@@ -216,3 +225,86 @@ def _without_unsupported_field_names(
     if item_field_names is None:  # item_cls doesn't define field names upfront
         return field_names[:]
     return list(set(field_names) & set(item_field_names))
+
+
+@attrs.define
+class SelectFields:
+    """This is used as a dependency in :class:`~.ItemPage` to control which
+    fields to populate its returned item class.
```
kmike, line 233: I wonder if we should rename it, and maybe even move (though not sure), because it's not necessarily about web-poet fields used to populate the item class. It's about the item attributes user wants to get in the output. Page object may not use fields at all.

web_poet/fields.py
```diff
@@ -216,3 +225,86 @@ def _without_unsupported_field_names(
...
+
+    You can also use this to enable some fields that were disabled by default
+    via the ``@field(disabled=True)`` decorator.
+
+    Some usage examples:
+
+    * ``SelectFields({"name": True})`` — select one field
+    * ``SelectFields({"name": False})`` — unselect one field
+    * ``SelectFields({"*": True})`` — select all fields
```
kmike, line 242: Let's add a note about disabled fields: are they selected this way or not? E.g. "select all fields, including disabled"

web_poet/fields.py
```diff
@@ -216,3 +225,86 @@ def _without_unsupported_field_names(
...
+    * ``SelectFields({"*": True, "name": False})`` — select all fields but one
+    * ``SelectFields({"*": False, "name": True})`` — unselect all fields but one
+
+    """
+
+    #: Fields that the page object would use to populate the
```
kmike, line 248: It's more about attributes of the item which user wants to get.

web_poet/fields.py
```diff
@@ -216,3 +225,86 @@ def _without_unsupported_field_names(
...
+    #: :meth:`~.Returns.item_cls` it returns. It's a mapping of field names to
```
kmike, line 249: A mapping of item attribute names

web_poet/fields.py
Gallaecio: Brainstorm: `SelectAttributes`, `SelectAttrs`, `SelectKeys`, `AttributeFilter`, `AttrFilter`, `KeyFilter`, `ItemFilter`, `FilterAttributes`, `FilterAttrs`, `FilterKeys`, `FilterItem`.

BurnzZ: I'm inclined to have the word _"item"_ in there for clarity. `ItemFilter` could mean that the page object supports different types of items and a filter is needed to choose the right one. Somehow, we may want to also include the word _"attributes"_ for better clarity.

How about `ItemAttributes`?

web_poet/fields.py
BurnzZ: Speaking of `disabled` fields, are we good to go with this terminology? or continue considering renaming it as well as previously explored in https://github.com/scrapinghub/web-poet/pull/120#discussion_r1090424890.

web_poet/fields.py
```diff
@@ -2,14 +2,17 @@
 ``web_poet.fields`` is a module with helpers for putting extraction logic
 into separate Page Object methods / properties.
 """
+from __future__ import annotations
```
BurnzZ, line 5: Added to get around the cyclic imports so we could have this annotation: `page: web_poet.ItemPage`

PR #140, opened 2023-02-22
allow GitHub actions CI to run on PRs not merging with master

BurnzZ:
**Motivation**: A PR like #138 doesn't run its CI since it's not merging with master


kmike:
👍

PR #142, opened 2023-02-22
fix `skip_nonitem_fields=True` doesn't work when the page object is an attrs class

### LINKED ISSUE #141: `skip_nonitem_fields=True` doesn't work when the page object is an attrs class
Currently, this works fine:

```python
import attrs
from web_poet import HttpResponse, Returns, ItemPage, field

@attrs.define
class BigItem:
    x: int
    y: int

class BigPage(ItemPage[BigItem]):
    @field
    def x(self):
        return 1

    @field
    def y(self):
        return 2

@attrs.define
class SmallItem:
    x: int

class SmallXPage(BigPage, Returns[SmallItem], skip_nonitem_fields=True):
    pass

page = SmallXPage()
item = await page.to_item()

print(page._skip_nonitem_fields)  # True
print(item)  # SmallItem(x=1)
```

However, if we define an attrs class to have some page dependencies, it doesn't work:

```python
from web_poet import PageParams

@attrs.define
class SmallPage(BigPage, Returns[SmallItem], skip_nonitem_fields=True):
    params: PageParams

page = SmallPage(params=PageParams())
print(page._skip_nonitem_fields)  # False
item = await page.to_item()  # TypeError: __init__() got an unexpected keyword argument 'y'
```

From the examples above, this stems from `page._skip_nonitem_fields` being set to `False` when the page object is defined as an attrs class.

BurnzZ:
Resolves #141

### DIFF
--- a/tests/test_pages.py
+++ b/tests/test_pages.py
@@ -3,7 +3,7 @@ from typing import Optional
 import attrs
 import pytest
 
-from web_poet import HttpResponse, field
+from web_poet import HttpResponse, PageParams, field
 from web_poet.pages import (
     Injectable,
     ItemPage,
@@ -199,17 +199,23 @@ async def test_item_page_change_item_type_remove_fields() -> None:
     class Subclass(BasePage, Returns[Item], skip_nonitem_fields=True):
         pass
 
-    page = Subclass()
-    assert page.item_cls is Item
-    item = await page.to_item()
-    assert isinstance(item, Item)
-    assert item == Item(name="hello")
+    # Same as above but an attrs class with dependency.
+    # See: https://github.com/scrapinghub/web-poet/issues/141
+    @attrs.define
+    class SubclassWithDep(BasePage, Returns[Item], skip_nonitem_fields=True):
+        params: PageParams
 
-    # Item only contains "name", but not "price", but "price" should be passed
-    class SubclassStrict(BasePage, Returns[Item]):
-        pass
+    for page in [Subclass(), SubclassWithDep(params=PageParams())]:
+        assert page.item_cls is Item
+        item = await page.to_item()
+        assert isinstance(item, Item)
+        assert item == Item(name="hello")
 
-    page2 = SubclassStrict()
-    assert page2.item_cls is Item
-    with pytest.raises(TypeError, match="unexpected keyword argument 'price'"):
-        await page2.to_item()
+        # Item only contains "name", but not "price", but "price" should be passed
+        class SubclassStrict(BasePage, Returns[Item]):
+            pass
+
+        page2 = SubclassStrict()
+        assert page2.item_cls is Item
+        with pytest.raises(TypeError, match="unexpected keyword argument 'price'"):
+            await page2.to_item()
--- a/tests/test_pages.py
+++ b/tests/test_pages.py
@@ -211,11 +211,11 @@ async def test_item_page_change_item_type_remove_fields() -> None:
         assert isinstance(item, Item)
         assert item == Item(name="hello")
 
-        # Item only contains "name", but not "price", but "price" should be passed
-        class SubclassStrict(BasePage, Returns[Item]):
-            pass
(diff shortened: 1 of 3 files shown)


web_poet/pages.py
```diff
@@ -59,7 +59,12 @@ class ItemPage(Injectable, Returns[ItemT]):
 
     def __init_subclass__(cls, skip_nonitem_fields: bool = False, **kwargs):
         super().__init_subclass__(**kwargs)
-        cls._skip_nonitem_fields = skip_nonitem_fields
+
+        if getattr(cls, "_skip_nonitem_fields", None) is True:
+            # See: https://github.com/scrapinghub/web-poet/issues/141
+            cls._skip_nonitem_fields = True
```
BurnzZ, line 65: Attrs with `slot=True` causes this to run twice where the second call would always have `skip_nonitem_fields=False`. This workaround prevents overriding the value of `True` when it was intentionally passed earlier.

wRAR (approved):

tests/test_pages.py
```diff
@@ -199,17 +199,23 @@ def price(self):
...
-    with pytest.raises(TypeError, match="unexpected keyword argument 'price'"):
-        await page2.to_item()
+        # Item only contains "name", but not "price", but "price" should be passed
+        class SubclassStrict(BasePage, Returns[Item]):
+            pass
+
+        page2 = SubclassStrict()
+        assert page2.item_cls is Item
+        with pytest.raises(TypeError, match="unexpected keyword argument 'price'"):
+            await page2.to_item()
```
Gallaecio, line 221: Shouldn’t this remain outside the loop?

Gallaecio (approved):

tests/test_pages.py
BurnzZ: Great catch!

web_poet/pages.py
kmike: That's a super weird looking code: we're checking if value is True, and if it's True, we set it to True :) Magic!

Does it work when subclass sets skip_nonitem_fields to False, not to True, while the base class has skip_nonitem_fields=True? Could you please add a test for this?

BurnzZ: Good catch! It wasn't working before. Addressed in f829a95.

web_poet/pages.py
```diff
@@ -55,11 +55,22 @@ class ItemPage(Injectable, Returns[ItemT]):
...
+        )
 
     def __init_subclass__(cls, skip_nonitem_fields: bool = False, **kwargs):
         super().__init_subclass__(**kwargs)
-        cls._skip_nonitem_fields = skip_nonitem_fields
+
+        # See: https://github.com/scrapinghub/web-poet/issues/141
+        if not getattr(cls, f"_{cls.__name__}__skip_nonitem_fields_set", False):
+            setattr(cls, f"_{cls.__name__}__skip_nonitem_fields", skip_nonitem_fields)
+            setattr(cls, f"_{cls.__name__}__skip_nonitem_fields_set", True)
```
BurnzZ, line 73: Needed to do manual name mangling here since assigning `cls` with an attribute of 2 leading underscores would assign not to the subclass represented by `cls` but in the base class instead.

web_poet/pages.py
```diff
@@ -55,11 +55,22 @@ class ItemPage(Injectable, Returns[ItemT]):
...
+        return getattr(
+            self.__class__, f"_{self.__class__.__name__}__skip_nonitem_fields", False
+        )
 
     def __init_subclass__(cls, skip_nonitem_fields: bool = False, **kwargs):
         super().__init_subclass__(**kwargs)
-        cls._skip_nonitem_fields = skip_nonitem_fields
+
+        # See: https://github.com/scrapinghub/web-poet/issues/141
+        if not getattr(cls, f"_{cls.__name__}__skip_nonitem_fields_set", False):
```
BurnzZ, line 71: `__init_subclass__` is called multiple times but the first one is correct.

kmike (approved):

PR #145, opened 2023-02-27
Add support for default item processors

wRAR:
Now if the field doesn't have `out=` specified, the page object attribute named `default_processors` (feel free to propose an alternative name) will be used in the same way as the `out` decorator argument is used. So per-field processors override default ones, not add to it, just like in `itemloaders`. And just like in `itemloaders` you can skip the default processor by passing `out=[Identity()]` (after we add the `Identity` processor somewhere).

### DIFF
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -505,6 +505,38 @@ async def test_field_processors_async() -> None:
     assert await page.name == "namex"
 
 
+def test_field_processors_default() -> None:
+    def proc1(s):
+        return s + "x"
+
+    @attrs.define
+    class Page(ItemPage):
+        default_processors = [str.strip, proc1]
+
+        @field
+        def name(self):  # noqa: D102
+            return "  name\t "
+
+    page = Page()
+    assert page.name == "namex"
+
+
+def test_field_processors_override_default() -> None:
+    def proc1(s):
+        return s + "x"
+
+    @attrs.define
+    class Page(ItemPage):
+        default_processors = [str.strip]
+
+        @field(out=[proc1])
+        def name(self):  # noqa: D102
+            return "  name\t "
+
+    page = Page()
+    assert page.name == "  name\t x"
+
+
 def test_field_mixin() -> None:
     class A(ItemPage):
         @field
--- a/web_poet/fields.py
+++ b/web_poet/fields.py
@@ -77,11 +77,8 @@ def field(
                 raise TypeError(
                     f"@field decorator must be used on methods, {method!r} is decorated instead"
                 )
-            method = self._processed(method)
-            if cached:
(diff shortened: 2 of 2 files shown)


Gallaecio (approved):
Looks good to me.

Naming-wise, I am a bit unhappy with the fact that `out` and `processors` are used to refer to the same concept, but using something like `default_out` does not sound right either. I cannot think of a better choice than your proposal.

tests/test_fields.py
```diff
@@ -505,6 +505,38 @@ async def name(self):  # noqa: D102
     assert await page.name == "namex"
 
 
+def test_field_processors_default() -> None:
+    def proc1(s):
+        return s + "x"
+
+    @attrs.define
+    class Page(ItemPage):
+        default_processors = [str.strip, proc1]
```
kmike, line 514: Hm, I think default processors should be field-specific, i.e. it should be something like

```py
default_processors = {"name": [str.strip]}
```

I'm not sure about having a default processor for everything, because different fields have different data types.
Also, processors for different fields are likely different.

kmike: > So that redefining a field in a subclass retains the out of the parent field?

Yes, I think that was the idea. Probably you don't even need to define the fields in the base class. 

This way you can create a base class with default processing logic for each field. We can't use e.g. attrs converters in item classes for that, as it makes page.field != item.field, but that's a useful feature, so we need another approach.

kmike: > if most of your fields are strings

In zyte-common-items's Product 10/26 fields are not strings. So, while technically most fields are strings, there are many non-string fields; also, you'd want different default processing for the string fields anyways.

wRAR: Well, that's not the idea I was implementing. If we decide that we don't need to reimplement this `poe` feature this can be closed.

kmike: How would you call `default_processors = {"name": [str.strip]}` attribute? I wonder if we should do something like

```py
default_processors = {
    "name": [str.strip],
    "*": [..],
}
```

kmike: though with such dictionary it's a bit hard to add/change things in a subclass

BurnzZ: > So, while technically most fields are strings, there are many non-string fields; also, you'd want different default processing for the string fields anyways.

If we add the following field to this test page object:

```python
@field
def num(self):
    return 2
```

then calling `page.num` would result in `TypeError: descriptor 'strip' for 'str' objects doesn't apply to a 'int' object` for this test.

But I think in this case, it would be the user's responsibility to take into account the field type when declaring their field processors. For example, they could use `lambda x: x.strip if isinstance(x, str)` instead of `str.strip`.

PR #146, opened 2023-02-27
Another attrs.define fix

kmike:
The idea of the fix: apparently, `__init_subclass__` is called multiple times (once per base class?), but in the incorrect invocations class arguments are not passed. So, we just skip those - do nothing if a default is not set.

I prefer it to the name mangling because there can be multiple classes with the same `cls.__name__`, which means we'd need to switch to something like get_fq_class_name, which makes the implementation more complex.

Checked it using the following code:

```py
from web_poet import Injectable
import attrs


NOT_SET = object()

class BasePage(Injectable):
    _skip_nonitem_fields = NOT_SET
    
    def __init_subclass__(cls, skip_nonitem_fields = NOT_SET, **kwargs):        
        super().__init_subclass__(**kwargs)
        print(cls, kwargs)
        
        if skip_nonitem_fields is not NOT_SET:            
            print(getattr(cls, "skip_nonitem_fields", "NOT_SET"))
            cls._skip_nonitem_fields = skip_nonitem_fields
            print(cls._skip_nonitem_fields)
        print("-"*48)
        print()
        
        
@attrs.define
class MyPage(BasePage, skip_nonitem_fields=True):
    pass


@attrs.define
class MyPage2(MyPage, skip_nonitem_fields=66):
    pass


@attrs.define
class MyPage3(BasePage):
    pass


@attrs.define
class MyPage4(MyPage2):
    pass


@attrs.define
class MyPage5(MyPage2, skip_nonitem_fields=55):
    pass
        
```

### DIFF
--- a/web_poet/pages.py
+++ b/web_poet/pages.py
@@ -62,7 +62,7 @@ class ItemPage(Injectable, Returns[ItemT]):
 
     def _get_skip_nonitem_fields(self) -> bool:
         value = self._skip_nonitem_fields
-        return False if value is _NOT_SET else value
+        return False if value is _NOT_SET else bool(value)
 
     def __init_subclass__(cls, skip_nonitem_fields=_NOT_SET, **kwargs):
         super().__init_subclass__(**kwargs)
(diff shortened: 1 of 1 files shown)


BurnzZ (approved):
I agree, this is a much better solution! :tada: Thanks @kmike !

Gallaecio (approved):

PR #147, opened 2023-03-06
add .readthedocs.yml config

BurnzZ:
Solves the issue of ReadTheDocs errors since it uses `async-lru==2.0.3` instead of `1.0.3` which is intended for Python 3.7 ([reference](https://readthedocs.org/projects/web-poet/builds/19661597/)).

The issue stems from not pinning the Python version in the `docs` toxenv which uses arbitrary Python version in ReadTheDocs.

This PR also adjusts GitHub actions as well.

### DIFF
--- a/.github/workflows/tests-ubuntu.yml
+++ b/.github/workflows/tests-ubuntu.yml
@@ -48,7 +48,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.10']
+        python-version: ['3.11']
         tox-job: ["mypy", "docs", "linters", "types"]
 
     steps:
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -7,7 +7,7 @@ sphinx:
 build:
   os: ubuntu-22.10
   tools:
-    python: "3.7"  # Keep in sync with .github/workflows/tests-ubuntu.yml
+    python: "3.11"  # Keep in sync with .github/workflows/tests-ubuntu.yml
 
 python:
   install:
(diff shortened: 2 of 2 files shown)


.github/workflows/tests-ubuntu.yml
```diff
@@ -48,8 +47,30 @@ jobs:
...
+      run: |
+        tox -e ${{ matrix.tox-job }}
+
+  docs:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.7"]
+        tox-job: ["docs"]
```
BurnzZ, line 73: Not the cleanest solution as there's currently no way for GitHub actions to reuse steps ([reference](https://stackoverflow.com/a/61488170)). If there is, we can make it much shorter.

kmike:
Hey! Given that we'd probably be dropping Python 3.7 at some point, I wonder if it'd be better to use readthedocs.yml file instead, and use a more recent Python version to build docs (see https://docs.readthedocs.io/en/stable/config-file/v2.html#build-tools-python).

BurnzZ:
How does cfbcebe look?

.readthedocs.yml
```diff
@@ -0,0 +1,15 @@
+version: 2
+formats: all
+sphinx:
+  configuration: docs/conf.py
+  fail_on_warning: true
+
+build:
+  os: ubuntu-22.10
+  tools:
+    python: "3.7"  # use min version for async-lru==1.0.3
```
kmike, line 10: Why not use recent Python to build the docs?

BurnzZ: Updated in https://github.com/scrapinghub/web-poet/pull/147/commits/7ddf27624f43aab6a005a921878525ede1781a77

kmike:
Thanks @BurnzZ!

PR #148, opened 2023-03-06
use supported Ubuntu version

kmike:
There is an error in readthedocs:

> Problem in your project's configuration. Invalid "build.os": expected one of (ubuntu-20.04, ubuntu-22.04), got ubuntu-22.10

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,7 +5,7 @@ sphinx:
   fail_on_warning: true
 
 build:
-  os: ubuntu-22.10
+  os: ubuntu-22.04
   tools:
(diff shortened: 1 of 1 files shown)


BurnzZ (approved):

BurnzZ:
Interesting! Thanks @kmike . Applied to scrapinghub/scrapy-poet#136 as well

kmike:
There is one more issue though, which makes the build fail:

> WARNING: conf value "version" should not be empty for EPUB3

BurnzZ:
I propose the latter. #149

PR #149, opened 2023-03-06
turn off fail_on_warning

BurnzZ:
From https://github.com/scrapinghub/web-poet/pull/148#issuecomment-1455611228

Default value is `False`. https://docs.readthedocs.io/en/stable/config-file/v2.html#sphinx-fail-on-warning


kmike:
Thanks @BurnzZ! Docs are building now!

PR #150, opened 2023-03-09
Support for passing instance to the field processors

wRAR:


web_poet/fields.py
```diff
@@ -91,27 +88,41 @@ def __set_name__(self, owner, name):
             getattr(owner, _FIELDS_INFO_ATTRIBUTE_WRITE)[name] = field_info
 
         def __get__(self, instance, owner=None):
+            if self.unbound_method is None:
```
wRAR, line 91: This is partially copied from #145

tests/test_fields.py
```diff
@@ -505,6 +505,47 @@ async def name(self):  # noqa: D102
...
+        @field(out=[proc1])
+        def a(self):  # noqa: D102
+            return "a"
+
+        @field(out=[proc2])
+        def b(self):  # noqa: D102
+            return "b"
+
+    page = Page()
+    with pytest.raises(RecursionError):
```
wRAR, line 544: We need a better test here (one that doesn't raise) but I haven't thought yet what would be a real example of fields that could reference each other (and how should that work).

web_poet/fields.py
```diff
@@ -91,27 +88,41 @@ def __set_name__(self, owner, name):
...
+
             return self.unbound_method(instance)
 
         @staticmethod
-        def _process(value):
-            for processor in out:
-                value = processor(value)
+        def _process(value, processors, instance):
+            for processor in processors:
+                sig = inspect.signature(processor)
```
kmike, line 104: I think we should do it once, when the field is defined, not at every call to the field.

docs/page-objects/fields.rst
```diff
@@ -165,13 +165,27 @@ They will be applied to the field value before returning it:
...
+        @field(out=[clean_tabs, str.strip, add_brand])
         def name(self):
             return self.response.css(".name ::text").get()
 
+        @field(cached=True)
+        def brand(self):
+            return self.response.css(".brand ::text").get()
+
+If a processor takes an argument named ``instance`` it will receive the page
+object instance in it so that values of other fields can be used. You should
```
kmike, line 183: It may be good to note that only the values of sync fields can be used.

Gallaecio (approved):

kmike:
Thanks @wRAR!

PR #151, opened 2023-03-14
Default field processors

wRAR:


web_poet/fields.py
```diff
@@ -80,48 +80,56 @@ def __init__(self, method):
...
             getattr(owner, _FIELDS_INFO_ATTRIBUTE_WRITE)[name] = field_info
 
-        def __get__(self, instance, owner=None):
+        def __get__(self, page, owner=None):
             if self.unbound_method is None:
-                for processor in out or []:
+                if out:
+                    processors = out
+                elif hasattr(page, "Processors"):
+                    processors = getattr(page.Processors, self.name, [])
```
wRAR, line 98: Should we look in the parent classes if there is no processor for the field in this one?

Gallaecio: That’s an option.

Another is to encourage subclassing of processor classes:

```python
    class Page(BasePage):
        class Processors(BasePage.Processors):
            name = [proc1]
```

kmike: I think we should encourage subclassing, it's more explicit.

kmike: A couple more examples:

```py
    class Page(BasePage):
        class Processors(BasePage.Processors):
            name = BasePage.Processors.name + [proc1]

        @field(out=BasePage.Processors.description + [str.strip])
        def description(self): ...
```

kmike:
The implementation and tests look good to me.

web_poet/fields.py
kmike: If we encourage this pattern, it could make sense to add a test for it, to ensure it won't break.

wRAR: I've found an interesting case when implementing this: 

```python
@attrs.define
class BasePage(ItemPage):
    class Processors:
        f = [proc1]

    @field
    def f(self):
        ...

@attrs.define
class Page(BasePage):
    class Processors(BasePage.Processors):
        f = [proc2]

p1 = BasePage()
#p1.f
p2 = Page()
#p2.f
```

There is only one `_field` instance and it gets the processor list only once, on the first call of `__get__()`, so `p2.f`  will use the `proc1` processor if `p1.f` was accessed first.

Looks like we either need to compute the processor list on each call or store it in `owner` instead of `self`? In the second option we could use the `FieldInfo` dict in `owner`?

Gallaecio: The owner approach sounds good to me.

tests/test_fields.py
```diff
@@ -547,6 +570,94 @@ def b(self):  # noqa: D102
         page.b
 
 
+def test_field_processors_default() -> None:
+    @attrs.define
```
Gallaecio, line 574: I think my comments off-GitHub might have lead you to overuse `@define`. I believe `@define` is only needed when new `__init__` parameters (page object inputs) are defined, which is not the case in most of this PR.

The thing I thought worth testing on this front is that, if you have:

```
class BasePage(ItemPage):
    class Processors:
        ...
```

You can later on use `BasePage.Processors`, i.e. `attrs` does not somehow hide the nested class. And if it *does* hide it, we may want to cover it in the docs, and I wonder if using `@define` in that case would solve the issue. But it may just work without `@define`, and I am being too paranoid about `attrs`.

wRAR: I think I may have confused myself, I remembered it as "new fields need `@attrs.define`", I've even seen that some test worked without it...

wRAR:
There is one regression left: accessing `super().fieldname` in `def fieldname()` now raises RecursionError because of the specifics of [descriptor access via super()](https://docs.python.org/3/howto/descriptor.html#invocation-from-super) (basically, you cannot get the base class inside the descriptor, both `owner` and `instance` refer to the subclass).

I think it can be fixed (to help the field defined in the base class find it you can store a reference to it or to the dict with processed methods inside the field itself) but there is another problem with accessing the correct `Processors` class from the descriptor. And this is harder to fix considering there are many combinations here: the `Processors` class can be defined in the base class, the subclass or both, the field can be defined in the base class and redefined in the subclass or not. The following tests are passing after saving `owner` from `__set_name__` and using it instead of `owner` in `__get__` (so the field always uses the processors from the class it is defined in):

```python
def test_field_subclassing_super() -> None:
    class Page(ItemPage):
        @field
        def field1(self):
            return 1

    class Page2(Page):
        @field
        def field1(self):
            return super().field1 + 1

    page = Page()
    assert page.field1 == 1
    page2 = Page2()
    assert page2.field1 == 2

def test_field_processors_super() -> None:
    class BasePage(ItemPage):
        class Processors:
            name = [str.strip]

        @field
        def name(self):
            return "name "

    class Page(BasePage):
        class Processors:
            name = []

        @field
        def name(self):
            base_name = super().name
            return base_name + "2 "

    base_page = BasePage()
    assert base_page.name == "name"
    page = Page()
    assert page.name == "name2 "
```

but the existing `test_field_processors_override` fails because now `Page.f1` (not redefined in `Page` uses the processors from `BasePage.procesors.f1` (redefined in `Page`). I feel like it may even be possible to construct an example where the desired result is not clear.

It may be easier to just forbid using `super()` in fields but I've found the existing code that uses it to try some additional extraction logic and if it fails to defer to the base implementation, there may be other use cases that can't be replaced with overridden processors. It also has very clear semantics when processors are not used, and user could expect that it works.

Gallaecio:
What if we use some kind of sentinel to handle recursion, similar to how I handled recursion in https://github.com/scrapinghub/web-poet/pull/113?

```diff
diff --git a/web_poet/fields.py b/web_poet/fields.py
index 074570c..0b267c7 100644
--- a/web_poet/fields.py
+++ b/web_poet/fields.py
@@ -90,7 +90,10 @@ def field(
             field_info = FieldInfo(name=name, meta=meta, out=out)
             getattr(owner, _FIELDS_INFO_ATTRIBUTE_WRITE)[name] = field_info
 
-        def __get__(self, instance, owner=None):
+            self.owner = owner
+
+        def __get__(self, instance, owner):
+            owner = self.owner if hasattr(instance, "_in_field") else owner
             method = self._get_processed_method(owner, self.name)
             if method is None:
                 if out is not None:
@@ -108,7 +111,13 @@ def field(
                     method = cached_method(method)
                 self._set_processed_method(owner, self.name, method)
 
-            return method(instance)
+            if hasattr(instance, "_in_field"):
+                return method(instance)
+
+            instance._in_field = True
+            result = method(instance)
+            delattr(instance, "_in_field")
+            return result
 
         @staticmethod
         def _get_processed_method(cls, field_name):
```

This seems to get all tests to pass, although I must make it clear that I do not completely understand all elements at play here, and it would not surprise me if such an approach introduced other issues (beyond preventing multithread support).

Gallaecio:
If we consider the forbidding options, the behavior I was the least against in my “playthrough” was that calling `super().f` would get you `f` *without* post-processing (ideally not just without that of the parent class, but also without the one of the current class). It may be even desirable in some cases. And otherwise I can think of workarounds. But I am not sure how complicated that would be, and I imagine 3+-level inheritance trees could also be a problem.

wRAR:
> I think we should really try to make it work without prohibiting some use cases which look like regular Python.

Yeah.

> behavior I was the least against in my “playthrough” was that calling super().f would get you f without post-processing

Unfortunately this should break the use case I mentioined before, 

```python
def f():
    result = self.css('h1::text')
    if not result:
        return super().f
```

wRAR:
The problem is more complicated when the processed method caching is implemented as we need to put the cached method in some correct place so it makes sense to ignore this part when thinking about the main part.

I think the main problem with super() is that we want to use processors from some intermediate class while all ther use cases either use the class of the instance itself, i.e. the "final" subclass, the class where the field itself is defined, or the first ancestor where the thing is defined if it's not defined on the class we are using:

* If the field we access is defined in the class of the instance, the code path is very simple: the field has access to this instance and can request its `Processors` (which is either defined in that class or in some of its ancestors), it doesn't need anything else.
* If the field is defined in some ancestor of the instance, it has access to that ancestor (`owner` in `__set_name__`) but doesn't need it, as it needs `Processors` from the subclass and can get it via the instance.
* When a field method calls the base class field method via super(), if no processors are involved, Python will just find the first ancestor with the field defined and call it with `self`, no problems here because the only thing we need from that ancestor is the field and the method decorated with it.
* But when a field method uses super() and there are some processors involved, we want some special resolution:
  * We want the field from the closest ancestor class where it's present - this is done by Python automatically when we access the field via super().
  * We want `Processors` also from the closest ancestor class where it's present and this ancestor needs to be found separately from the field:
    * we cannot use the field owner (`Processors` we need may be in a subclass of it),
    * we cannot use the instance itself (we need to skip its `Processors`),
    * we cannot do the resolution in the descriptor itself (we don't know if it's called via super() or normally),
    * we don't have a reference to the immediate ancestor of the instance (and we cannot get it manually because again we don't know if we should).
  * We also can't skip looking for processors because we want to return the processed value (also I think we don't know when we could skip it if we decided to).
  * By using the proposed `_in_field` attribute we can guess that the descriptor is called via super(), but if the parent field also calls super() we can't distinguish this. We can try incrementing it instead just setting it to True and calling super() the same number of times, though I would prefer something cleaner.

Gallaecio:
Scenario to consider:

```ptyhon
class A:
    @field
    def a(self):
        return "1"

class B(A):
    @field
    def a(self):
        return super().a
```

Imagine each class is developed by a different team. Now, A is refactored:

```
class A:
    @field(out=[str])
    def a(self):
        return 1
```

B now returns an int?

kmike:
@Gallaecio 

Hm, yes, I'd be surprised if B.a doesn't return int. In the example B.a implementation is just a pass-through, isn't it?

wRAR:
> Using processors from the current class when super().field is called looks consistent with how Python works. See Page2 and Page3 classes - accessing the field from the base class probably shouldn't work differently in these cases.

This actually makes sense, though my initial thought was that this doesn't look consistent with what the user of super() would expect (but I may be wrong with that too as the semantics are not 100% clear).

The two use cases of super() I've seen probably have the following user expectations:

* Subclass gets the actual data from the base class and applies some post-processing inside the field method. On one hand this should be refactored into a processor, on the other hand we want to keep some compatibility until it's refactored. But what is the user expectation when fields have processors? I think it's getting the processed value (using the processors from the base class? or it doesn't matter as it would be the same processors in both classes?) inside the method and *not* running the processors again after the method returns. I think this can be done by setting `out=[]` on the subclass field.
* Subclass has two code paths: its own extraction and `return super().f` as a fallback. For the data extracted in the subclass we want to run processors (so the field can't have `out=[]`), for the fallback we want to run processors, not sure which ones, but only once.

Gallaecio:
> Hm, yes, I'd be surprised if B.a doesn't return int. In the example B.a implementation is just a pass-through, isn't it?

It is. It is a passthrough to return a `str`, that would start returning an `int` after a refactor of the parent class that still returns a `str`. How can you refactor A.a to return a string from `1` without potentially breaking subclasses? Do you go for `return str(1)` and no processors?

---

On a different note, I guess for some use cases composition may be best:

```python
class B(A):
    a: A

    @field
    def a(self):
        return a.a
```

Maybe we do not need to aim to support with `super()` scenarios that can be supported through composition.

wRAR:
> Using processors from the current class when super().field is called looks consistent with how Python works.

This is now implemented.

The issue with running processors twice still exists, I would even say it's a matter for a separate improvement.

> provide a way (a function / method / property) to get field value without applying any processors

This is basically calling the underlying method decorated by `@field` (though it needs separate caching for `cached=True`), but I don't know what API can we use for this, as `A.f` returns the field value, not something we could attach a method/attribute/property to.

> provide a way (a function / method) to get field value with a specific Processors class applied

Depending on the API we want this can be easy or more complicated (in the simplest case it's just extracting the callable list and running it).

tests/test_fields.py
kmike: Shouldn't attrs.define be removed?

kmike: New dependencies need attrs.define; new fields shouldn't need it.

kmike:
@Gallaecio I misunderstood your example :) But it seems it should work as expected, because the behavior of out=... processors shouldn't be changed in this PR, unless I'm missing something.

tests/test_fields.py
wRAR: I think I've kept this one in f64385f38a25fabf83c89823f020b8d35d3933a3 to show that the comment in this thread  ("The thing I thought worth testing on this front [...]") indeed works.

web_poet/fields.py
```diff
@@ -81,54 +83,80 @@ def __init__(self, method):
...
                 )
             self.original_method = method
-            self.unbound_method = None
-            self.processors: List[Tuple[Callable, bool]] = []
+            self.name: Optional[str] = None
+            self.full_name: Optional[str] = None
 
         def __set_name__(self, owner, name):
+            self.name = name
+            self.full_name = f"{get_fq_class_name(owner)}.{name}"
```
kmike, line 91: I *think* it's possible to have 2 classes with the same fq name, e.g. if they're defined inside a function. Would using something like `id` work here? Either `f'{id(owner).{name}`, or some id of the field object itself. The attribute won't be named `full_name`, but it might provide a more robust "id" for the field.

wRAR: We can just use `id(self)` then.

web_poet/fields.py
```diff
@@ -81,54 +83,77 @@ def __init__(self, method):
...
+            # take into account the instance class and the field object. So we
+            # use the field object id() as a key when caching the method in
+            # the instance class.
+            cache_key = id(self)
+            method = self._get_processed_method(owner, cache_key)
+            if method is None:
+                if out is not None:
+                    processor_methods = out
+                elif hasattr(instance, "Processors"):
+                    processor_methods = getattr(owner.Processors, self.name, [])
```
kmike, line 108: Is there a difference between instance.Processors and owner.Processors? I'm asking because the code checks that instance has "Processors" attribute, but then acesses owner.Processors. 

If there is no difference, it'd be probably better to use the same approach in both cases. If there is a difference, could you please add a comment to the code which explains it?

web_poet/fields.py
```diff
@@ -81,54 +83,77 @@ def __init__(self, method):
...
+            # We use the original method and the out arg from the field and
+            # the Processors class from the instance class, so caching needs to
+            # take into account the instance class and the field object. So we
+            # use the field object id() as a key when caching the method in
+            # the instance class.
+            cache_key = id(self)
+            method = self._get_processed_method(owner, cache_key)
+            if method is None:
+                if out is not None:
+                    processor_methods = out
```
kmike, line 106: processor_functions? They're not usually methods, right?

kmike (approved):
Thanks @wRAR! The behavior and the implementation looks good to me. I think that's fine to address remaining use cases separately.

Re-requesting review from @Gallaecio :)

web_poet/fields.py
wRAR: > I'm asking because the code checks that instance has "Processors" attribute, but then acesses owner.Processors.

Good catch, this is indeed a mistake, but which of them we use is not that important, unless the user changes the instance's `Processors` attribute.

Gallaecio (approved):

PR #152, opened 2023-03-14
Remove intro/from-ground-up and spread its info

Gallaecio:
Major changes:
- Remove the “intro/from-ground-up” page altogether. This is the most radical change, and I am willing to back-pedal somewhat, i.e. keep this page in some form.
  My thinking is that this page tries to cover too many things at once that may be better done separately:
  - It justifies the usefulness of web-poet, which I think the overview page does better with its 4 bullet-point list.
  - It seemingly explains how its design came to be, which I find fascinating, but not useful. It feels better suited for a blog post.
  - It covers basic concepts not covered elsewhere, such as what a page object is, but only in the middle of the page, making those sections hard to discover (feels like you have to read a lot to even get a description of a page object). Separate topics for those concepts, to be expanded as the project evolves, feels like a better approach to me.
- Add new topics loosely based on the removed content, and a few additional details so that the new pages do not feel noticeably lacking:
  - Page objects, to cover the concept and basic usage.
  - Inputs, to cover defining page object inputs.
  - Items, to cover the output of the to_item method.
  - Frameworks, to cover known web-poet ~frameworks~ framework.
  -  Framework specification, an entry page for then section about writing frameworks.

Minor changes:
- Capitalization normalization on table of contents entries.
- Show the table of contents on the index page, as it felt lackluster to me otherwise.
- Sorted pages in the page object section based on how soon I think their concepts should be introduced, which is loosely based on the order in which those concepts are introduced in the tutorial.

Please, *read the new content carefully*. While it is based on the removed page, it also includes some new content, and I fear I might have made the wrong assumptions in some cases. For example:
- Do we want to require frameworks to support both returning items and page objects, as I wrote, or should we allow only one of those, and if so, any of them or a specific one?
- I state frameworks must support all page_input classes, but maybe we should be less strict, or make a special case out of BrowserHtml.


docs/page-objects/inputs.rst
```diff
@@ -0,0 +1,51 @@
+.. _inputs:
+
+======
+Inputs
+======
+
+:ref:`Page object classes <page-object-classes>`, in their ``__init__`` method,
+must define input parameters with type hints pointing to classes from the
+:mod:`web_poet.page_inputs` module.
```
kmike, line 9: As you mention later, it could be custom classes as well.

Also, one may use other injectable things as dependencies: other page objects, items (given that the framework implements providing items).

docs/page-objects/items.rst
```diff
@@ -0,0 +1,59 @@
...
+To define a base class that subclasses :class:`~web_poet.pages.ItemPage` or
+:class:`~web_poet.pages.WebPage` but does not define an actual output item
+class either, use :data:`~web_poet.pages.ItemT` as its output item class:
+
+.. code-block:: python
+
+   from web_poet.pages import ItemT
+
+
+   class BasePage(WebPage[ItemT]):
```
kmike, line 42: Hm, I was thinking you can skip ItemT in this case, and just inherit from WebPage

docs/page-objects/index.rst
```diff
@@ -0,0 +1,121 @@
...
+
+.. literalinclude:: attrs.py
+
+Also, it is often best to subclass :class:`~web_poet.pages.ItemPage`, which
+subclasses :class:`~web_poet.pages.Injectable` and provides an implementation of the
+``to_item`` method based on :ref:`declared fields <fields>`:
+
+.. literalinclude:: itempage.py
+
+If you page object class needs :class:`~web_poet.page_inputs.HttpResponse` as
```
wRAR, line 53:
```suggestion
If your page object class needs :class:`~web_poet.page_inputs.HttpResponse` as
```

wRAR (approved):

docs/page-objects/items.rst
Gallaecio: I’m now thinking we can remove that scenario entirely from the docs.

Gallaecio:
@kmike I’ve “restored” from-ground-up, however, I have edited it significantly.

I feel like the version proposed here still accomplishes the goal of showing how a simple web scraping code snippet can be transformed into web-poet code through iterations of improvements, which being much shorter than the original version, and skipping some details to be covered in latter pages.

However, I am OK with restoring a version almost identical to the original one if you prefer that approach.

docs/intro/from-ground-up.rst
```diff
@@ -219,542 +183,70 @@ objects, so ``web-poet`` provides a shortcut for it: inherit from
...
 
-It might be easier to read the code written this way. Also, this style
-allows to extract only some of the attributes - if you don't need
-the complete to_item() output, you still can access individual properties.
-
-web-poet provides a small framework to simplify writing Page Objects
-in this style; see :ref:`fields`. The example above can be simplified
-using web-poet fields - there is no need to write ``to_item`` boilerplate:
+At this point you may be wondering why web-poet requires you to write a class
+with a ``to_item`` method rather than a function. The answer is flexibility.
```
kmike, line 189: There is a second reason, related to andi: classes can be used as types in Python, while functions can't. So, you can type annotate "I need an instance of this class", but you can't say "I want this function specifically". Currently typing is used for dependency injection.

If not for dependency injection, https://github.com/scrapinghub/web-poet/pull/44/files wouldn't need to create a class from a function.

I'm not sure if we should go into explaining all this in this document though. Could be worth adding a note, but I'm not sure :)

docs/page-objects/index.rst
```diff
@@ -0,0 +1,121 @@
+.. _page-objects:
+
+============
+Page objects
+============
+
+A page object is a code wrapper for a webpage that implements the logic to
```
kmike, line 7: It may be also a wrapper for a part of webpage.

docs/page-objects/index.rst
```diff
@@ -0,0 +1,121 @@
...
+
+.. _page-object-classes:
+
+Defining a page object class
+============================
+
+A page object class is a Python class that:
+
+-   Subclasses :class:`~web_poet.pages.Injectable` or is registered or decorated
+    with :class:`Injectable.register <web_poet.pages.Injectable.register>`.
```
kmike, line 22: Some page object features are not supported with Injectable: fields, test generation (or something else about tests, I don't recall), input validation. So I wonder if we should recommend ItemPage by default, and maybe explain Injectable separately, if there is a need.

kmike: The same might apply to from-ground-up document; it uses Injectable.

docs/page-objects/index.rst
```diff
@@ -0,0 +1,121 @@
...
+
+.. _output-item:
+
+Getting the output item
+=======================
+
+You should :ref:`include your page object classes into a page object
+registry <rules-intro>`, e.g. decorate them with :func:`~.handle_urls`:
+
+.. literalinclude:: register.py
```
kmike, line 70: What do you think about moving .py files to a sub-folder, so that they're not mixed with .rst files?

docs/page-objects/index.rst
```diff
@@ -0,0 +1,121 @@
...
+
+Then, provided your page object class code is imported (see
+:func:`~web_poet.rules.consume_modules`), your :ref:`framework <frameworks>`
+can build the output item after you provide the required input, such as the
+target URL and the desired :ref:`output item class <items>`, as :ref:`shown in
```
kmike, line 75: In the register.py example the item class is dict, so this doesn't work. I don't think it's intended that you can ask for a "dict" as an item, and get result from some page object which returns a dict.

docs/page-objects/index.rst
```diff
@@ -0,0 +1,121 @@
...
+the tutorial <tutorial-create-page-object>`.
+
+Your framework chooses the right page object class based on your input
+parameters, downloads the required data, builds a page object, and calls the
+``to_item`` method of that page object. All transparently to you.
```
kmike, line 80: I'd remove "All transparently to you." note, it sounds a bit scary :)

docs/page-objects/index.rst
```diff
@@ -0,0 +1,121 @@
...
+
+        .. code-block:: python
+
+           item = await foo_page.to_item()
+
+    This also means that, if the underlying page object class ever switches its
+    implementation from synchronous to asynchronous or the other way around,
+    your code will stop working.
+
+-   ``to_item`` may raise a :exc:`~web_poet.exceptions.core.Retry` exception
```
kmike, line 109: Should it be extended with other exceptions which we have now?

docs/page-objects/index.rst
```diff
@@ -0,0 +1,121 @@
...
+    which, depending on your :ref:`framework <frameworks>`, may not be handled
+    automatically when getting a page object instead of an item.
+
+
+Building a page object manually
+-------------------------------
+
+While not recommended, it is possible to create a page object from a page
+object class passing its inputs as parameters. For example, to manually create
+an instance of the ``FooPage`` page object class defined above:
```
kmike, line 119: What do you think about explaining why it's not recommended? The most important thing is that the code breaks if a page object requires a different input, while in case of using framework a page object is free to change its inputs.

docs/page-objects/inputs.rst
```diff
@@ -0,0 +1,65 @@
...
+:ref:`Page object classes <page-object-classes>`, in their ``__init__`` method,
+must define input parameters with type hints pointing to input classes.
+
+Those input classes may be:
+
+-   :ref:`Built-in web-poet input classes <built-in-inputs>`.
+
+-   :ref:`Custom input classes <custom-inputs>`.
+
+-   Other :ref:`page object classes <page-object-classes>`.
```
kmike, line 16: Technically, it's other Injectable classes. Just a heads up: if you change the example from Injectable to ItemPage in the explanation for page object classes, it may require some changes here.

docs/page-objects/inputs.rst
```diff
@@ -0,0 +1,65 @@
...
+
+-   :ref:`Item classes <items>`, when using a :ref:`framework <frameworks>`
+    that can provide item classes.
+
+Based on the target URL and parameter type hints, :ref:`frameworks
+<frameworks>` automatically build the required objects at run time, and pass
+them to the ``__init__`` method of the corresponding page object class.
+
+For example, if a page object class has an ``__init__`` parameter of type
+:class:`~web_poet.page_inputs.HttpResponse`, and the target URL is
```
kmike, line 26: The link doesn't work for some reason.

docs/page-objects/inputs.rst
```diff
@@ -0,0 +1,65 @@
...
+https://example.com, your framework would send an HTTP request to
+https://example.com, download the response, build an
+:class:`~web_poet.page_inputs.HttpResponse` object with the response data, and
```
kmike, line 29: The link doesn't work for some reason.

docs/page-objects/inputs.rst
```diff
@@ -0,0 +1,65 @@
...
+
+-   :class:`~web_poet.page_inputs.HttpClient`, to send  :ref:`additional
+    requests <additional-requests>`.
+
+-   :class:`~web_poet.page_inputs.RequestUrl`, the target URL before following
+    redirects. Useful, for example, to skip the target URL download, and
+    instead use :class:`~web_poet.page_inputs.HttpClient` to send a custom
+    request based on parts of the target URL.
+
+-   :class:`~web_poet.page_inputs.PageParams`, to receive data from the
```
kmike, line 52: The links here don't work for me, when building docs locally.

docs/page-objects/inputs.rst
```diff
@@ -0,0 +1,65 @@
...
+
+Built-in input classes
+======================
+
+The :mod:`web_poet.page_inputs` module defines multiple classes that you can
+define as inputs for a page object class, including:
+
+-   :class:`~web_poet.page_inputs.HttpResponse`, a complete HTTP response,
+    including URL, headers, and body. This is the most common input for a page
+    object class.
```
kmike, line 42: Would it make sense to mention BrowserHtml as well? It looks different from other classes (e.g. HttpResponseBody), because it's not covered by HttpResponse.

docs/api-reference.rst
```diff
@@ -1,14 +1,16 @@
 .. _api-reference:
 
 =============
-API Reference
+API reference
```
kmike, line 4: I'm not sure if that's new or not, but we got all the `bytes` / `str` methods documented for page input classes; I don't think it's intended. That's probably unrelated to this PR.

docs/page-objects/items.rst
```diff
@@ -0,0 +1,47 @@
...
+=====
+Items
+=====
+
+The ``to_item`` method of a :ref:`page object class <page-object-classes>` must
+return an item.
+
+An item is a data container object supported by the itemadapter_ library.
+Because itemadapter_ allows implementing support for arbitrary types, any kind
+of Python object can potentially work as an item.
```
kmike, line 12: What do you think about mentioning that common examples of items which you can use are

* dict
* attrs instance
* dataclass instance?

docs/page-objects/items.rst
```diff
@@ -0,0 +1,47 @@
...
+
+.. code-block:: python
+
+   class MyPage(WebPage[MyItem]):
+       ...
+
+If you have a subclass of :class:`~web_poet.pages.ItemPage` or
+:class:`~web_poet.pages.WebPage` that declares an output item class, and
+you wish to create a subclass of it that changes the output item class, use
+the :class:`~web_poet.pages.Returns` mixin to re-declare its output item class:
```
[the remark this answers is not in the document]
kmike: I'd probably not go into explaining Returns here, as it's covered in the field docs, and it's a more advanced concept.

docs/page-objects/items.rst
kmike: It may also worth it to add a simple example of an item definition.

docs/page-objects/items.rst
```diff
@@ -0,0 +1,47 @@
...
+.. code-block:: python
+
+   from web_poet import Returns
+
+
+   class ParentPage(WebPage[ParentItem]):
+       ...
+
+   class ChildPage(ParentPage, Returns[ChildItem]):
+       ...
```
kmike, line 47: What do you think about adding a section here, which explains that it's an anti-pattern to use items which are very "smart", i.e. use features like attrs convertors? We don't have to do it in this PR though.

docs/frameworks/index.rst
```diff
@@ -0,0 +1,55 @@
...
+
+    To do that:
+
+    -   Use ``web_poet.default_registry`` to determine the right :ref:`page
+        object class <page-object-classes>` to use.
+
+        Optionally, let users indicate their own
+        :class:`~web_poet.rules.RulesRegistry` object.
+
+    -   Support building :ref:`input objects <inputs>` based on type hints,
```
kmike, line 32: This deserves its own point. I think that's the most important thing a framework should implement. 

Some time ago scrapy-poet didn't support returning item classes directly; it also didn't support additional requests, and not all standard inputs are supported even now. I would still consider such framework a framework, as soon as it can build an instance of a page object, given a page object class. 

Everything else should probably be optional for a framework. 
It's also nice if a framework fails clearly when some standard features are not supported.

docs/frameworks/index.rst
```diff
@@ -0,0 +1,55 @@
...
+        i.e. dependency injection.
+
+        You must support all input classes from the :mod:`web_poet.page_inputs`
+        module.
```
kmike, line 36: I don't think it's a requirement. scrapy-poet doesn't support all input classes from page_inputs. If something is not supported, a framework should raise an error.

docs/frameworks/index.rst
```diff
@@ -0,0 +1,55 @@
...
+
+        You can use the andi_ library for that. For practical examples, see the
+        source code of scrapy-poet_ and of the :mod:`web_poet.example` module.
```
kmike, line 39: web_poet.example link doesn't work for me

docs/frameworks/index.rst
```diff
@@ -0,0 +1,55 @@
...
+
+        .. _andi: https://github.com/scrapinghub/andi
+        .. _scrapy-poet: https://github.com/scrapinghub/scrapy-poet
+
+        In addition to signature inspection, andi_ it also handles
+        :class:`~typing.Optional` and :class:`~typing.Union` annotations, and
```
kmike, line 45: These links don't work for me. Hm, is there something with my setup?

kmike (approved):
Fantastic changes @Gallaecio! The from-ground-up.rst is so much better now, as well as the overall docs structure.

docs/page-objects/index.rst
Gallaecio: I agree with recommending `ItemPage` here.

However, I also think this section needs to be accurate about what is a page object. Is a class inheriting directly from `Injectable` a valid page object? If so, then that’s the “minimum” base class that makes a class a page object.

kmike: Yeah, I was thinking about the same. It seems raw Injectable doesn't have all the page object features now - no validation, no fields, tests don't always work. So, probably we should say that page object is an instance of ItemPage, and Injectable is for more advanced use cases, if some ItemPage features get in a way - does it make sense? E.g. it should be fine to define a dependency like SemanticMarkupData (which itself depends on HtpResponse) as a subclass of Injectable.

docs/page-objects/index.rst
Gallaecio: This works:

```python
from web_poet import WebPage, field, handle_urls
from web_poet.example import get_item


@handle_urls("books.toscrape.com")
class BookPage(WebPage[dict]):

    @field
    async def title(self):
        return self.css("h1::text").get()


item = get_item(
    "http://books.toscrape.com/catalogue/the-exiled_247/index.html",
    dict,
)
print(item)
# {'title': 'The Exiled'}
```

But it would make sense to me to explain at this point that, even though this works, it is a good practice to use something other than dict.

kmike: I think it's an anti-pattern to use Page[dict]. Probably it'd be better to remove it from all examples. For example, consider dependency injection. If you have some parameter which is of type dict (e.g. an optional argument which user defind for some reason), it can be surprising that it's populated by some random page object - `dict` doesn't tell anything about the kind of data you'd like to get.

docs/api-reference.rst
Gallaecio: Not what I’m getting:

Screenshot_20230330_093847
Screenshot_20230330_093923

docs/page-objects/inputs.rst
kmike: So, is the plan to add it here once we implement BrowserHtml support e.g. in scrapy-zyte-api?

docs/page-objects/items.rst
Gallaecio: I was about to do it in this PR, as I agree that it makes perfect sense in this page, but I cannot think of a good wording, and I expect a few iterations, so it may indeed be better left for a follow-up pull request.

docs/intro/from-ground-up.rst
Gallaecio: https://github.com/scrapinghub/web-poet/pull/152/commits/fcd364061034500a74d32dfd315572a05837e3f8

docs/page-objects/index.rst
Gallaecio: https://github.com/scrapinghub/web-poet/pull/152/commits/73f59eb278d85b895a8279e7cd26e97f17d2b002

docs/page-objects/index.rst
Gallaecio: https://github.com/scrapinghub/web-poet/pull/152/commits/ba33547fcd27ec843c522eb57934ec84a9a99609

docs/page-objects/index.rst
Gallaecio: https://github.com/scrapinghub/web-poet/pull/152/commits/9f58ae7a428d519fae252f29c9b76ed865043988

docs/page-objects/index.rst
Gallaecio: https://github.com/scrapinghub/web-poet/pull/152/commits/946c8d7b72b701b96ba02c1477ef278ec0bdc488

docs/page-objects/index.rst
Gallaecio: https://github.com/scrapinghub/web-poet/pull/152/commits/d29cb07ce315762ced5aad90cf7afe36af555aab

docs/page-objects/index.rst
```diff
@@ -0,0 +1,121 @@
...
+
+    -   If the ``to_item`` method is asynchronous:
+
+        .. code-block:: python
+
+           item = await foo_page.to_item()
+
+    This also means that, if the underlying page object class ever switches its
+    implementation from synchronous to asynchronous or the other way around,
+    your code will stop working.
```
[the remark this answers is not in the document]
Gallaecio: https://github.com/scrapinghub/web-poet/pull/152/commits/67840f675c04b4e1134d74fe49239815420129ca

docs/page-objects/index.rst
Gallaecio: https://github.com/scrapinghub/web-poet/pull/152/commits/6b3b7ae0e6e3883ab2b43867076ef54de16ffa9f

docs/page-objects/index.rst
Gallaecio: https://github.com/scrapinghub/web-poet/pull/152/commits/e6d9116c7c82e7e6ccad31a9a06994c9f3b3cb66

docs/page-objects/inputs.rst
Gallaecio: [I’ve added an extra bullet point for “other injectable classes”](https://github.com/scrapinghub/web-poet/pull/152/commits/71ddcdfb760b7e8f9fc732a70e56e1f21aa796cb).

docs/page-objects/inputs.rst
Gallaecio: https://github.com/scrapinghub/web-poet/pull/152/commits/b958a1a25cc9d04a9d16de4b5f114fc74c44d480

docs/page-objects/inputs.rst
Gallaecio: https://github.com/scrapinghub/web-poet/pull/152/commits/b958a1a25cc9d04a9d16de4b5f114fc74c44d480

docs/page-objects/inputs.rst
Gallaecio: https://github.com/scrapinghub/web-poet/pull/152/commits/b958a1a25cc9d04a9d16de4b5f114fc74c44d480

docs/page-objects/items.rst
Gallaecio: https://github.com/scrapinghub/web-poet/pull/152/commits/78d03f5a383d3471f619cb8593dc1ed4d5b81a38
https://github.com/scrapinghub/web-poet/pull/152/commits/e18ccc9570547a5dd7919158e63e949e35f3f58e

docs/page-objects/items.rst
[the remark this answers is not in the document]
Gallaecio: https://github.com/scrapinghub/web-poet/pull/152/commits/15a3973fa19f58f6e877e8f592ada308c5162751

docs/frameworks/index.rst
Gallaecio: https://github.com/scrapinghub/web-poet/pull/152/commits/3baff1225d325685e36c9d84467822da382c240b
https://github.com/scrapinghub/web-poet/pull/152/commits/0553b6c6e0745c29cfb35e5be319c604665d4cc7

docs/frameworks/index.rst
Gallaecio: https://github.com/scrapinghub/web-poet/pull/152/commits/9e13da9b944889240ed6e2421775f507a920ca6b

docs/frameworks/index.rst
Gallaecio: https://github.com/scrapinghub/web-poet/pull/152/commits/3b2f88ec09810b58f024b03f73e66290c6fbf1a8

docs/api-reference.rst
Gallaecio: I’ve tried recreating the `docs` Tox environment, and I still get the expected rendering. Could you try recreating your `docs` Tox environment on your end?

kmike (approved):
+1 to merge.

docs/api-reference.rst
kmike: It turns out I was not using the right docs environment. I've installed docs/requirements.txt, and the problem went away. Old Sphinx, probably.

PR #153, opened 2023-03-17
Fix mixin support

Gallaecio:


wRAR:
I think I'm missing the context :)

Gallaecio:
This is fixing 2 issues that I found while working out why https://github.com/zytedata/zyte-common-items/pull/29 was happening.

The problems are covered by the the addition of `class C(Mixin, A)` to the new test, as well as the corresponding new assert: `assert set(get_fields_dict(C)) == {"a", "c", "mixin"}`.

Before the changes here, `set(get_fields_dict(C))` is actually `{'a', 'b', 'c', 'mixin'}`, i.e. `b` gets leaked from `B` through `Mixin`. This is solved by making `Mixin` inherit from `FieldsMixin`, which I realize now is a base class that any mixin providing fields needs to subclass to work. `FieldsMixin.__init_subclass__` takes care of avoiding that leak.

However, when you solve that *test issue*, changing `Mixin` to `Mixin(FieldsMixin)` in the test, another issue occurs: `set(get_fields_dict(B))` (notice it’s the old `B` assert, not the new `C` assert) becomes `{'b', 'mixin'}`, i.e. loses the `a` from its parent class `A`. This is because `FieldsMixin` did not support multiple inheritance. In fact, if you change `B(Mixin, A)` to `B(A, Mixin)`, the outcome becomes `{'a', 'b'}`, which is also wrong, but now the missing field is `mixin`, not `a`.

So, this second issue, which is really the only issue in scrapy-poet itself (not inheriting from FieldsMixin was a test issue), is addressed by modifying `FieldsMixin.__init_subclass__` to take multiple-inheritance into account, inspecting the fields of all its base classes and not just those of the first one.

wRAR (approved):

kmike:
Thanks @Gallaecio! It seems we also might want to add some notes to the docs about FieldsMixin, that it's required if you want to create a mixin which provides fields.

PR #158, opened 2023-03-30
Added HttpResponse.status in example implementation

### LINKED ISSUE #157: Map HttpResponse.status in the example implementation
Update https://github.com/scrapinghub/web-poet/blob/35c67d6bf018a823b3fa973d0b729761e394d934/web_poet/example.py#L35-L41 to also map the response status code into the returned HttpResponse object.

sjdex:
This closes #157.

### DIFF
--- a/web_poet/example.py
+++ b/web_poet/example.py
@@ -36,6 +36,7 @@ def _get_http_response(url: str) -> HttpResponse:
     response = requests.get(url)
     return HttpResponse(
         response.url,
+        status=response.status_code,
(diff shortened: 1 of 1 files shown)


Gallaecio (approved):
Great!

It would be even better if you could add a test for this to https://github.com/scrapinghub/web-poet/blob/master/tests/test_example.py, e.g. mocking request.get. However, given the lack of existing test coverage for this part of the code, I am OK with merging as is.

PR #161, opened 2023-04-04
Improve and document processors for nested items

wRAR:
This extracts `ItemBase` from `ItemPage`, the only difference between them being `ItemPage` inheriting from `Injectable`. It also adds `ItemPartial` which too inherits from `ItemBase` but adds `FieldsMixin` instead of `Injectable`. Actually, as `Injectable` inherits from `FieldsMixin` we coud have just `ItemBase` (inheriting from `FieldsMixin`) and `ItemPage` and use `ItemBase` directly in the user code.

Both new classes can be renamed.

### DIFF
--- a/docs/page-objects/fields.rst
+++ b/docs/page-objects/fields.rst
@@ -272,7 +272,7 @@ Some item fields contain nested items (e.g. a product can contain a list of
 variants) and it's useful to have processors for fields of these nested items.
 You can use the same logic for them as for normal fields if you define an
 extractor class that produces these nested items. Such classes should inherit
-from :class:`~.Extractor`. In simplest cases you need to pass a selector to
+from :class:`~.Extractor`. In the simplest cases you need to pass a selector to
 them:
 
 .. code-block:: python
--- a/docs/page-objects/fields.rst
+++ b/docs/page-objects/fields.rst
@@ -316,6 +316,7 @@ dictionaries with some data:
 
 .. code-block:: python
 
+    @attrs.define
     class VariantExtractor(Extractor):
         variant_data: dict
 
--- a/docs/page-objects/fields.rst
+++ b/docs/page-objects/fields.rst
@@ -272,7 +272,8 @@ Some item fields contain nested items (e.g. a product can contain a list of
 variants) and it's useful to have processors for fields of these nested items.
 You can use the same logic for them as for normal fields if you define an
 extractor class that produces these nested items. Such classes should inherit
-from :class:`~.Extractor`:
+from :class:`~.Extractor`. In simplest cases you need to pass a selector to
+them:
 
 .. code-block:: python
 
@@ -299,6 +300,28 @@ from :class:`~.Extractor`:
         @field(out=[str.strip])
         def color(self):
             return self.sel.css(".name::text")
+
+In such cases you can also use :class:`~.SelectorExtractor` as a shortcut that
+provides ``css()`` and ``xpath()``:
+
+.. code-block:: python
+
+    class VariantExtractor(SelectorExtractor):
+        @field(out=[str.strip])
+        def color(self):
+            return self.css(".name::text")
+
+You can also pass other data in addition to, or instead of, selectors, such as
+dictionaries with some data:
+
+.. code-block:: python
+
+    class VariantExtractor(Extractor):
+        variant_data: dict
+
+        @field(out=[str.strip])
+        def color(self):
+            return self.variant_data["color"]
 
 
 .. _item-classes:
--- a/docs/page-objects/fields.rst
+++ b/docs/page-objects/fields.rst
@@ -270,15 +270,15 @@ Processors for nested fields
 
 Some item fields contain nested items (e.g. a product can contain a list of
 variants) and it's useful to have processors for fields of these nested items.
-You can use the same logic for them as for normal fields if you define a
-"partial page object" class that produces these nested items. Such classes
-should inherit from :class:`~.ItemPartial`:
+You can use the same logic for them as for normal fields if you define an
+extractor class that produces these nested items. Such classes should inherit
+from :class:`~.Extractor`:
 
 .. code-block:: python
 
     import attrs
     from parsel import Selector
-    from web_poet import ItemPage, HttpResponse, field
+    from web_poet import Extractor, ItemPage, HttpResponse, field
 
     @attrs.define
     class MyPage(ItemPage):
@@ -287,13 +287,13 @@ should inherit from :class:`~.ItemPartial`:
         @field
         async def variants(self):
             variants = []
-            for color in self.response.css(".color"):
-                variant = await VariantPartial(color).to_item()
+            for color_sel in self.response.css(".color"):
+                variant = await VariantExtractor(color_sel).to_item()
                 variants.append(variant)
             return variants
 
     @attrs.define
-    class VariantPartial(ItemPartial):
+    class VariantExtractor(Extractor):
         sel: Selector
 
         @field(out=[str.strip])
--- a/docs/page-objects/fields.rst
+++ b/docs/page-objects/fields.rst
@@ -206,6 +206,7 @@ named ``Processors``:
 
 .. code-block:: python
 
+    import attrs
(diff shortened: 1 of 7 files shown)


wRAR:
> Actually, as Injectable inherits from FieldsMixin we coud have just ItemBase (inheriting from FieldsMixin) and ItemPage and use ItemBase directly in the user code.

This is now done, `ItemBase` is renamed to `Extractor`. Item validation is moved to `ItemPage`, but I needed to keep a `_validate_input` stub in `Extractor` as the field machinery expects it, it's possible that it can be refactored instead.

I've also added `SelectorExtractor` that provides `self.selector` and shortcuts for it. Shortcuts are copied from `SelectableMixin`, I don't think we need a separate mixin for those but it's possible :)

web_poet/__init__.py
```diff
@@ -12,7 +12,15 @@
     RequestUrl,
     ResponseUrl,
 )
-from .pages import Injectable, ItemPage, ItemWebPage, Returns, WebPage, validates_input
+from .pages import (
+    Extractor,
```
kmike, line 16: we should also include SelectorExtractor here

web_poet/pages.py
```diff
@@ -100,15 +100,23 @@ def __init_subclass__(cls, skip_nonitem_fields=_NOT_SET, **kwargs):
             return
         cls._skip_nonitem_fields = skip_nonitem_fields
 
-    @validates_input
+    def _validate_input(self) -> None:
```
kmike, line 103: It seems it'd be a bit cleaner if FieldsMixin can handle a case when _validate_input is not defined

wRAR: Yeah, and it's both FieldsMixin and @field that need to handle it, I'll do that and see how it looks.

wRAR:
I've forgot to mention that all of this needs better docstrings now, especially to show the differences between classes.

web_poet/pages.py
```diff
@@ -137,3 +150,20 @@ class WebPage(ItemPage[ItemT], ResponseShortcutsMixin):
...
+
+@attr.s(auto_attribs=True)
+class SelectorExtractor(Extractor[ItemT]):
+    """Extractor that takes a :class:`parsel.Selector` and provides
+    XPath / CSS shortcuts.
+    """
+
+    selector: parsel.Selector
+
+    def xpath(self, query, **kwargs) -> parsel.SelectorList:
```
kmike, line 163: This code is exactly the same as in SelectableMixin. What do you think about creating CssXpathMixin, and using it both here and in SelectableMixin?

wRAR: "Shortcuts are copied from SelectableMixin, I don't think we need a separate mixin for those but it's possible :)" was about this :)

kmike: Yeah, let's do it :) The code is simple, but still, there is query and **kwargs, and typing, it seems it's worth it to avoid duplication in this case.

web_poet/pages.py
```diff
@@ -137,3 +150,20 @@ class WebPage(ItemPage[ItemT], ResponseShortcutsMixin):
 
 
 ItemWebPage = _create_deprecated_class("ItemWebPage", WebPage, warn_once=False)
+
+
+@attr.s(auto_attribs=True)
+class SelectorExtractor(Extractor[ItemT]):
```
kmike, line 156: We should have it documented somewhere; I think only having reference API docs and not mentioning it elsewhere is not enough.

kmike:
👍 It looks good to me overall @wRAR!

web_poet/pages.py
wRAR: This isn't as straightforward as I thought as `selector` is a property in `SelectableMixin` but an attribute in `SelectorExtractor` so I'll look into it later (unless we want to just silence mypy on this).

docs/page-objects/fields.rst
```diff
@@ -263,6 +265,65 @@ explicitly accessing or subclassing the ``Processors`` class:
...
+        @field(out=[str.strip])
+        def color(self):
+            return self.css(".name::text")
+
+You can also pass other data in addition to, or instead of, selectors, such as
+dictionaries with some data:
+
+.. code-block:: python
+
+    class VariantExtractor(Extractor):
```
kmike, line 319: attrs.define is missing

kmike (approved):

web_poet/pages.py
```diff
@@ -137,3 +147,20 @@ class WebPage(ItemPage[ItemT], ResponseShortcutsMixin):
...
+    """Extractor that takes a :class:`parsel.Selector` and provides
+    XPath / CSS shortcuts.
+    """
+
+    selector: parsel.Selector
+
+    def xpath(self, query, **kwargs) -> parsel.SelectorList:
+        """A shortcut to ``.selector.xpath()``."""
+        return self.selector.xpath(query, **kwargs)
```
kmike, line 162: Could you please add tests for these methods?

web_poet/pages.py
wRAR: I went with silencing mypy because even if we find a clean way to declare `selector: parsel.Selector` in `CssXpathMixin` we still need to redeclare it in `SelectorExtractor` because of `attrs` (unless there is a way around that too).

web_poet/mixins.py
```diff
@@ -5,7 +5,17 @@
 from w3lib.html import get_base_url
 
 
-class SelectableMixin(abc.ABC):
+class CssXpathMixin(abc.ABC):
```
kmike, line 8: Why does it inherit from abc.ABC?

web_poet/mixins.py
```diff
@@ -5,7 +5,17 @@
 from w3lib.html import get_base_url
 
 
-class SelectableMixin(abc.ABC):
+class CssXpathMixin:
```
Gallaecio, line 8: Could we go for a more generic name? (I’m thinking of [JMESPath support in the future](https://github.com/scrapy/parsel/pull/181)) e.g. `SelectorShortcutsMixin` would be in line with `ResponseShortcutsMixin`.

web_poet/pages.py
```diff
@@ -81,10 +82,9 @@ def _to_item(self, *args, **kwargs):
     return _to_item  # type: ignore[return-value]
 
 
-class ItemPage(Injectable, Returns[ItemT]):
-    """Base Page Object, with a default :meth:`to_item` implementation
-    which supports web-poet fields.
-    """
+class Extractor(Returns[ItemT], FieldsMixin):
+    """The base class for page objects and other extractors, providing support
+    for fields."""
```
Gallaecio, line 87: To make it clear that `ItemPage` is the base class for page objects, maybe it would be best not to mention page objects here.

```suggestion
    """Base class for field support."""
```

web_poet/pages.py
```diff
@@ -100,15 +100,20 @@ def __init_subclass__(cls, skip_nonitem_fields=_NOT_SET, **kwargs):
...
             self,
             item_cls=self.item_cls,
             skip_nonitem_fields=self._get_skip_nonitem_fields(),
         )
 
+
+class ItemPage(Extractor[ItemT], Injectable):
+    """Base Page Object, with a default :meth:`to_item` implementation
+    which supports web-poet fields.
+    """
```
Gallaecio, line 115: I would refrain mentioning the `to_item` implementation, since it basically comes from Extractor.

```suggestion
    """Base class for page objects."""
```

web_poet/pages.py
```diff
@@ -137,3 +147,12 @@ class WebPage(ItemPage[ItemT], ResponseShortcutsMixin):
 
 
 ItemWebPage = _create_deprecated_class("ItemWebPage", WebPage, warn_once=False)
+
+
+@attr.s(auto_attribs=True)
+class SelectorExtractor(Extractor[ItemT], CssXpathMixin):
+    """Extractor that takes a :class:`parsel.Selector` and provides
+    XPath / CSS shortcuts.
+    """
```
Gallaecio, line 156:
```suggestion
    """Extractor that takes a :class:`parsel.Selector` and provides shortcuts 
    for its methods."""
```

PR #163, opened 2023-04-11
Store HttpError exceptions in HttpClient serialization

wRAR:
This allows us to test requests that fail with HttpError, not just return a non-200 status.

### DIFF
--- a/tests/test_testing.py
+++ b/tests/test_testing.py
@@ -12,7 +12,7 @@ from itemadapter import ItemAdapter
 from zyte_common_items import Item, Metadata, Product
 
 from web_poet import HttpClient, HttpRequest, HttpResponse, WebPage, field
-from web_poet.exceptions import HttpResponseError
+from web_poet.exceptions import HttpRequestError, HttpResponseError
 from web_poet.page_inputs.client import _SavedResponseData
 from web_poet.testing import Fixture
 from web_poet.testing.__main__ import main as cli_main
@@ -336,7 +336,7 @@ def test_httpclient_no_response(pytester, book_list_html_response) -> None:
 
 
 @attrs.define
-class ClientExceptionPage(WebPage):
+class ClientResponseErrorPage(WebPage):
     client: HttpClient
 
     async def to_item(self) -> dict:
@@ -348,7 +348,7 @@ class ClientExceptionPage(WebPage):
         return {"foo": "bar", "exception": msg}
 
 
-def test_httpclient_exception(pytester, book_list_html_response) -> None:
+def test_httpclient_response_exception(pytester, book_list_html_response) -> None:
     url = "http://books.toscrape.com/1.html"
     request = HttpRequest(url)
     response = HttpResponse(url=url, body=b"body1", status=404, encoding="utf-8")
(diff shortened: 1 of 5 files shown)


web_poet/page_inputs/client.py
```diff
@@ -195,6 +196,10 @@ async def execute(
         if self.return_only_saved_responses:
             for fp, saved_data in self._saved_responses.items():
                 if request_fingerprint(request) == fp:
+                    if saved_data.exception:
+                        raise saved_data.exception
+                    if not saved_data.response:
+                        continue
```
Gallaecio, line 202: If there is no exception there is always going to be a response, right?

```suggestion
                    assert saved_data.response
```

web_poet/serialization/functions.py
```diff
@@ -115,9 +118,16 @@ def _serialize_HttpClient(o: HttpClient) -> SerializedLeafData:
...
         for k, v in serialized_request.items():
             serialized_data[f"{i}-HttpRequest.{k}"] = v
-        serialized_response = serialize_leaf(data.response)
-        for k, v in serialized_response.items():
-            serialized_data[f"{i}-HttpResponse.{k}"] = v
+        if data.response:
+            serialized_response = serialize_leaf(data.response)
+            for k, v in serialized_response.items():
+                serialized_data[f"{i}-HttpResponse.{k}"] = v
+        if data.exception:
```
Gallaecio, line 125: Would this be right?

```suggestion
        else:
            assert data.exception
```

web_poet/page_inputs/client.py
wRAR: Right.

web_poet/serialization/functions.py
wRAR: Well, assuming the data on-disk is well-formed.

web_poet/serialization/functions.py
```diff
@@ -128,24 +138,42 @@ def _deserialize_HttpClient(
...
         elif type_name == "HttpResponse":
             serialized_responses.setdefault(key, {})[suffix] = v
+        elif type_name == "exception_type":
+            serialized_exception_types[key] = v.decode()
+        elif type_name == "exception":
+            serialized_exceptions.setdefault(key, {})[suffix] = v
 
     for key, serialized_request in serialized_requests.items():
         serialized_response = serialized_responses.get(key)
-        if not serialized_response:
```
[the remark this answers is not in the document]
wRAR: I think it could only happen if the data on-disk is not well-formed.

Gallaecio (approved):
Looks good to me.

I do think at some point we might want to revisit how we handle reading bad fixtures, I think raising an exception may be appropriate in such cases, but I am OK with not addressing that here.

tests/test_testing.py
```diff
@@ -363,14 +363,55 @@ def test_httpclient_exception(pytester, book_list_html_response) -> None:
...
+        "exception": "Bad Request",
+    }
+    fixture = _save_fixture(
+        pytester,
+        page_cls=ClientRequestErrorPage,
+        page_inputs=[book_list_html_response, client],
+        expected=item,
+    )
+    assert (fixture.input_path / "HttpClient-0-exception_type.txt").exists()
+    assert (fixture.input_path / "HttpClient-0-exception.msg.txt").exists()
```
kmike, line 407: Would it be possible to store the exception in a single file?

wRAR: It's possible, though we will need to stop using `serialize_leaf` for that and do the serialization inside `_serialize_HttpClient` (as to use the leaf serialization API we need to store the type name somewhere).

And I wonder what should we do with exceptions that have the `request` attribute non-empty. Currently it's serialized via the normal serialization API and so it's stored as separate file(s). Options:

1. Continue doing that, so that an exception will have its own data in a file created directly inside `_serialize_HttpClient` and also some file(s) for its embedded request.
2. Write some code inside `_serialize_HttpClient` that serializes the request into a single dict to put it inside the exception JSON.
3. Don't serialize `exception.request` and on deserialization set it to the request that this exception is related to (only if a flag that says "exception.request wasn't empty" is true?).

It's actually something worth discussing in any case: currently after deserialization `request` and `exception.request` will be different objects, I don't know if this can cause problems. Also note that scrapy-poet doesn't set the `request` attribute when raising `HttpError`.

kmike: > It's possible, though we will need to stop using serialize_leaf for that and do the serialization inside _serialize_HttpClient

This looks ok to me, because exceptions are not leaf dependencies.

> And I wonder what should we do with exceptions that have the request attribute non-empty. Currently it's serialized via the normal serialization API and so it's stored as separate file(s).

A good catch. No strong opinion on that, but options 2 and 3 make more sense to me. (3) is probably fine.

> Also note that scrapy-poet doesn't set the request attribute when raising HttpError.

A good catch; looks like a bug?

kmike: It may even be fine to not store any exception arguments now, because we're only checking the exception type.

web_poet/page_inputs/client.py
kmike: Hey! What was the resolution here?

kmike (approved):
👍

web_poet/page_inputs/client.py
kmike: Is the idea to address it later?

PR #165, opened 2023-04-11
Cover best practices for item types

Gallaecio:
Follow-up on https://github.com/scrapinghub/web-poet/pull/152#discussion_r1152308278.

### DIFF
--- a/docs/page-objects/items.rst
+++ b/docs/page-objects/items.rst
@@ -23,3 +23,25 @@ class. For example:
 
 Because itemadapter_ allows implementing support for arbitrary types,
 any kind of Python object can potentially work as an item.
+
+Best practices for item types
+=============================
+
+To keep your code maintainable, we recommend you to:
+
+-   Reuse item types.
+
+    For example, if you want to extract product details data from 2 e-commerce
+    websites, try to use the same item type for both of them. Or at least try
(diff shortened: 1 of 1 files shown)


docs/page-objects/items.rst
```diff
@@ -23,3 +23,25 @@ class. For example:
...
+    For example, if you want to extract product details data from 2 e-commerce
+    websites, try to use the same item type for both of them. Or at least try
+    to define a base item type with shared fields, and only keep
+    website-specific fields in website-specific items.
+
+-   Keep item types as logic-free as possible.
+
+    For example, any parsing and field cleanup logic is better handled through
+    :ref:`page object classes <page-object-classes>`, e.g. using :ref:`field
+    processors <field-processors>`.
```
kmike, line 43: It may be worth explaining briefly why we are making this recommendation; it's not obvious, so people might think it's e.g. a stylistic choice to promote web-poet fields.

kmike (approved):
+1 to merge with or without addressing the comment.

docs/page-objects/items.rst
Gallaecio: I agree in explaining it, and I think we should do it in this PR, but I don’t think I fully know the explanation myself :sweat_smile:

wRAR (approved):

PR #166, opened 2023-04-12
Support for expected exceptions in tests

wRAR:
There are several things to discuss here, e.g. whether we want to check the exception args.

### DIFF
--- a/tests/test_testing.py
+++ b/tests/test_testing.py
@@ -66,11 +66,11 @@ class MyItemPage2(WebPage):
 
 
 def _save_fixture(
-    pytester, page_cls, page_inputs, *, expected=None, expected_exception=None
+    pytester, page_cls, page_inputs, *, expected_output=None, expected_exception=None
 ):
     base_dir = pytester.path / "fixtures" / get_fq_class_name(page_cls)
     return Fixture.save(
-        base_dir, inputs=page_inputs, item=expected, exception=expected_exception
+        base_dir, inputs=page_inputs, item=expected_output, exception=expected_exception
     )
 
 
@@ -79,7 +79,7 @@ def test_pytest_plugin_pass(pytester, book_list_html_response) -> None:
         pytester,
         page_cls=MyItemPage,
         page_inputs=[book_list_html_response],
-        expected={"foo": "bar"},
+        expected_output={"foo": "bar"},
     )
     result = pytester.runpytest()
     result.assert_outcomes(passed=3)
@@ -90,7 +90,7 @@ def test_pytest_plugin_bad_field_value(pytester, book_list_html_response) -> Non
         pytester,
         page_cls=MyItemPage,
         page_inputs=[book_list_html_response],
-        expected={"foo": "not bar"},
+        expected_output={"foo": "not bar"},
     )
     result = pytester.runpytest()
     result.assert_outcomes(failed=1, passed=2)
@@ -102,7 +102,7 @@ def test_pytest_plugin_bad_field_value_None(pytester, book_list_html_response) -
         pytester,
         page_cls=MyItemPage2,
         page_inputs=[book_list_html_response],
-        expected={"foo": "bar"},
+        expected_output={"foo": "bar"},
     )
     result = pytester.runpytest()
     result.assert_outcomes(failed=1, passed=2)
@@ -115,7 +115,7 @@ def test_pytest_plugin_missing_field(pytester, book_list_html_response) -> None:
         pytester,
         page_cls=MyItemPage,
         page_inputs=[book_list_html_response],
-        expected={"foo": "bar", "foo2": "bar2"},
+        expected_output={"foo": "bar", "foo2": "bar2"},
     )
     result = pytester.runpytest()
     result.assert_outcomes(failed=1, passed=3)
@@ -127,7 +127,7 @@ def test_pytest_plugin_extra_field(pytester, book_list_html_response) -> None:
         pytester,
         page_cls=MyItemPage,
         page_inputs=[book_list_html_response],
-        expected={"foo2": "bar2"},
+        expected_output={"foo2": "bar2"},
     )
     result = pytester.runpytest()
     result.assert_outcomes(failed=2, passed=1)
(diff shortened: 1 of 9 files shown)


wRAR:
For the wrong exception raised, I've tried two options with the existing `comparison_error_message()` and they are very different:

```
to_item() raised a wrong exception.
Expected: <class 'web_poet.exceptions.core.Retry'>, got: <class 'ValueError'>
```

```
to_item() raised a wrong exception.
'ValueError' == 'Retry'
- Retry
+ ValueError
```

The first option doesn't look clean to me, and I think in the second one the diff is unnecessary. We can also not use `comparison_error_message()` here and print someting simpler, like "to_item() raised a wrong exception: ValueError instead Retry" in one line.
Also we can use fully qualified class names instead of short names.

Thoughts?

wRAR:
We probably also want to show the traceback for the wrong exception? Not sure how to fit it here, maybe just after the message?

```
to_item() raised a wrong exception ZeroDivisionError instead of Retry.
Traceback (most recent call last):
[...]
  File "tests/test_testing.py", line 430, in to_item
    1/0
    ~^~

ZeroDivisionError: division by zero
```

Gallaecio:
+1 to something like “to_item() raised a wrong exception. Expected web_poet.exceptions.core.Retry, got ValueError”.

About the traceback, no preference, but I think we could even go with a regular Python exception chain, where the readable error is the last exception, hopefully with a short traceback, and above it there is the original exception traceback.

web_poet/testing/pytest.py
```diff
@@ -144,6 +153,38 @@ def reportinfo(self):
...
+        if isinstance(excinfo.value, WrongExceptionRaised):
+            got = excinfo.value.__cause__
+            inner_excinfo = pytest.ExceptionInfo.from_exc_info(
+                (type(got), got, got.__traceback__)
+            )
+            return (
+                f"to_item() raised a wrong exception. Expected"
+                f" {get_fq_class_name(type(expected))}, got"
+                f" {get_fq_class_name(type(got))}.\n\n"
+                + str(super().repr_failure(inner_excinfo, style))
```
wRAR, line 183: pytest code doesn't allow us to make a report in the way I wanted so here is a custom "header" and the traceback from pytest converted to a string (which I think only removes colors and an exception name after "test-1::TO_ITEM_RAISES" in the "short test summary info"). I think formatting the traceback manually doesn't bring us anything but loses the ability to customize it via `--tb=`.

kmike:
The code looks good to me. 

How does the final output look like?

wRAR:
```
_________________________________________________________ tests.test_testing.RetryItemPage/test-1: to_item raises Retry __________________________________________________________
to_item() raised a wrong exception. Expected web_poet.exceptions.core.Retry, got builtins.ZeroDivisionError.

self = <web_poet.testing.fixture.Fixture object at 0x7f115b9f0650>

    def assert_toitem_exception(self) -> None:
        """Assert that to_item() raises an exception of the expected type"""
        try:
>           self.get_output()

web_poet/testing/fixture.py:232: 
[...]

self = RetryItemPage(response=HttpResponse(url=ResponseUrl('http://books.toscrape.com/index.html'), body=b'\n<!DOCTYPE html>\...    <!-- Version: N/A -->\n\n    </body>\n</html>\n', status=None, headers=<HttpResponseHeaders()>, _encoding='utf-8'))

    async def to_item(self):
>       1/0
E       ZeroDivisionError: division by zero

tests/test_testing.py:429: ZeroDivisionError
============================================================================ short test summary info =============================================================================
FAILED fixtures/tests.test_testing.RetryItemPage::test-1::TO_ITEM_RAISES
=============================================================================== 1 failed in 0.21s ================================================================================
```


```
_________________________________________________________ tests.test_testing.RetryItemPage/test-1: to_item raises Retry __________________________________________________________
to_item() didn't raise an exception. web_poet.exceptions.core.Retry was expected.
============================================================================ short test summary info =============================================================================
FAILED fixtures/tests.test_testing.RetryItemPage::test-1::TO_ITEM_RAISES
=============================================================================== 1 failed in 0.08s ================================================================================
```

Gallaecio (approved):

web_poet/testing/fixture.py
```diff
@@ -232,11 +262,19 @@ def save(
...
         if meta:
             with fixture.meta_path.open("w") as f:
                 json.dump(meta, f, ensure_ascii=False, indent=4)
 
+        if exception:
+            with fixture.exception_path.open("w") as f:
+                exc_data = {
+                    "type_name": get_fq_class_name(type(exception)),
+                }
+                json.dump(exc_data, f, ensure_ascii=False, indent=4)
```
kmike, line 278: Is there a reason we have two ways/formats to store an exception - here and in https://github.com/scrapinghub/web-poet/pull/163/files#diff-fddeae0fca9e0d2911f103f05ecd5c10af4494f7b24fad04b214aae96427128cR127? Would it make sense to refactor it?

tests/test_testing.py
```diff
@@ -65,9 +65,13 @@ async def to_item(self) -> dict:
         return {"foo": None}
 
 
-def _save_fixture(pytester, page_cls, page_inputs, expected):
+def _save_fixture(
+    pytester, page_cls, page_inputs, *, expected=None, expected_exception=None
```
kmike, line 69: expected -> expected_output?

web_poet/testing/fixture.py
```diff
@@ -232,11 +262,17 @@ def save(
...
+                f.write(cls.item_to_json(item))
 
         if meta:
             with fixture.meta_path.open("w") as f:
                 json.dump(meta, f, ensure_ascii=False, indent=4)
 
+        if exception:
+            with fixture.exception_path.open("w") as f:
+                exc_data = _exception_to_dict(exception)
+                json.dump(exc_data, f, ensure_ascii=False, indent=4)
```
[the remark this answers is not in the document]
wRAR: What do you think about making a function that does this? I think we have quite a lot of places that do this already.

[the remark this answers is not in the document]
kmike: +1 to add a function

kmike:
Thanks @wRAR!

PR #167, opened 2023-04-19
jmespath shortcut

kmike:
I opted in for a runtime check to simplify the implementation a bit. We can drop it after bumping min parsel version requirement.

### DIFF
--- a/web_poet/mixins.py
+++ b/web_poet/mixins.py
@@ -26,7 +26,8 @@ class SelectorShortcutsMixin:
 class SelectableMixin(abc.ABC, SelectorShortcutsMixin):
     """
     Inherit from this mixin, implement ``._selector_input`` method,
-    get ``.selector`` property and ``.xpath`` / ``.css`` methods.
+    get ``.selector`` property and ``.xpath`` / ``.css`` / "``.jmespath``
+    methods.
     """
 
     __cached_selector = None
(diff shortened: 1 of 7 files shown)


Gallaecio:
I wonder if we should use a try-except here instead, even if it won’t make much of a difference.

web_poet/mixins.py
```diff
@@ -14,6 +14,14 @@ def css(self, query) -> parsel.SelectorList:
         """A shortcut to ``.selector.css()``."""
         return self.selector.css(query)  # type: ignore[attr-defined]
 
+    def jmespath(self, query: str, **kwargs) -> parsel.SelectorList:
+        """A shortcut to ``.selector.jmespath()``."""
+        if not hasattr(self.selector, "jmespath"):  # type: ignore[attr-defined]
+            raise AttributeError(
+                "Please install parsel >= 1.8.1 to get jmespath support"
+            )
+        return self.selector.jmespath(query, **kwargs)  # type: ignore[attr-defined]
```
kmike, line 23: @Gallaecio 

> I wonder if we should use a try-except here instead, even if it won’t make much of a difference.

It seems it'd be a bit more code, with the same result.
 ```suggestion
        selector = self.selector
        try:
            jmespath = selector.jmespath
        except AttributeError:
            raise AttributeErrror("please install...")
        return jmespath(query, **kwargs)
```

Gallaecio (approved):

PR #169, opened 2023-04-24
Refactor the rules documentation

Gallaecio:
My goal was to greatly simplify https://web-poet.readthedocs.io/en/stable/page-objects/rules.html.

However, I wonder if I went too far; the new pages are ~25% of the old page in terms of line count.

I removed much of the content, and although I aimed to keep the core information around, I did not always do that. For example, the [entire last section](https://web-poet.readthedocs.io/en/stable/page-objects/rules.html#using-rules-from-external-packages) is gone without its information making it in any form into the new page; I wonder if we should keep some of that information.

I recommend reading the old page, then the new pages, and see how you feel about the missing information.

### DIFF
--- a/docs/frameworks/rules.rst
+++ b/docs/frameworks/rules.rst
@@ -18,9 +18,6 @@ a page object based on rules:
 
     page_cls = default_registry.page_cls_for_item("https://example.com", MyItem)
 
-To give your users more flexibility, you can also let them configure an
-alternative :class:`~.RulesRegistry` object to be used for rule resolution.
-
 You should also let your users know what is the best approach to :ref:`load
 rules <load-rules>` when using your framework. For example, let them know the
 best location for their calls to the :func:`~.web_poet.rules.consume_modules`
--- a/docs/frameworks/rules.rst
+++ b/docs/frameworks/rules.rst
@@ -1,0 +1,27 @@
+.. _framework-rules:
+
+================
+Supporting rules
+================
+
+Ideally, a framework should support returning the right :ref:`page object
+<page-objects>` or :ref:`output item <items>` given a target URL and a desired
+:ref:`output item type <items>` when :ref:`rules <rules>` are used.
+
+To provide basic support for rules in your framework, use the
+:class:`~.RulesRegistry` object at ``web_poet.default_registry`` to choose
+a page object based on rules:
+
+.. code-block:: python
+
+    from web_poet import default_registry
+
+    page_cls = default_registry.page_cls_for_item("https://example.com", MyItem)
+
+To give your users more flexibility, you can also let them configure an
+alternative :class:`~.RulesRegistry` object to be used for rule resolution.
+
+You should also let your users know what is the best approach to :ref:`load
+rules <load-rules>` when using your framework. For example, let them know the
+best location for their calls to the :func:`~.web_poet.rules.consume_modules`
+function.
--- a/docs/frameworks/rules.rst
+++ b/docs/frameworks/rules.rst
@@ -6,7 +6,7 @@ Supporting rules
 
 Ideally, a framework should support returning the right :ref:`page object
 <page-objects>` or :ref:`output item <items>` given a target URL and a desired
-:ref:`output item type <items>` when :ref:`rules <rules>` are used.
+:ref:`output item class <items>` when :ref:`rules <rules>` are used.
 
 To provide basic support for rules in your framework, use the
 :class:`~.RulesRegistry` object at ``web_poet.default_registry`` to choose
--- a/docs/intro/from-ground-up.rst
+++ b/docs/intro/from-ground-up.rst
@@ -181,7 +181,21 @@ can then refactor your ``BookPage`` class as follows:
     class BookPage(WebPage):
         def to_item(self) -> dict:
             return {
-                "url": self..url,
+                "url": self.response.url,
+                "title": self.response.css("h1").get(),
+            }
+
+:class:`~.WebPage` even provides shortcuts for some response attributes and
+methods:
+
+.. code-block:: python
+
+    from web_poet import WebPage
+
+    class BookPage(WebPage):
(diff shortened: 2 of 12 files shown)


docs/frameworks/rules.rst
```diff
@@ -0,0 +1,27 @@
...
+a page object based on rules:
+
+.. code-block:: python
+
+    from web_poet import default_registry
+
+    page_cls = default_registry.page_cls_for_item("https://example.com", MyItem)
+
+To give your users more flexibility, you can also let them configure an
+alternative :class:`~.RulesRegistry` object to be used for rule resolution.
```
kmike, line 22: Do you have a use case in mind?

wRAR:
Not sure if completely removing docs is a good idea in general but I've found the old version of the page to be very complicated, especially for someone who is not familiar with scrapy-poet (when reading it back then) yet so trimming it down is welcome.

wRAR (approved):

docs/page-objects/rules.rst
```diff
@@ -1,782 +1,166 @@
...
-===========
+=====
+Rules
+=====
 
-Basic Usage
------------
+Rules are :class:`~.ApplyRule` objects that tell web-poet which :ref:`page
+object class <page-object-classes>` to use for a given combination of input URL
+and :ref:`item type <item-types>`.
```
kmike, line 9: Technically, instead of item type one may be using just "instead_of" feature. Is it intentional that this is downplayed/removed from the updated docs?

docs/page-objects/rules.rst
```diff
@@ -1,782 +1,166 @@
...
-    # page object can be applied on any URL from the example.com domain,
-    # or from any of its subdomains
-    @handle_urls("example.com")
-
-    # page object can be applied on example.com pages under /products/ path
-    @handle_urls("example.com/products/")
-
-    # page object can be applied on any URL from example.com, but only if
-    # it contains "productId=..." in the query string
-    @handle_urls("example.com?productId=*")
```
kmike, line 70: I may be wrong, but it seems that having these basic examples cover almost all use cases, without asking users to check the external docs (which e.g. talk about some proxies in the examples, not page objects).

docs/page-objects/rules.rst
```diff
@@ -1,782 +1,166 @@
...
+
     from my_items import MyItem
 
     @handle_urls("example.com")
     class MyPage(ItemPage[MyItem]):
-        # ...
-
-
-``handle_urls("example.com")`` can serve as documentation, but it also enables
-getting the information about page objects programmatically.
```
kmike, line 30: This might not be the best way to explain it, but I think the new docs lack some information about why would anyone want to use rules.

docs/page-objects/rules.rst
Gallaecio: I keep forgetting that resolving a URL + item type is not a framework requirement.

docs/page-objects/rules.rst
Gallaecio: I have restored these 2 advantages in the page introduction, and covered how rules are needed to get item objects directly or to implement generic pave object classes with website-specific overrides.

docs/intro/from-ground-up.rst
```diff
@@ -181,8 +181,8 @@ can then refactor your ``BookPage`` class as follows:
     class BookPage(WebPage):
         def to_item(self) -> dict:
             return {
-                "url": self.response.url,
-                "title": self.response.css("h1").get(),
+                "url": self..url,
+                "title": self.css("h1").get(),
```
kmike, line 185: ..url is a typo.

I'm also not sure that the new example illustrates what's written before that: we explain that WebPage defines response argument, but then the example doesn't feature response.

docs/page-objects/rules.rst
```diff
@@ -5,8 +5,17 @@ Rules
 =====
 
 Rules are :class:`~.ApplyRule` objects that tell web-poet which :ref:`page
-object class <page-object-classes>` to use for a given combination of input URL
-and :ref:`item type <item-types>`.
+object class <page-object-classes>` to use based on user input, i.e. the target
+URL and an output class (a :ref:`page object class <page-object-classes>` or an
+:ref:`item class <items>`).
+
+Rules are necessary to enable using an item class as output class, and can also
```
kmike, line 12: Hm, I'm not sure it's clear here what "enable using an item class as output class" means

docs/page-objects/rules.rst
```diff
@@ -164,3 +203,42 @@ point of your code for that:
...
+    with conflicting rules, with a default priority (500), and you want
+    ``package1.A`` to take precedence, declare a new page object class as
+    follows:
+
+    .. code-block:: python
+
+        from package1 import A
+        from web_poet import handle_urls
+
+        @handle_urls(..., priority=501)
```
kmike, line 236: should we use priority=510, BASIC-style? :)

docs/intro/from-ground-up.rst
Gallaecio: I’ll revert. I was thinking we mention helper methods but then don’t show them used, but showing response here is probably more important.

docs/page-objects/rules.rst
Gallaecio: https://github.com/scrapinghub/web-poet/pull/169/commits/7b9c39a439cbdff5eefa164bead0ba5e22fb1310 ?

docs/intro/from-ground-up.rst
Gallaecio: https://github.com/scrapinghub/web-poet/pull/169/commits/28634d14ac80a46d64628a4a5777fdddcb35b029 ?

kmike:
Hey @Gallaecio! These docs look much better than before, so I'm merging it. I still think though that we should improve the introduction: the "output class" thing doesn't look clear to me, but I don't have a concrete suggestion.

PR #171, opened 2023-04-26
Support custom item adapters in fixtures

wRAR:


docs/page-objects/testing.rst
```diff
@@ -322,3 +322,42 @@ The coverage for page object code is reported correctly if tools such as
...
+
+    class MyAdapter(DictAdapter):
+        # any needed customization
+        ...
+
+    class MyItemAdapter(ItemAdapter):
+        ADAPTER_CLASSES = deque([MyAdapter])
+
+You can then put the fully qualified name of ``MyItemAdapter`` into
+``adapter_type_name`` and it will be used by the testing framework.
```
kmike, line 355: Are we using "type name" elsewhere? For MyItemAdapter I'd expect type name to be MyItemAdapter. In Scrapy docs we use "import path" name, see https://docs.scrapy.org/en/latest/topics/settings.html#import-paths-and-classes. I guess this applies both to docs and to the meta key names.

docs/page-objects/testing.rst
```diff
@@ -322,3 +322,42 @@ The coverage for page object code is reported correctly if tools such as
 `coverage`_ are used when running web-poet tests.
 
 .. _coverage: https://coverage.readthedocs.io/
+
+Item adapters
```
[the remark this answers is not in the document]
wRAR: If you mean `ItemAdapter(page_object.to_item()).asdict()` then I guess we need to replace this code with some description? As we can't show a one-liner anymore.

docs/page-objects/testing.rst
wRAR: The serialized exception has "type_name", though it's less user-visible. `adapter_import_path` sounds fine.

kmike: we might change type_name there as well; backwards compat should be easy to achieve, and it's maybe not even necessary - release a new version, move on :)

docs/page-objects/testing.rst
```diff
@@ -322,3 +322,42 @@ The coverage for page object code is reported correctly if tools such as
...
+=============
+
+The testing framework uses the itemadapter_ library to convert items to dicts
+when storing them in fixtures and when comparing the expected and the actual
+output. As adapters may influence the resulting dicts, it's important to use
+the same adapter in both cases and it may be useful to use a different adapter
+in tests and in production (for example, you may want to omit empty fields in
+production but be able to distinguish between empty and absent fields in
+tests). For this you can set the ``adapter_type_name`` field in the metadata
```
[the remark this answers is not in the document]
Gallaecio: I would go even further and move the last sentence to its own paragraph as well, however short.

docs/page-objects/testing.rst
```diff
@@ -322,3 +322,42 @@ The coverage for page object code is reported correctly if tools such as
...
+dictionary to the type name of a class that inherits from
```
Gallaecio, line 336: I think both here and in the section about `frozen_time` we should replace “the metadata dictionary” with “the metadata dictionary passed to Fixture.save()”, to give more of a hint on how to send that metadata dictionary. We could alternative link “metadata dictionary” to the Fixtures section above that explains that it is passed to Fixture.save().

docs/page-objects/testing.rst
Gallaecio: Would it make sense to call it `adapter` and expect the class object itself? I imagine scrapy-poet allowing to set this with an import path string in a setting, but scrapy-poet could load the class before passing it to web-poet.

PR #175, opened 2023-05-04
Implement Stats

Gallaecio:


wRAR (approved):

web_poet/page_inputs/stats.py
```diff
@@ -0,0 +1,17 @@
+from typing import Any
+
+
+class Stats:
```
kmike, line 4: I think we need an abc for this, because providers are not going to be providing this exact class. For example, scrapy-poet should provide another object, which writes to Scrapy stats.

web_poet/page_inputs/stats.py
```diff
@@ -0,0 +1,18 @@
+from abc import ABC
+from typing import Any
+
+
+class Stats(ABC):
+    """Stat collector."""
+
+    def __init__(self):
+        self._stats = {}
```
kmike, line 9: Another thought, sorry :) May we should drop the concrete implementation, and only provide abc. The reason is that stats is a global state, so you always need the framework to provide it; Stats object should hold a reference to some global state, and change it in set/inc methods. It doesn't seem frameworks can reuse the implementation which we have here.

Gallaecio:
OK, `Stats` is abstract now. What do we do about serialization for test fixtures? Is that something for frameworks to handle on their own?

kmike:
> OK, Stats is abstract now. What do we do about serialization for test fixtures? Is that something for frameworks to handle on their own?

I'm not sure. Stats are write-only, so it seems the Stats object itself shouldn't affect the output of the page object in any way, unless there is some bug/exception. 

We need to ensure though that the stats class is passed to the page object during the test, and we can't rely on a framework like scrapy-poet to do so, because the tests shouldn't depend on it.

Maybe create something like DummyStats, and pass it during the test running? So, to serialize do nothing (or mark as something empty), and to deserialize return DummyStats? 

@wRAR any feedback on this?

Gallaecio:
I think we may want to *test* stat writing, i.e. make sure the same stats are written by the page object as when the text fixture was generated (well, the final stats, replacing inc + inc with inc(2) should not break a test I think).

kmike:
> I think we may want to test stat writing

We can do it, but it'd be quite a big shift from what the current framework tests. You'd need to have some special support for this particular dependency in the testing framework, and generate a new test item. 

I'd consider it a separate feature, and implement it only after the usefulnesss is clear. E.g. is it good or bad that your tests start to fail when you add some new stats to the page object? What kind of bugs do we try to prevent?

Gallaecio:
> Maybe create something like DummyStats, and pass it during the test running? So, to serialize do nothing (or mark as something empty), and to deserialize return DummyStats?

I have tried 2 approaches. [Serializing as `{}`](https://github.com/scrapinghub/web-poet/pull/175/commits/e233bceb4742da9ca627c93b4279004dc1686440) and [not serializing at all](https://github.com/scrapinghub/web-poet/pull/175/files/ad1eb817561caa0a3ac0bbbe895df30769a11618..5756bf73dcdcb86a28dd85656bb050cc9b742c6e). I definitely prefer the latter from a user perspective, but I have some doubts about my specific implementation.

web_poet/page_inputs/stats.py
```diff
@@ -0,0 +1,17 @@
+from abc import abstractmethod
+from typing import Any
+
+from web_poet.pages import Injectable
+
+
+class Stats(Injectable):
```
Gallaecio, line 7: This is the only page input that inherits from `Injectable`. I wonder if I am missing some potential issue by doing that.

web_poet/serialization/api.py
```diff
@@ -189,5 +191,9 @@ def deserialize(cls: Type[InjectableT], data: SerializedData) -> InjectableT:
     for fn_or_cls, kwargs_spec in plan[:-1]:
         if fn_or_cls in externally_provided:
             continue
-        deps[fn_or_cls] = fn_or_cls(**kwargs_spec.kwargs(deps))
+        fn_or_cls = cast(Type[Injectable], fn_or_cls)
+        if fn_or_cls in DUMMY_MAP:
+            deps[fn_or_cls] = DUMMY_MAP[fn_or_cls]()
+            continue
+        deps[fn_or_cls] = fn_or_cls(**kwargs_spec.kwargs(deps))  # type: ignore
```
Gallaecio, line 198: mypy throws `web_poet/serialization/api.py:198: error: <nothing> not callable  [misc]` after I added `cast` above, which I added to prevent another complaint from mypy on `DUMMY_MAP[fn_or_cls]`.

web_poet/page_inputs/stats.py
wRAR: This would cause problems in `deserialize()` but you already have specoial handling for this class there.

Maybe it could cause similar problems in other places that use `andi` and need to determine which deps are leaf deps?

wRAR (approved):

web_poet/page_inputs/stats.py
kmike: Why do you need to make it Injectable? Would it make sense to add a comment about it somewhere in the code?

web_poet/page_inputs/stats.py
```diff
@@ -0,0 +1,17 @@
...
+    """Stat collector."""
+
+    @abstractmethod
+    def set(self, key: str, value: Any) -> None:
+        """Sets the value of stat *key* to *value*."""
+
+    @abstractmethod
+    def inc(self, key: str, value: int = 1) -> None:
+        """Increments the value of stat *key* by *value*, or sets it to *value*
+        if *key* has no value."""
```
kmike, line 17: Another idea (not sure if it's good). It seems we expect frameworks to return custom Stats objects, not instances of this class in particular. And that's different from all other dependencies, including HttpClient dependency. It seems this is what causes problems when trying to support it in tests.

Would it make sense to try another approach, where Stats dependency is not an abc, but there is a customizable stats storage, similar to how the concrete request downloader is customizable by HttpClient?

Gallaecio: Applied, along with the [“serialize to nothing, return dummy stats” approach discussed earlier](https://github.com/scrapinghub/web-poet/pull/175#issuecomment-1535973215).

web_poet/page_inputs/stats.py
```diff
@@ -0,0 +1,40 @@
+from typing import Any, Dict, Protocol
+
+
+class StatCollectorProtocol(Protocol):
```
kmike, line 4: It seems it'd be easier to have an abc for stats collector, not a protocol. 
Also, it seems the protocol is currently unused?

web_poet/page_inputs/stats.py
kmike: Nice! Looks much more straightforward.

web_poet/page_inputs/stats.py
```diff
@@ -0,0 +1,40 @@
...
+    def inc(self, key: str, value: int = 1) -> None:  # noqa: D102
+        if key in self._stats:
+            assert isinstance(self._stats[key], int)
+            self._stats[key] += value
+        else:
+            self._stats[key] = value
+
+
+class Stats:
+    """Stat collector."""
```
kmike, line 28: It'd be nice to document it a bit more, at least the the stat_collector approach. It seems "frameworks" section is the right place for it.

web_poet/page_inputs/stats.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def __init__(self, stat_collector=None):
+        self._stats = stat_collector or DummyStatCollector()
+
+    def set(self, key: str, value: Any) -> None:
+        """Sets the value of stat *key* to *value*."""
```
kmike, line 34:
```suggestion
        """Set the value of stat *key* to *value*."""
```

web_poet/page_inputs/stats.py
```diff
@@ -0,0 +1,40 @@
...
+        self._stats.set(key, value)
+
+    def inc(self, key: str, value: int = 1) -> None:
+        """Increments the value of stat *key* by *value*, or sets it to *value*
```
kmike, line 38:
```suggestion
        """Increment the value of stat *key* by *value*, or sets it to *value*
```

web_poet/page_inputs/stats.py
```diff
@@ -0,0 +1,40 @@
...
+class DummyStatCollector:
+    def __init__(self):
+        self._stats: Dict[str, Any] = {}
+
+    def set(self, key: str, value: Any) -> None:  # noqa: D102
+        self._stats[key] = value
+
+    def inc(self, key: str, value: int = 1) -> None:  # noqa: D102
+        if key in self._stats:
+            assert isinstance(self._stats[key], int)
```
kmike, line 21: Should we support float increments?

kmike: Though I'm not sure.. Let's not support floats here for now :)

kmike:
Looks great, thanks @Gallaecio!

PR #176, opened 2023-05-28
BrowserResponse dependency

kmike:
Here BrowserResponse is added, similar to HttpResponse. The issue with BrowserHtml is that it doesn't allow to access the URL and status code of the response.

I haven't added headers for BrowserResponse for now; it seems we can add them later, as they'd be optional, like in HttpResponse.

### DIFF
--- a/docs/page-objects/inputs.rst
+++ b/docs/page-objects/inputs.rst
@@ -59,9 +59,6 @@ define as inputs for a page object class, including:
 -   :class:`~web_poet.page_inputs.page_params.PageParams`, to receive data from
     the crawling code.
 
--   :class:`~web_poet.page_inputs.browser.BrowserHtml`, an HTML rendering of
-    the `Document Object Model`_ of a web page.
-
 -   :class:`~web_poet.page_inputs.browser.BrowserResponse`, which includes URL,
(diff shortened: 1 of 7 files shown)


docs/page-objects/inputs.rst
```diff
@@ -62,6 +62,10 @@ define as inputs for a page object class, including:
 -   :class:`~web_poet.page_inputs.browser.BrowserHtml`, an HTML rendering of
     the `Document Object Model`_ of a web page.
 
+-   :class:`~web_poet.page_inputs.browser.BrowserResponse`, which includes URL,
+    status code and :class:`~web_poet.page_inputs.browser.BrowserHtml`
+    of a rendered web page.
```
Gallaecio, line 67: What about replacing rather than adding it to the list?

```suggestion
-   :class:`~web_poet.page_inputs.browser.BrowserResponse`, which includes URL,
    status code and :class:`~web_poet.page_inputs.browser.BrowserHtml`
    of a rendered web page.
```

Gallaecio (approved):

docs/page-objects/inputs.rst
kmike: Makes sense!

wRAR (approved):

PR #178, opened 2023-05-30
support non-standard status codes

kmike:


web_poet/page_inputs/client.py
```diff
@@ -265,3 +265,18 @@ async def batch_execute(
...
         """Return saved requests and responses."""
         return self._saved_responses.values()
+
+
+def _http_status_name(status: int):
+    """
+    >>> _http_status_name(200)
+    'OK'
+    >>> _http_status_name(404)
+    'NOT_FOUND'
```
kmike, line 275: There is an option to use "phrase" instead of "name", to make it "Not Found" instead of "NOT_FOUND"; I haven't changed it in this PR.

Gallaecio (approved):

PR #180, opened 2023-06-06
Set the minimum packaging version.

wRAR:
[`Version.major` is added in 20.0.](https://packaging.pypa.io/en/latest/changelog.html#id17)


kmike:
👍

PR #183, opened 2023-07-28
Drop Python 3.7 support

wRAR:


tests/test_serialization.py
```diff
@@ -72,7 +72,7 @@ class ResponseData(Injectable):
 
     @attrs.define
     class MyWebPage(WebPage):
-        url: ResponseUrl
+        url_: ResponseUrl
```
wRAR, line 75: It turned out `WebPage.url` already exists, as a shortcut property.

web_poet/mixins.py
```diff
@@ -78,26 +79,23 @@ def urljoin(self, url: Union[str, RequestUrl, ResponseUrl]) -> RequestUrl:
         return RequestUrl(urljoin(self._base_url, str(url)))
 
 
-# TODO: when dropping Python 3.7 support,
-# fix untyped ResponseShortcutsMixin.response using typing.Protocol
```
wRAR, line 82: I don't see how does this need a protocol.

tests/test_serialization.py
kmike: Would it make sense to use ItemPage instead? It looks a bit strange to define 2 url attributes, with similar names.

wRAR:
> RuntimeWarning: Conversion of MetadataLocalTime(dateDownloaded='2022-51-05T03:51:22Z', probability=1.0, searchText=None, dateDownloadedLocal='2022-51-05T03:51:22+0000') into <class 'zyte_common_items.components.ProductMetadata'> is dropping the non-default values of the following attributes: {'dateDownloadedLocal'}.

Ouch. I wonder if we can fix this without depending on newer zyte-common-items, though it may be fine as the test is dep-only.

web_poet/mixins.py
kmike: The idea was to fail the type check if ResponseShortcutsMixin is added to a class without response attribute. But no strong opinion on that.

Gallaecio (approved):

kmike:
Thanks @wRAR!

PR #184, opened 2023-07-31
Add get_generic_param() and fix finding Returns[]

### LINKED ISSUE #182: Returns doesn't work in some subclasses
Similar to https://github.com/zytedata/zyte-common-items/issues/49, even though the code is different, it's also not recursive and fails on e.g. this:

```python
def test_returns_inheritance() -> None:
    @attrs.define
    class MyItem:
        name: str

    class BasePage(ItemPage[MyItem]):
        @field
        def name(self):
            return "hello"

    MetadataT = TypeVar("MetadataT")

    class HasMetadata(Generic[MetadataT]):
        pass

    class DummyMetadata:
        pass

    class Page(BasePage, HasMetadata[DummyMetadata]):
        pass

    page = Page()
    assert page.item_cls is MyItem
```

wRAR:
Fixes #182 and will help in https://github.com/zytedata/zyte-common-items/issues/49


Gallaecio (approved):

web_poet/pages.py
```diff
@@ -35,25 +39,40 @@ class Injectable(abc.ABC, FieldsMixin):
...
 
 
-class Returns(typing.Generic[ItemT]):
+class Returns(Generic[ItemT]):
     """Inherit from this generic mixin to change the item class used by
     :class:`~.ItemPage`"""
 
     @property
-    def item_cls(self) -> typing.Type[ItemT]:
+    def item_cls(self) -> type:
```
kmike, line 56: What's the reason for this change? Are there cases where a class which inherits from Returns[ItemType] doesn't return ItemType from item_cls?

kmike: Or was it not working properly?

kmike (approved):
👍 👍

web_poet/pages.py
wRAR: It can return a dict, when the page does not actually inherit from `Returns[ItemType]` but just from `Returns[ItemT]`.

kmike: It seems we should have added a couple of tests for that in https://github.com/scrapinghub/web-poet/blob/master/tests_typing/test_item_page.mypy-testing. We have the same issue with dict for other cases, like to_item. Also, Returns doesn't work properly if a type is changed in a subclass. But for a common case of not using dicts and not changing the type it should work correctly, as tests show; it seems this change might break it. I wonder if is it a right way to think about it or not :)

wRAR: We have a test that makes sure the method returns a dict: https://github.com/scrapinghub/web-poet/blob/90eba4b94614b16a75821dfbb3223b5f5d602a77/tests/test_pages.py#L119

wRAR: Do you want type hints here to say that this method returns `ItemType` when the class inherits from `Returns[ItemType]`?

kmike: Yes, I think that was the intention before the change (but no test in mypy-testing). 
It's not a big deal overall, because this property should rarely be used by the end users though.

kmike: > We have a test that makes sure the method returns a dict

It's not a typing test (see https://github.com/scrapinghub/web-poet/tree/master/tests_typing); from the code point of view, it all should work fine.

wRAR: Isn't https://github.com/scrapinghub/web-poet/blob/90eba4b94614b16a75821dfbb3223b5f5d602a77/tests_typing/test_item_page.mypy-testing#L24 about it?

wRAR: Well, it and the next one (though I'm not sure why does the next one still pass).

kmike: Yes, it's a right test; it is marked as xfail.

kmike:
Thanks @wRAR!

PR #185, opened 2023-07-31
Justify recommending logic-free items

Gallaecio:
> It may be worth explaining briefly why we are making this recommendation; it's not obvious, so people might think it's e.g. a stylistic choice to promote web-poet fields. — https://github.com/scrapinghub/web-poet/pull/165#discussion_r1163190244

### DIFF
--- a/docs/page-objects/items.rst
+++ b/docs/page-objects/items.rst
@@ -45,9 +45,7 @@ To keep your code maintainable, we recommend you to:
     Having code that makes item field values different from their counterpart
     page object field values can subvert the expectations of users of your
     code, which might need to access page object fields directly, for example
-    for `field subset selection`_.
-
-    .. _field subset selection: https://github.com/scrapinghub/web-poet/issues/115
+    for field subset selection.
 
 If you are looking for ready-made item classes, check out `zyte-common-items`_.
 
--- a/docs/page-objects/items.rst
+++ b/docs/page-objects/items.rst
@@ -42,6 +42,12 @@ To keep your code maintainable, we recommend you to:
     :ref:`page object classes <page-object-classes>`, e.g. using :ref:`field
     processors <field-processors>`.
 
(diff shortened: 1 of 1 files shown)


docs/page-objects/items.rst
```diff
@@ -42,6 +42,12 @@ To keep your code maintainable, we recommend you to:
     :ref:`page object classes <page-object-classes>`, e.g. using :ref:`field
     processors <field-processors>`.
 
+    Having code that makes item field values different from their input values
+    could hinder the use of future web-poet features, such as `field subset
+    selection`_.
```
kmike, line 47: I was thinking mostly about current features: if you have a page object, and `value = await page.field` is not the same as `item = await page.to_item(); value = item.field`, then it could lead to issues in practice, when people are expecting a different behavior. Accessing `page.field` is the current way to do the field subset selection.

Gallaecio: https://github.com/scrapinghub/web-poet/pull/185/commits/421e83874c490ac8baf9be1c0673ab9255f1ff9b ?

docs/page-objects/items.rst
```diff
@@ -42,6 +42,13 @@ To keep your code maintainable, we recommend you to:
     :ref:`page object classes <page-object-classes>`, e.g. using :ref:`field
     processors <field-processors>`.
 
+    Having code that makes item field values different from their counterpart
+    page object field values can subvert the expectations of users of your
+    code, which might need to access page object fields directly, for example
+    for `field subset selection`_.
+
+    .. _field subset selection: https://github.com/scrapinghub/web-poet/issues/115
```
kmike, line 50: Still, I'd avoid linking from the docs to open issues. We don't need any additional features to be implemented in web-poet to justify this suggestion, as users can already use individual fields of the page objects.

PR #193, opened 2023-12-18
Refactor the field docs

Gallaecio:


docs/page-objects/items.rst
```diff
@@ -24,11 +25,46 @@ class. For example:
...
+between brackets:
+
+.. code-block:: python
+
+    @attrs.define
+    class MyPage(ItemPage[MyItem]):
+        ...
+
+To change the item class of a subclass that does not directly inherit from
+:class:`~.ItemPage`, use :class:`~.Returns`:
```
kmike, line 41: If I'm not mistaken, technically you still can annotate any ItemPage subclass which hasn't been annotated with an item yet. For example, `WebPage[MyItem]` should work.

docs/page-objects/fields.rst
```diff
@@ -4,161 +4,312 @@
...
+Alternatively, you can consider :ref:`composition <composition>` for renaming
+fields. Composition is more verbose than subclassing, because you need to
+define every field in your page object class, but it can catch some mismatches
+between page object class fields and item class fields that would otherwise be
+hidden by ``skip_nonitem_fields``.
+
+
+.. _composition:
+
+Composition
```
BurnzZ, line 267: Nice! 💯 

Shall we also cover the case when the PO uses an item directly? For example: https://github.com/scrapinghub/scrapy-poet/blob/master/tests/test_web_poet_rules.py#L1054-L1065

docs/page-objects/items.rst
```diff
@@ -24,11 +25,61 @@ class. For example:
...
+
+To change the item class of a subclass that has already defined its item class,
+use :class:`~.Returns`:
+
+.. code-block:: python
+
+    @attrs.define
+    class MyOtherPage(MyPage, Returns[MyOtherItem]):
+        ...
+        ...
```
BurnzZ, line 71: Extra line

BurnzZ (commented):
Looking much more organized! ✨

wRAR (approved):

docs/page-objects/fields.rst
Gallaecio: What do you think about [this](https://github.com/scrapinghub/web-poet/pull/193/commits/41d6c743a070f804968750c9efdc26adb8afc032)?

BurnzZ (approved):

docs/page-objects/fields.rst
```diff
@@ -4,55 +4,16 @@
 Fields
 ======
 
-Background
-----------
+A field is a read-only property in a :ref:`page object class <page-objects>`
+that is decorated with :meth:`@field <web_poet.fields.field>`, is named after a
+key of the :ref:`item <items>` that the page object class returns, and returns
+the value for that item key based on :ref:`inputs <inputs>`.
```
kmike, line 10: I found this paragraph a bit hard to read. I guess it starts with "..a read-only property in a page-objects
that is decorated with..", where it's not immediately clear if a property is decorated, or a page object is decorated, or if you should use Python's `@property` (which must be read-only), and decorate it with `@field`.

Would it be possible to re-phrase it? Using bullet points instead of many commas may also help.

docs/page-objects/fields.rst
```diff
@@ -148,389 +91,442 @@ function when accessing the fields:
...
 
-    def add_brand(s, page):
-        return f"{page.brand} - {s}"
+    @attrs.define
+    class CustomPage(BasePage):
 
-    class MyPage(ItemPage):
-        response: HttpResponse
+        @field
+        def foo(self) -> str:
```
kmike, line 129: this field should be async def

kmike:
Looks good overall!

docs/page-objects/fields.rst
Gallaecio: What about https://github.com/scrapinghub/web-poet/pull/193/commits/1b812682a06fcb899f373f80e414883cebd10d10?

kmike:
Thanks @Gallaecio!

PR #195, opened 2024-01-16
add new AnyResponse

BurnzZ:
A way to address https://github.com/zytedata/zyte-spider-templates/issues/25.

Related PRs:
- https://github.com/scrapinghub/scrapy-poet/pull/184
- https://github.com/scrapy-plugins/scrapy-zyte-api/pull/161
- https://github.com/zytedata/zyte-spider-templates/pull/28

~Not quite sure if we should have a detailed documentation about this class in the tutorials since it may be deprecated in the next few months when Zyte API handles Auto-Configuration (_i.e. selecting the best extraction source between HttpResponse and BrowserResponse_).~ **EDIT:** Added docs.

### DIFF
--- a/web_poet/page_inputs/response.py
+++ b/web_poet/page_inputs/response.py
@@ -1,4 +1,4 @@
-from typing import Union
+from typing import Optional, Union
 
 import attrs
 
@@ -25,6 +25,13 @@ class AnyResponse(SelectableMixin, UrlShortcutsMixin):
         if isinstance(self.response, BrowserResponse):
             return self.response.html
         return self.response.text
+
+    @property
+    def status(self) -> Optional[int]:
+        """The int status code of the HTTP response, if available."""
+        if getattr(self.response, "status", None):
+            return self.response.status
+        return None
 
     def _selector_input(self) -> str:
         return self.text
--- a/web_poet/page_inputs/response.py
+++ b/web_poet/page_inputs/response.py
@@ -2,13 +2,14 @@ from typing import Union
 
 import attrs
 
+from web_poet.mixins import SelectableMixin, UrlShortcutsMixin
 from web_poet.page_inputs.browser import BrowserResponse
 from web_poet.page_inputs.http import HttpResponse
 from web_poet.page_inputs.url import ResponseUrl
 
 
 @attrs.define
-class AnyResponse:
+class AnyResponse(SelectableMixin, UrlShortcutsMixin):
     """A container that holds either :class:`~.BrowserResponse` or :class:`~.HttpResponse.`"""
 
     response: Union[BrowserResponse, HttpResponse]
@@ -24,3 +25,6 @@ class AnyResponse:
         if isinstance(self.response, BrowserResponse):
             return self.response.html
         return self.response.text
+
+    def _selector_input(self) -> str:
+        return self.text
--- a/web_poet/page_inputs/response.py
+++ b/web_poet/page_inputs/response.py
@@ -1,0 +1,26 @@
+from typing import Union
+
+import attrs
+
+from web_poet.page_inputs.browser import BrowserResponse
+from web_poet.page_inputs.http import HttpResponse
+from web_poet.page_inputs.url import ResponseUrl
+
+
+@attrs.define
+class HttpOrBrowserResponse:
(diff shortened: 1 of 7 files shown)


web_poet/page_inputs/response.py
```diff
@@ -0,0 +1,26 @@
...
+
+import attrs
+
+from web_poet.page_inputs.browser import BrowserResponse
+from web_poet.page_inputs.http import HttpResponse
+from web_poet.page_inputs.url import ResponseUrl
+
+
+@attrs.define
+class HttpOrBrowserResponse:
```
kmike, line 11: Could you please add selectors support to this dependency?

kmike:
> Not quite sure if we should have a detailed documentation about this class in the tutorials since it may be deprecated in the next few months when Zyte API handles Auto-Configuration (i.e. selecting the best extraction source between HttpResponse and BrowserResponse).

I think we need to have docs for this class; I'm not sure it makes sense to deprecate it later. It shouldn't be only about the extraction source, it should be possible to use this dependency without any extraction as well. The logic is the following: if there is BrowserResponse dependency in the plan, BrowserResponse is going to be used; if not, it probably should use HttpResponse.

web_poet/page_inputs/response.py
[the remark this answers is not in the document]
wRAR: I also thought it's better to call it Any<something>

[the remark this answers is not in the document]
BurnzZ: Agreed, that sounds much better 👍 https://github.com/scrapinghub/web-poet/pull/195/commits/9d6c61ceb52b5d580723e883d1cc116ea1d1a498

BurnzZ: Good one! https://github.com/scrapinghub/web-poet/pull/195/commits/1021a11be8247007279ea08065d32f49844d45df

web_poet/page_inputs/response.py
```diff
@@ -0,0 +1,30 @@
...
+import attrs
+
+from web_poet.mixins import SelectableMixin, UrlShortcutsMixin
+from web_poet.page_inputs.browser import BrowserResponse
+from web_poet.page_inputs.http import HttpResponse
+from web_poet.page_inputs.url import ResponseUrl
+
+
+@attrs.define
+class AnyResponse(SelectableMixin, UrlShortcutsMixin):
```
kmike, line 12: Would it make sense to add status as well?

web_poet/page_inputs/response.py
```diff
@@ -0,0 +1,30 @@
...
+    """A container that holds either :class:`~.BrowserResponse` or :class:`~.HttpResponse.`"""
+
+    response: Union[BrowserResponse, HttpResponse]
+
+    @property
+    def url(self) -> ResponseUrl:
+        """URL of the response."""
+        return self.response.url
```
PyExplorer, line 20: Should we convert self.response.url to str here or in heuristics? Because this is failed https://github.com/zytedata/zyte-spider-templates/pull/28/files#diff-ca9971971b7e9f63c22373022f4e32c974e800a6bc36f9e7e490453ed1038951R53

PyExplorer (approved):

web_poet/page_inputs/response.py
BurnzZ: Good catch! This should be `ResponseUrl`, so it's the same type when accessing `BrowserResponse.url` and `HttpResponse.url`. I'll fix the spider template code for this instead.

web_poet/page_inputs/response.py
BurnzZ: Missed that one. Thanks! https://github.com/scrapinghub/web-poet/pull/195/commits/895b4183018fe7b69292ae0f6c8c504ec658b7b0

web_poet/page_inputs/response.py
```diff
@@ -0,0 +1,37 @@
...
+        if isinstance(self.response, BrowserResponse):
+            return self.response.html
+        return self.response.text
+
+    @property
+    def status(self) -> Optional[int]:
+        """The int status code of the HTTP response, if available."""
+        if getattr(self.response, "status", None):
+            return self.response.status
+        return None
```
kmike, line 34: Why is returning self.response.status not enough?

kmike:
Looks good, thanks @BurnzZ!

PR #197, opened 2024-01-25
Refactor the additional request docs

Gallaecio:
Most of the existing content felt too reference-like to me, and we already have reference documentation, but I might have gone too far with the removing of  content, so please feel more than free to suggest bringing back stuff or featuring additional stuff.

I also moved the input validation page as a section of inputs, because it felt right there and it made sense to me that the content about input validation, mentioned in the additional requests page in the context of retries, was before the additional requests content in reading order.


docs/frameworks/additional-requests.rst
```diff
@@ -140,10 +140,10 @@ syntax.
...
 ------------------
 
-In the previous :ref:`exception-handling` section, we can see how Page Object
-developers could use the exception classes built inside **web-poet** to handle
-various ways additional requests MAY fail. In this section, we'll see the
-rationale and ways the framework MUST be able to do that.
+Page Object developers could use the exception classes built inside
+**web-poet** to handle various ways additional requests MAY fail. In this
+section, we'll see the rationale and ways the framework MUST be able to do
+that.
```
Gallaecio, line 146: The change here is the removal of the first part, ```In the previous :ref:`exception-handling` section, we can see how```, due to the removal of the referred content.

docs/page-objects/inputs.rst
```diff
@@ -84,3 +84,121 @@ You may define your own input classes if you are using a :ref:`framework
 
 However, note that custom input classes may make your :ref:`page object classes
 <page-object-classes>` less portable across frameworks.
+
+
+.. _input-validation:
+
+Input validation
```
kmike, line 91: Having it on a sepatate page makes sense to me, because the content look quite different from a general "inputs" page. 

I was thinking about the general "inputs" topic as something about explaining the concept of page object inputs - it's an uncommon concept.

On the other hand, input validation is a specific feature of page objects, with a prominent use case, which I think we should call out earlier: I think the majority of uses of validate_input would be dealing with unstable websites, which randomly return blank or invalid pages.

Gallaecio: OK, I’ll revert that. But would it be OK to move the topic above additional requests in the table of contents? (i.e. which topic should come first, input validation or additional requests?).

kmike: I don't have a strong opinion on that (haven't thought about that), but a first reaction is that additional request is a more widely used feature, so it should come before.

docs/page-objects/additional-requests.rst
```diff
@@ -4,799 +4,112 @@
...
 
+        @field
+        async def images(self):
+            api_url = f"https://api.example.com/v2/images?id={self.productId}"
+            try:
+                response = await self.http.get(api_url)
+            except HttpError:
+                return []
+            else:
+                return response.css(".product-images img::attr(src)").getall()
```
kmike, line 43: It seems the format doesn't fit Product.images schema

docs/page-objects/additional-requests.rst
```diff
@@ -4,799 +4,113 @@
...
 .. note::
 
-    The :meth:`~.HttpClient.batch_execute` method is a simple wrapper over
-    :py:func:`asyncio.gather`. Developers are free to use other functionalities
-    available inside :mod:`asyncio` to handle multiple requests.
-
-    For example, :py:func:`asyncio.as_completed` can be used to process the
-    first response from a group of requests as early as possible. However, the
-    order could be shuffled.
+    :meth:`HttpClient.execute <~.HttpClient.execute>` is expected to follow any
```
kmike, line 54: This applies to all HttpClient methods

docs/page-objects/additional-requests.rst
```diff
@@ -812,26 +126,30 @@ times before giving up:
...
             try:
-                data = await self.get_data()
-            except ValueError:
-                return {}
-            return {"data": data}
+                response = await self.get_images()
+            except HttpError:
+                return []
+            else:
+                return response.css(".product-images img::attr(src)").getall()
```
kmike, line 152: It doesn't fit the Product schema.
It seems we should add a processor for Product.images.

docs/page-objects/additional-requests.rst
```diff
@@ -4,799 +4,113 @@
...
-
-Fortunately, there are already some quick shortcuts on how to perform single
-additional requests using the :meth:`~.HttpClient.request`, :meth:`~.HttpClient.get`,
-and :meth:`~.HttpClient.post` methods of :class:`~.HttpClient`. These already
-define the :class:`~.HttpRequest` and executes it as well.
-
-.. _httpclient-get-example:
-
-A simple ``GET`` request
-------------------------
```
kmike, line 341: I think that's a good call to remove most of the reference-like content from these docs. However, I think it could make sense to bring back some of the example. The don't need to be complete working examples with page objects (it'd be too much code), but it seems it'd be great to show

* http.get, http.post and other methods
* passing custom http headers
* some of the response features, like response.css, response.text and response.json()
* response.status and it's relationship with error handling

Something similar to https://requests.readthedocs.io/en/latest/user/quickstart/ probably, but without some sections?

kmike: I think the goal here is for the users to be able to do 90% of their tasks without reading the reference docs.

Gallaecio: https://github.com/scrapinghub/web-poet/pull/197/commits/4f27bb879d1e9528e3fd6daf6d1ab8b6fedfd322

docs/page-objects/additional-requests.rst
```diff
@@ -4,799 +4,156 @@
...
-quick shortcuts for :meth:`~.HttpClient.request`:
+:class:`~.HttpClient` provides multiple asynchronous request methods, such as:
 
 .. code-block:: python
 
-    client = HttpClient()
+    http.get(url)
+    http.post(url)
+    http.request(url, method="...")
+    http.execute(HttpRequest(url, method="..."))
```
kmike, line 63:
```suggestion
    response = await http.get(url)
    response = await http.post(url, body=b"...")
    response = await http.request(url, method="...")
    response = await http.execute(HttpRequest(url, method="..."))
```

kmike (approved):

PR #201, opened 2024-02-26
Support annotated deps in serialization and testing.

wRAR:


web_poet/serialization/api.py
```diff
@@ -135,7 +136,19 @@ def serialize(deps: Iterable[Any]) -> SerializedData:
         cls = dep.__class__
         if is_injectable(cls):
             raise ValueError(f"Injectable type {cls} passed to serialize()")
-        result[_get_name_for_class(cls)] = serialize_leaf(dep)
+        if cls is AnnotatedResult:
+            key = f"AnnotatedResult {_get_name_for_class(dep.result.__class__)}"
```
wRAR, line 140: I needed a new separator and used a space. Suggestions for a better one a welcome.

web_poet/page_inputs/annotated.py
wRAR, line 1: This is not a dep but a helper class so maybe it shouldn't be in `page_inputs`, not sure.

setup.py
```diff
@@ -26,7 +26,7 @@
         "w3lib >= 1.22.0",
         "async-lru >= 1.0.3",
         "itemadapter >= 0.8.0",
-        "andi",
+        "andi >= 0.5.0",
```
Gallaecio, line 29: https://github.com/scrapinghub/web-poet/blob/6db4370b9771b0886c2f5e0b49bc4404812e54ee/tox.ini#L62-L66

Gallaecio (approved):

web_poet/page_inputs/annotated.py
```diff
@@ -0,0 +1,26 @@
+from dataclasses import dataclass
+from typing import Any, Tuple
+
+
+@dataclass
+class AnnotatedResult:
```
kmike, line 6: AnnotatedInstance? I think ...Result made sense in scrapy-poet, as it was a result of a provider, but here it doesn't sound right.

web_poet/serialization/api.py
kmike: What does it affect? Names of files on disk?

web_poet/serialization/api.py
```diff
@@ -135,7 +136,19 @@ def serialize(deps: Iterable[Any]) -> SerializedData:
...
+        else:
+            key = _get_name_for_class(cls)
+
+        if key in result:
```
kmike, line 144: It seems this condition won't be True if there is Annotated and non-Annotated version of the same dependency. Is it an issue?

web_poet/serialization/functions.py
```diff
@@ -193,3 +196,32 @@ def _deserialize_Stats(cls: Type[Stats], data: SerializedLeafData) -> Stats:
 
 
 register_serialization(_serialize_Stats, _deserialize_Stats)
+
+
+def _serialize_AnnotatedResult(o: AnnotatedResult) -> SerializedLeafData:
+    serialized_data: SerializedLeafData = {
+        "metadata.json": json.dumps(o.metadata).encode(),
```
kmike, line 203: Let's use some common json options, like indent and ensure_ascii, to make this file more readable.

web_poet/serialization/api.py
wRAR: Yes.

web_poet/serialization/api.py
wRAR: Annotated and non-Annotated versions of the same dependency can be serialized together (the key is different for the Annotated classes) so that should be fine.

web_poet/page_inputs/annotated.py
wRAR: We decided to move it to `web_poet.annotated` and remove it from the top-level `__init__.py`.

kmike (approved):

PR #205, opened 2024-06-13
Add base_url to Selector and implement HttpRequest.from_form

Gallaecio:
`HttpRequest.from_form` is based on [Scrapy’s FormRequest.from_response()](https://github.com/scrapy/scrapy/blob/1282ddf8f77299edf613679c2ee0b606e96808ce/scrapy/http/request/form.py#L69), but with a minimal API and without support for the `click` thing which I don’t fully understand.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[report]
+exclude_lines =
+    if TYPE_CHECKING:
--- a/tests/test_page_inputs.py
+++ b/tests/test_page_inputs.py
@@ -291,6 +291,26 @@ def test_http_request_from_form_select_no_selected() -> None:
     assert request.body == b"query=&bar=code&baz=ooka"
 
 
+def test_http_request_from_form_select_no_options() -> None:
+    url = "https://example.com"
+    response = HttpResponse(
+        url,
+        b"""
+        <!doctype html>
+        <title>a</title>
+        <form id="search-form" accept-charset="utf-8" action="/search" method="POST">
+            <input type="text" value="" name="query">
+            <select name="bar"></select>
+            <input type="submit">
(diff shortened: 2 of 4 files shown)


wRAR (approved):

kmike:
Hey! I think that's definitely a feature we should have in web-poet 👍 

How hard would it be to move the form submission code to a separate library, and then use it both in Scrapy and in web-poet, to avoid double-maintenance? Is it the "click" support which makes it hard?

Gallaecio:
I did not even implement the click support here, but I don’t imagine it would make it hard.

The main reason why I did not move this to w3lib is that the function builds a request. It we more it to w3lib, what should we return? A named tuple? What should be the format of headers, a dict of values, a dict of lists, or a tuple of tuples?

kmike:
> The main reason why I did not move this to w3lib

I was thinking about a separate library like itemloaders or itemadapter, though w3lib also works.

> If we move it to w3lib, what should we return? A named tuple? What should be the format of headers, a dict of values, a dict of lists, or a tuple of tuples?

That's a good point. It should probably return a dataclass in this case. The headers should be a either a list of tuples or a dict of lists, probably doesn't matter much.

kmike:
About click - as I understand, it's needed when there are multiple submit (or other?) buttons on a form, and the right one is not the first one (which Scrapy picks itself). For example, if you use Formasaurus to submit a search form, you may want to use the detected "submit button", not "reset/clear button" (see https://formasaurus.readthedocs.io/en/latest/usage.html#field-types), though it needs experiments.

wRAR:
It's too short for a separate new library IMO.

kmike:
My preference (+0.5) is a new library in scrapy org.

PR #206, opened 2024-07-12
Use form2request for the HttpRequest.from_form implementation

Gallaecio:


setup.py
```diff
@@ -20,7 +20,8 @@
     entry_points={"pytest11": ["web-poet = web_poet.testing.pytest"]},
     install_requires=[
         "attrs >= 21.3.0",
-        "parsel >= 1.5.0",
+        "form2request >= 0.1.0",
```
kmike, line 23: What do you think about requiring 0.1.1 once it's released?

Gallaecio:
@kmike What do you think about implementing a `from_dataclass` method (open to name suggestions) that supports dataclasses like `form2request.Request`?

In Formasaurus I’m thinking of using form2request directly, to not depend on web-poet, and return a `form2request.Request` as well, which users of both Formasaurus and web-poet will want to convert to `HttpRequest`.

It’s not a hard conversion to do, so users could do it manually. I just wonder whether or not it is a good idea. My main concerns are that “dataclass” is quite generic, if someone makes up some dataclass-based request data container we may not support it, but I also do not think a `from_form2request_request` class method is a good idea.

kmike:
I don't like "from_dataclass" for the same reason - it assumes a very specific dataclass, the one returned by form2request.

What do you think about adding to_... methods to form2request?

* to_scrapy
* to_poet
* to_requests

It seems it'd solve the issue, and also it'd simplify form2request docs and usage.

Gallaecio:
Shall we remove `HttpRequest.from_form` then? (has not been released yet)

kmike:
> Shall we remove HttpRequest.from_form then? (has not been released yet)

Hm, a good question. If we go with to.. methods, It looks redundant indeed, as HttpRequest.from_form(form) is the same as form2request(form).to_poet().

This does make this feature less discoverable for web-poet users though. So, no strong opinion on that.

Gallaecio:
Maybe we can address discoverability through docs.

kmike (approved):

PR #212, opened 2025-01-22
Do not consider type(None) injectable

Gallaecio:
I am hoping the comment is out of date or wrong, but I am not 100% sure of what the line was meant to fix. @kmike Do you recall why this line was there? (it comes from the initial commit).

This change does break 2 scrapy-poet tests, but I believe those tests expectations are what’s wrong (https://github.com/scrapinghub/scrapy-poet/pull/218).

cc @kalessin

### DIFF
--- a/tests/test_pages.py
+++ b/tests/test_pages.py
@@ -69,6 +69,7 @@ def test_is_injectable() -> None:
             }
 
     assert is_injectable(None) is False
+    assert is_injectable(type(None)) is False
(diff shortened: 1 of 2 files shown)


kmike:
Interesting; I don't recall what was the reason behind the comment. Added a suggestion for an extra test in scrapy-poet PR.

kmike (approved):

PR #215, opened 2025-02-07
Migrate to ruff

wRAR:
Enabling postponed annotations in tests/test_serialization.py breaks two tests because `get_type_hints()` probably doesn't resolve locally defined classes correctly in this case. In one test it's called from `singledispatch()`, and it seems to be a known incompatibility before Python 3.14. In another test it's called from andi, and it may suggest that andi doesn't support some user code that uses postponed annotations in the plan callables (andi has some tests for them, but maybe they don't test everything).

Edit: enabling postponed annotations in files that define attr-classes also leads to similar breakages. It could be possible to fix somewhere, though the actual crash is inside `typing.get_type_hints()`, but needs to be done separately and won't help older versions.


.pre-commit-config.yaml
```diff
@@ -23,3 +23,9 @@ repos:
           - flake8-string-format
     repo: https://github.com/pycqa/flake8
     rev: 7.1.1
+  - hooks:
+      - id: pyupgrade
+        args: [--py39-plus]
+        exclude: ^docs/tutorial-project/
+    repo: https://github.com/asottile/pyupgrade
+    rev: v3.19.1
```
Gallaecio, line 31: Would it make sense to switch to Ruff instead?

web_poet/utils.py
```diff
@@ -137,7 +139,7 @@ def __call__(cls, *args, **kwargs):
         # deprecated class is in jinja2 template). __module__ attribute is not
         # important enough to raise an exception as users may be unable
         # to fix inspect.stack() errors.
-        warn(f"Error detecting parent module: {e!r}")
+        warn(f"Error detecting parent module: {e!r}", stacklevel=1)
```
wRAR, line 142: Not sure what's the good level here, so I kept the implicit one.

web_poet/page_inputs/browser.py
```diff
@@ -7,7 +7,7 @@
 from .url import ResponseUrl
 
 
-class BrowserHtml(SelectableMixin, str):
+class BrowserHtml(SelectableMixin, str):  # noqa: SLOT000
```
wRAR, line 10: Not 100% sure how it works, but it looks like even if `__slots__` is defined here, it has no effect (I assume because the other base class doesn't have it).

setup.py
```diff
@@ -1,14 +1,12 @@
...
+from pathlib import Path
 
+from setuptools import find_packages, setup
 
 setup(
     name="web-poet",
     version="0.18.0",
     description="Zyte's Page Object pattern for web scraping",
-    long_description=long_description,
+    long_description=Path("README.rst").read_text(),
```
Gallaecio, line 9: Not an issue specific to these changes, but this may be a good opportunity to specify the encoding, for cross-platform support (not sure when that would become an issue, but it has hit us in the past precisely in setup.py).

Gallaecio (approved):

Issue #188, opened 2023-09-06
IDE-friendly unit tests


proway2:
Right now tests created for page objects aren't [discoverable](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#conventions-for-python-test-discovery) by `pytest`, therefore IDEs (e.g. VS Code) aren't aware of tests presence, though there might be other unit tests discoverable by IDE. This results in the next problems:
- Even if you need to run tests just for a single netloc you've got to run either all tests or to specify a path to a given test, which is not very convinient if you work with few different netlocs at a time.
- web-poet implicitly registers a `pytest` plugin which, sometimes, is not needed. Disabling is possible, but you've got to use something like `-p no:web-poet`. Which is again makes the whole process less natural for the IDE.
- No standard way to run a single test for a single URL.
- Overall, the whole approach to unit tests varies drastically from the usual way `pytest` tests are created and used.

Would be really nice to have IDE-friendly tests for page objects that are:
- discoverable by `pytest` and IDE without the need to configure some low-level stuff.
- possible to run and debug in IDE a single test for a single URL.
- `pytest` plugin could be easily disabled if needed.

kmike:
Hey! The problem clearly exists (you're having issues with VSCode and web-poet fixtures). I wanted to clarify some points though.

> Right now tests created for page objects aren't [discoverable](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#conventions-for-python-test-discovery) by pytest

The tests are discoverable, as can be seen when running pytest using CLI. We're using official pytest API to write plugins: https://docs.pytest.org/en/7.1.x/how-to/writing_plugins.html, or see e.g. an example: https://docs.pytest.org/en/7.1.x/example/nonpython.html#yaml-plugin. So, the issue is not that tests are not discoverable, but that the VSCode doesn't seem to support all pytest features.

> Even if you need to run tests just for a single netloc you've got to run either all tests or to specify a path to a given test, which is not very convinient if you work with few different netlocs at a time.

To run tests for a single netloc locally, one can pass path to the fixture to pytest. I think something like `pytest my_proj/fixtures/mywebsite` should work. I'm not using IDEs personally to run the tests (running them from the command line), but maybe it's possible to configure the IDE to support that - though that may not be possible, or it may depend on IDE.

[...]

PR #224, opened 2025-06-05
Allow running individual tests

### LINKED ISSUE #144: A way to run just one test case
I see three useful ways to run tests: by passing the `fixtures` directory / not passing anything, by passing the `fixtures/<page object>` directory and by passing the `fixtures/<page object>/<test case>` directory. Out of these, the last use case doesn't work, instead running all test cases in `fixtures/<page object>`, because `collect_file_hook()` always returns a `WebPoetFile` corresponding to `fixtures/<page object>`. On the other hand, if `WebPoetFile` is like a Python file with multiple test functions and one test case is like a single test function, then the correct usage would be `fixtures/<page object>::<test case>`? It currenty returns "directory argument cannot contain ::".

Gallaecio:
Resolves #144.

This is a backward-incompatible PR that I am completely OK with closing without merging.

# Implementation

It makes it so that the node IDs of tests use non-filesystem sections for all components (page object import path, test case ID, field test or special item test) instead of using a path for those first 2.

To achieve that, it requires a `test.py` file (its contents are ignored, the filename was an arbitrary choice) at the root of `fixtures/`. While ideally `scrapy savefixture` should create that file, the proposed implementation hackily creates it when you run `pytest` on its own or `pytest fixtures/`, for usability purposes.

The result is that an old node ID like:

> `fixtures/books_scraper.pages.news_ycombinator_com.HackerNewsHomePage/test-1::next_page_url`

Now becomes:

> `fixtures/test.py::books_scraper.pages.news_ycombinator_com.HackerNewsHomePage::test-1::next_page_url`

# The good

Unlike before, that complete node ID for a single test can be run:

```
$ pytest fixtures/test.py::books_scraper.pages.news_ycombinator_com.HackerNewsHomePage::test-1::next_page_url
===================================================== test session starts =====================================================
platform linux -- Python 3.13.3, pytest-8.4.0, pluggy-1.6.0
rootdir: /home/adrian/temporal/project
configfile: pytest.ini
plugins: time-machine-2.16.0, web-poet-0.18.0
collected 1 item                                                                                                              

fixtures/test.py .                                                                                                      [100%]

====================================================== 1 passed in 0.05s ======================================================
```

# The bad

This is of course backward-incompatible when it comes to running tests. `pytest` and `pytest fixtures/` work as before, but everything else has a different node ID now, so you need to change the command to run subsets of tests.

Also, a significant tradeoff of supporting running a single test is that you can no longer take advantage of terminal autocompletion to build the command to run all tests for a given page object class or for one of its test cases.

# The ugly

My main goal that lead me to this PR was not to address #144, but to address #188. But I was not able to achieve that in a reasonable time, and I am not sure if it is worth pursuing at this point.

The main issue seems to be that the [build_test_tree](https://github.com/microsoft/vscode-python/blob/02476f0ac9873be3aed81d39b62d367c6947e389/python_files/vscode_pytest/__init__.py#L526) function of vscode-python makes some assumptions regarding pytest node IDs, with `::`-based splits and similar. That may be what’s causing other issues reported upstream like https://github.com/microsoft/vscode-python/issues/24455, https://github.com/microsoft/vscode-python/issues/23868, but not sure.

Making our test collection work might require something bold like (in no particular order):
- Rewrite our fixture test to be more like regular pytest tests. I think having a file inside each testcase folder might be get us a long wait, possibly address things and not require the root `test.py` (maybe I’ll test this)
- Make upstream changes to that function to work with custom test collections like ours. Not sure how feasible it is to do, though; you have to assume that the complexity of the current implementation is that way for a reason.
- Write our own pytest integration, at least for fixtures. Not ideal in the least, in my opinion.

### DIFF
--- a/web_poet/testing/pytest.py
+++ b/web_poet/testing/pytest.py
@@ -264,11 +264,10 @@ def collect_file_hook(
         if not fixture.is_valid():
             return None
         fixtures_test_file = page_dir.parent / "test.py"
-        if (
-            fixtures_test_file in _found_fixtures_test_files
-            or not fixtures_test_file.exists()
-        ):
+        if fixtures_test_file in _found_fixtures_test_files:
             return None
+        if not fixtures_test_file.exists():
+            fixtures_test_file.write_text("")
         _found_fixtures_test_files.add(fixtures_test_file)
         return _get_file(parent, path=fixtures_test_file)
     if file_path.name == "test.py" and file_path not in _found_fixtures_test_files:
--- a/web_poet/testing/pytest.py
+++ b/web_poet/testing/pytest.py
@@ -32,7 +32,7 @@ class _PathCompatMixin:
         return self.path if _new_pytest else Path(self.fspath)
 
 
-class WebPoetFile(pytest.File, _PathCompatMixin):
+class FixturesTestFile(pytest.File, _PathCompatMixin):
     """Represents a directory containing test subdirectories for one Page Object."""
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


Gallaecio:
> * possibly address things and not require the root `test.py` (maybe I’ll test this)

Making `output.json` or `exception.json` a “test file” did the trick, I’ll open a separate PR soon.

PR #232, opened 2025-08-21
Fix getting type hints for WebPage, add a test.

wRAR:


wRAR:
OK, this needs a less straightforward fix.

Looks like this is broken for some time, but nobody ran `get_type_hints()` (or `ItemAdapter.get_json_schema()`) on `WebPage` or its subclasses.

wRAR:
(I don't even know if calling those on those classes is useful, it was found by mistake)

Gallaecio (approved):

PR #244, opened 2026-01-14
Support annotations with frozenset, bytes or nested tuples

Gallaecio:


Gallaecio:
Alternative, better(?) approach: https://github.com/scrapinghub/web-poet/pull/245

PR #258, opened 2026-04-06
Decouple fixtures from POs

wRAR:
Partial fix for #255

Fixture instances are no longer know tied to a single PO type, and all methods that need a PO now take the class object. The pytest discovery behavior is currently unchanged and still requires the parent folder name to be named after the PO.


tests/test_testing_pytest.py
wRAR, line 1: No need to read this.

web_poet/testing/pytest.py
```diff
@@ -25,50 +26,73 @@ class TestCase(pytest.File):
     """Represents the ``output.json`` or ``exception.json`` file in a testcase
     directory."""
```
kmike: It's a bit more than this now, because it also includes page object

kmike:
The way code is organized makes sense to me. I haven't checked if it works or not.

tests/test_testing.py
```diff
@@ -133,607 +115,86 @@ def test_fixture_adapter(book_list_html_response, tmp_path) -> None:
     assert saved_output["foo"] == "Bar"
 
     loaded_fixture = Fixture(base_dir / "test-1")
-    loaded_output = loaded_fixture.get_output()
+    loaded_output = loaded_fixture.get_output(MyItemPage)
```
kmike: loaded_output feels wrong; po_output?

tests/test_testing.py
```diff
@@ -133,607 +115,86 @@ def test_fixture_adapter(book_list_html_response, tmp_path) -> None:
...
+) -> None:
+    type_name = get_fq_class_name(MyItemPage3)
+    base_dir = tmp_path / "fixtures" / type_name
+    fixture = Fixture.save(
+        base_dir,
+        inputs=[book_list_html_response],
+        item={"foo": "bar2", "egg": "spam", "hello": "world"},
     )
-    cli_main(["rerun", str(fixture.path), "--fields=foo,egg"])
+    cli_main(["rerun", str(fixture.path), "-p", type_name, "--fields=foo,egg"])
```
kmike: is the API for rerun CLI command different now?

web_poet/testing/__main__.py
```diff
@@ -37,6 +43,9 @@ def main(argv=None):
         "",
     )
     parser_rerun.add_argument("fixture_path", type=str, help="Path to a fixture")
+    parser_rerun.add_argument(
+        "--page-object", "-p", type=str, help="Page object type name", required=True
```
kmike: Can we keep it backwards compat? In any way, we should update docs about it, rerun command is mentioned in docs.

kmike (approved):

PR #259, opened 2026-04-06
Built-in framework

### LINKED ISSUE #256: Make a public built-in framework to run page objects
We have https://github.com/scrapinghub/web-poet/blob/master/web_poet/example.py. It'd be nice to make it more suitable for runinng POs quickly. Cover more dependencies, maybe make IO async, document as not tutorial-only. Allow to create test fixtures using this framework as well.

AdrianAtZyte:
Resolves #256.

Notes:
- All naming choices are up for discussion, including the `Poet` class and the `simple_framework` module.
- I have considered implementing a **session()** context manager à la python-zyte-api, but thought it best to keep it out of scope for now.
- Same for a **sync** API. I wonder if we even need one.
- I went with **niquests** instead of aiohttp mostly to try it out. I have no problem switching to aiohttp or something else. I keep typing `nirequests`, and I am starting to believe that may be what contributes the most to adoption not being higher.
- I have considered implementing **Retry** support, but eventually changed my mind. I was going to implement a simple approach, where you only select max retry times, but I kept thinking about delays between retries, exponential backoff, tenacity, and thought it best to leave it up to users to handle retries instead.
- I did not implement test fixture saving as I think it may make more sense to post-pone it until other related changes are implemented (e.g. removing the requirement for a page object).


.vscode/settings.json
wRAR, line 1: Is this intentional?

tests/test_extras.py
```diff
@@ -0,0 +1,6 @@
+import pytest
+
+
+def test_simple_framework():
```
wRAR: How does this pass when the deps are available? I don't see it being skipped.

.vscode/settings.json
AdrianAtZyte: Yes. But no strong opinion on keeping it.

tests/test_extras.py
AdrianAtZyte: The `min-framework` and `framework` tox envs target `tests/test_framework.py` specifically, i.e. they do not run this test.

docs/simple-framework.rst
```diff
@@ -0,0 +1,124 @@
...
+===========
+
+The main limitation of the simple framework is that it is not a complete
+scraping framework like :doc:`Scrapy <scrapy:index>`, which can support
+web-poet thanks to :doc:`scrapy-poet <scrapy-poet:index>`.
+
+As a web-poet framework, the simple framework also lacks support for
+:ref:`custom input classes <custom-inputs>`, :exc:`~web_poet.exceptions.Retry`
+and :exc:`~web_poet.exceptions.UseFallback`.
+
```
kmike: This seems like a small list; so, as a web-poet framework the current one looks almost feature-complete. So, I wonder if we shouldn't downplay it as "simple framework", but called is "aio framework", "built-in framework", "framework" or something like this.

docs/framework.rst
```diff
@@ -0,0 +1,124 @@
...
+    consume_modules("myproject.pages")
+
+    poet = Poet()
+
+    # Get an item directly.
+    item = await poet.get_item("http://example.com/book/1", Book)
+
+    # Or, if you prefer, get a page object first.
+    page = await poet.get_page("http://example.com/book/1", BookPage)
+    item = await ensure_awaitable(page.to_item())
```
kmike: I think it'd be good to have first-class PO running, something like item = await poet.get_item(url, BookPage). I think it also can be a first example, not the last, as it shouldn't require consume_modules. So, you define a PO, and have an easy way to run it. All the extra stuff - handle_urls, item classes, etc. is optional, for more advanced use cases.

web_poet/framework/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+
+try:
+    import niquests  # noqa: F401
+    from playwright.async_api import async_playwright  # noqa: F401
+except ImportError:
+    message = (
+        "Could not import web_poet.simple_framework dependencies. "
+        "Install `web-poet[simple_framework]`."
+    )
+    raise ImportError(message) from None
```
kmike: does it still show the original traceback from the above imports?

web_poet/framework/_api.py
```diff
@@ -0,0 +1,194 @@
...
+    Example usage:
+
+    .. code-block:: python
+
+        Annotated[BrowserResponse, browser("firefox")]
+    """
+    return annotation_encode(f"{ANNOTATION_PREFIX}{name}")
+
+
+RequestLike: TypeAlias = HttpRequest | RequestUrl | str
```
kmike: should we support ResponseUrl as well?

web_poet/framework/_api.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+def _normalize_request(request: RequestLike) -> HttpRequest:
+    if not isinstance(request, HttpRequest):
+        request = HttpRequest(url=request)
+    return request
+
+
+class Poet:
```
kmike: this is a cute name :) But I wonder if it'd be more straightforward for users to understand if we call if Framework.get_page, or Runner.get_page, something like this. Poet in the library name is "page objects, etc", not 100% match for what this class does.

web_poet/framework/_providers.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+def _get_http_response_from_nirequests_response(
+    request: HttpRequest, response: niquests.Response
+) -> HttpResponse:
+    return HttpResponse(
+        response.url or request.url,
+        status=response.status_code,
+        body=response.content or b"",
+        headers=response.headers,
```
kmike, line 43: Do we test explicitly that headers conversion works for niquests? I'm thinking about something like this: https://github.com/scrapinghub/web-poet/blob/2bd22ca69a0845784c82a5892395172d77b9c991/tests/test_page_inputs.py#L262

web_poet/framework/_providers.py
```diff
@@ -0,0 +1,213 @@
...
+    )
+
+
+async def _get_http_response_from_http_request(request: HttpRequest) -> HttpResponse:
+    response = await niquests.aget(str(request.url), timeout=300)
```
kmike: Where are exceptions converted to https://github.com/scrapinghub/web-poet/blob/2bd22ca69a0845784c82a5892395172d77b9c991/web_poet/exceptions/http.py#L14? Is it needed?

web_poet/framework/_providers.py
wRAR: Btw niquests (and, presumably, requests) converts multiple values of the same header to one comma-separate value instead of using some multi-value structure.

web_poet/framework/__init__.py
AdrianAtZyte: It does not:

```
$ .tox/py/bin/python -c "import web_poet.framework"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
    import web_poet.framework
  File "/home/adrian/Documents/web-poet/web_poet/framework/__init__.py", line 11, in <module>
    raise ImportError(message) from None
ImportError: Could not import web_poet.framework dependencies. Install `web-poet[framework]`.
```

That’s on purpose, thought, since this should not happen provided the extra is installed.

Do you think we should let it show, in case some future version of the packages or some weird system issue is causing the problem?

web_poet/framework/_providers.py
AdrianAtZyte: ~~Is it OK to map all of them as `HttpRequestError`? [That’s what niquests does](https://niquests.readthedocs.io/en/stable/_modules/niquests/exceptions.html), although I guess we should be more picky.~~

> Frameworks implementing **web-poet** should **NOT** raise this  exception. — [HttpResponseError](https://github.com/scrapinghub/web-poet/blob/2bd22ca69a0845784c82a5892395172d77b9c991/web_poet/exceptions/http.py#L54)

Nevermind :slightly_smiling_face:

web_poet/framework/_providers.py
```diff
@@ -42,28 +43,53 @@ def _get_http_response_from_nirequests_response(
...
+            "Browser provider only supports plain GET requests with a URL (no "
+            "headers or body)",
+            request=request,
+        )
+    if request.headers and len(request.headers):
+        raise HttpRequestError(
+            "Browser provider does not support requests with headers; only "
+            "plain GET with a URL is supported",
+            request=request,
+        )
```
kmike, line 69: I wonder if it'd be better to ignore headers and log a warning, instead of raising. E.g. you have some code, and want to switch from HttpResponse to BrowserHtml; maybe you implemented some basic header setting (e.g. referer).

docs/intro/tutorial.rst
```diff
@@ -209,29 +208,28 @@ And the ``print(item)`` statement should output the following:
...
 
 -   the URL from which you want to extract data, and
 -   the desired output, either a :ref:`page object class <page-object-classes>`
     or, in this case, an :ref:`item class <items>`.
 
-Notice that you must also call :func:`~web_poet.rules.consume_modules` once
-before your first call to ``get_item``. ``consume_modules`` ensures that the
-specified Python modules are loaded. You pass ``consume_modules`` the import
+:Notice that you must also call :func:`~web_poet.rules.consume_modules` once
+before your first call to :meth:`~web_poet.framework.Framework.get_item`.
```
kmike: This only applies if you use items, not page objects with get_item

docs/framework.rst
```diff
@@ -0,0 +1,164 @@
...
+    class BookPage(WebPage[Book]):
+        @field
+        def title(self) -> str:
+            return self.response.css("h1::text").get()
+
+
+    framework = Framework()
+    item = await framework.get_item("https://books.example.com/book/1", BookPage)
+
+    # Or, if you prefer, get a page object first.
```
kmike:
```suggestion
    # Or, if you prefer, get a page object instance first.
```

kmike: Use case could be getting just some of the fields

web_poet/framework/__init__.py
kmike: Yes, I think it can be helpful to see traceback. E.g. maybe you think playwright is installed, but it's in a wrong virtualenv. Or there is some weird installation issue, right.

web_poet/framework/_api.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+class Framework:
+    """Manager of the :ref:`built-in framework <framework>`.
+
+    *registry* is the :class:`~web_poet.rules.RulesRegistry` from where page
+    objects resolve their dependencies. If ``None``,
+    :data:`~web_poet.default_registry` is used.
+
+    *default_browser* is the Playwright browser engine to use when browser
```
kmike: "browser" bothers me a bit; based on the name, I'd expect to pass some kind of browser instance here, for the framework to use. 

The same applies for annotations ("browser"). I wonder if it'd make sense to make it more explicit, e.g. default_playwright_engine, or playwright_engine_name, something like this? Or do you want to keep it opaque that we use Playwright under the hood?

If we want something really standard, it would belong to page_inputs, not to framework (I think it's a right call to have it in framework-only now).

kmike (approved):
Looks good overall!

kmike:
I don't know if we're merging or squash-merging these days :) Anyways, thanks @AdrianAtZyte!

PR #263, opened 2026-04-07
Fall back to a field method if to_item() fails

AdrianAtZyte:
Resolves #254.


web_poet/testing/fixture.py
```diff
@@ -135,21 +138,47 @@ def _get_output(self) -> dict:
...
+        frozen_time_parsed = self._parse_frozen_time(frozen_time)
+        with time_machine.travel(frozen_time_parsed):
+            return callback()
+
+    @memoizemethod_noargs
+    def _get_readable_fields(self) -> set[str]:
+        return set(get_fields_dict(self.get_page()))
+
+    def _can_read_field_separately(self, name: str) -> bool:
+        return name in self._get_readable_fields()
```
wRAR: What's the logic here, what fields would not be in there?

AdrianAtZyte: Non-fields, i.e. "fields" returned from `to_item()` for which a `@field`-decorated method does not actually exist.

wRAR (approved):

AdrianAtZyte:
@kmike Thoughts on the general approach?

kmike:
Hm, I'm not sure :) On one hand, it's a practical approach, with minimal changes, which works in 95% cases and solves an issue. 

On another hand, checking final item and checking individual fields - two very different things. For example, you may have fields, and custom to_item, which changes some of the values. One field raises an error, now we switch from item check to field checks, and many fields start to fail (because they no longer account for item processing). That's why I preferred an explicit toggle - do you want to run tests in "to_item" mode, or in "per-field" mode.

AdrianAtZyte:
Shall we make the new mode the default, even if it is a backward-incompatible change? Or do you prefer to keep `to_item` as the default field mode?

AdrianAtZyte:
https://github.com/scrapinghub/web-poet/pull/263/commits/61292f232f138ccfeaf4ce4f15224f2caee0bc99 keeps the old behavior as default, and adds a new CLI option to enable the new behavior.

For the record, I kept the old behavior as default more for a personal "that way I don’t have to change tens of tests" than for "keeping backward compatibility is nice for users".

wRAR:
This has conflicts.

tests/test_testing.py
```diff
@@ -135,6 +157,172 @@ def test_fixture_adapter(book_list_html_response, tmp_path) -> None:
     assert actual_output["foo"] == "Bar"
 
 
+def test_get_output_field_missing_raises(pytester, book_list_html_response) -> None:
```
wRAR: Can you please move `pytester` tests into the recently split out `test_testing_pytest.py`?

wRAR (approved):

PR #265, opened 2026-04-10
Support max_retries in Retry

AdrianAtZyte:
Post-merge:
- [ ] Read that parameter from scrapy-poet, [here](https://github.com/scrapinghub/scrapy-poet/blob/f2556ea6bc92a5cfa2f6da1152d1932ccbde031e/scrapy_poet/spidermiddlewares.py#L25-L29).

### DIFF
--- a/docs/page-objects/input-validation.rst
+++ b/docs/page-objects/input-validation.rst
@@ -62,10 +62,8 @@ For example:
    def validate_input(self):
        if self.css(".product-id::text") is not None:
            return
-       if self.css(".http-429-error"):
-           raise Retry()
        if self.css(".http-503-error"):
-           raise Retry(max_retries=2)
+           raise Retry()
        if self.css(".product"):
            raise UseFallback()
        if self.css(".product-list"):
--- a/docs/page-objects/input-validation.rst
+++ b/docs/page-objects/input-validation.rst
@@ -62,8 +62,10 @@ For example:
    def validate_input(self):
        if self.css(".product-id::text") is not None:
            return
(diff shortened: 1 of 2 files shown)


docs/page-objects/input-validation.rst
```diff
@@ -62,8 +62,10 @@ For example:
    def validate_input(self):
        if self.css(".product-id::text") is not None:
            return
-       if self.css(".http-503-error"):
+       if self.css(".http-429-error"):
            raise Retry()
+       if self.css(".http-503-error"):
+           raise Retry(max_retries=2)
```
kmike: Not sure if this is right - we kind-of rely on framework to have infinite retries by default. 

I think it may be ok to skip the example, and just add a feature.

kmike (approved):

kmike:
should it be supported in the framework?

AdrianAtZyte:
:shrug: 

> I have considered implementing **Retry** support, but eventually changed my mind. I was going to implement a simple approach, where you only select max retry times, but I kept thinking about delays between retries, exponential backoff, tenacity, and thought it best to leave it up to users to handle retries instead. — https://github.com/scrapinghub/web-poet/pull/259

If we do implement retries, we should make some decisions, at the very least what should be the default max retries, how to space retries out in time.

kmike:
ah, I misremembered it; I was thinking there is some hardcoded 3-times retry somewhere, but can't find it anymore :) human hallucination. We shouldn't add Retry support to framework as a part of this task.

wRAR (approved):

PR #271, opened 2026-05-14
Support date and datetime JSON encoding

### LINKED ISSUE #270: Improve documentation for register_serialization
This might belong in `scrapy-poet` but I think it rightfully belongs here because, at minimum, this is a web-poet documentation issue.

The docs say

> Each dependency is serialized to one or several bytes objects, each of which is saved as a single file. web_poet.serialization.serialize_leaf() and web_poet.serialization.deserialize_leaf() are used to convert between a dependency and this set of bytes objects. They are implemented using [functools.singledispatch()](https://docs.python.org/3/library/functools.html#functools.singledispatch) and while the types provided by web-poet are supported out of the box, user-defined types need a pair of implementation functions that need to be registered using web_poet.serialization.register_serialization().

I can't seem to find any docs about *when* this should happen, so I'd put it in your settings.py file as

```
def serialize(obj: datetime):
    return {"json": obj.isoformat().encode("utf-8")}

def deserialize_datetime(cls: type[datetime], data: dict[str, bytes]) -> datetime:
    return datetime.fromisoformat(data["json"].decode("utf-8"))

register_serialization(serialize, deserialize_datetime)
```

But it doesn't seem to make a difference. The registering of the (de)serializers *does* happen, as I've traced the code doing it.

```

  File "[snip]/web_poet/serialization/utils.py", line 41, in default
    return super().default(o)
           ^^^^^^^^^^^^^^^^^^
  File "[snip]/python3.12/json/encoder.py", line 180, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type datetime is not JSON serializable
```

I figure this must already have a solution, because surely I'm not the first person to have this issue. The docs seem to acknowledge a solution as well, but it doesn't provide an example.

AdrianAtZyte:
Fixes #270.

### DIFF
--- a/tests/test_testing.py
+++ b/tests/test_testing.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import datetime as dt
 import json
 from collections import deque
 from typing import TYPE_CHECKING, Any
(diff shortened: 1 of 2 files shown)


wRAR:
Is decoding out of scope?

AdrianAtZyte:
I believe so. The way things seem to work is that both values, actual and expected, get encoded and then compared. I don’t think decoding is needed, at least at the moment.

wRAR (approved):

PR #274, opened 2026-06-08
Fix field typing

### LINKED ISSUE #273: Object of type "_field" is not callable
The documentation itself demonstrates code like

```
@field(meta={"expensive": True})
def ...
```

However, because `field` returns a `_field` class that lacks a `__call__` method, this fails type checking with the error:

```
Object of type "_field" is not callable
  Attribute "__call__" is unknown
```

Only 

```
@field
def ...
```

seems to typecheck.

AdrianAtZyte:
Fixes #273.

### DIFF
--- a/tests_typing/test_fields.mypy-testing
+++ b/tests_typing/test_fields.mypy-testing
@@ -59,6 +59,31 @@ class Item:
 
 
 @pytest.mark.mypy_testing
+def test_field_with_args_return_type() -> None:
+    f = field(meta={"expensive": True})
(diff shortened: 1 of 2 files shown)


wRAR:
Does this supersede #268?

AdrianAtZyte:
Damn, I thought we had already merged that one. I think https://github.com/scrapinghub/web-poet/pull/268 may supersede this one instead. Let me make sure, and if so, close this PR and mark #268 as fixing #273.

AdrianAtZyte:
:facepalm:

PR #276, opened 2026-06-17
Add support for pytest 9.1.0 to comparison_error_message().

wRAR:
Private API :-/

### DIFF
--- a/web_poet/testing/utils.py
+++ b/web_poet/testing/utils.py
@@ -1,11 +1,9 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any
+from typing import Any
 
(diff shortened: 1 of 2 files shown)


web_poet/testing/utils.py
```diff
@@ -1,11 +1,9 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any
+from typing import Any
 
-from _pytest.assertion.util import assertrepr_compare
-
-if TYPE_CHECKING:
-    import pytest
+import _pytest.assertion.util as assertion_util
```
wRAR, line 5: Import style changed to be able to access `get_assertion_text_diff_style()` added in 9.1.0

PR #277, opened 2026-07-30
Provide a declarative selector API

AdrianAtZyte:
To do:
- [x] https://github.com/scrapinghub/web-poet/pull/284 (this is built on top, so its changes will show until merged and this PR gets master merged again)
- [x] Get frostwork published on PyPI, then update this PR, including the addition of an extra for it.
- [x] Self-review.


tests/test_selectors.py
```diff
@@ -0,0 +1,394 @@
...
+        name = field(jmespath("website.name"))
+        price = field(jmespath("price"))
+        tags = field(jmespath("tags", all=True))
+        missing = field(jmespath("missing"))
+
+    response = HttpResponse(
+        "http://example.com",
+        b'{"website": {"name": "homepage"}, "price": 10, "tags": ["a", "b"]}',
+        encoding="utf-8",
+    )
```
kmike, line 180: How would syntax look like if one wants to use jmespath for JSON extracted from a web page, or from API response coming from additional request, via HttpClient?

tests/test_selectors.py
```diff
@@ -0,0 +1,394 @@
...
+    @attrs.define
+    class Page(BrowserPage):
+        name = field(css("h1::text"), out=[str.strip])
+
+    response = BrowserResponse(url="http://example.com", html=HTML)
+    assert asyncio.run(Page(response=response).to_item()) == {"name": "Foo"}
+
+
+def test_lazy_extraction(response) -> None:
+    """Declarations are extracted one at a time, on demand."""
```
kmike: hm, is it a behavior we want? I'm not sure I understand this test case

tests/test_selectors.py
```diff
@@ -0,0 +1,394 @@
...
+
+    @attrs.define
+    class BrokenPage(WebPage):
+        name = field(css("h1::text"))
+        broken = field(css("::::"))
+
+    page = BrokenPage(response=response)
+    assert page.name == " Foo "
+    with pytest.raises(SelectorSyntaxError):
+        page.broken
```
kmike, line 431: I wonder if it'd be possible to fail immediately, on incorrect definition, not at runtime (it'd be failing on every call, right?)

tests/test_selectors.py
```diff
@@ -0,0 +1,394 @@
...
+
+
+def test_shared_declaration(response) -> None:
+    """The same declaration object can be used by different attributes and
+    classes."""
+    shared = css(".sku::text")
```
kmike: Is there a use case for this?

AdrianAtZyte: ~~For supporting non-field declarative selectors? I assume that’s something that someone will eventually want. Like e.g. extracting a string from a selector and then using regex in 2 fields so that each extracts something different from that string.~~

~~Or do you meant something else?~~

(an LLM pointed out this was not what you meant)

tests/test_selectors.py
AdrianAtZyte: I will add a test to make sure we cover the first scenario, but I assume we implement no declarative chaining, so you instead define a selector (like the shared example you commented on later) and in fields you use that selector with jmespath().

For the second scenario, I think #275 would give us this for free.

kmike:
hey @AdrianAtZyte! I think the API makes sense on its own, but I wonder if it'd be better to finalize its design together with frostwork optimization implementation. The reason: optimization use case may change the API. I don't see concrete issues right now though.

tests/test_selectors.py
kmike: but I'd probably want several selectors, not one, if there are optimizations. Jmespath is maybe not a right example; let's consider a use case where you fetch HTML as an additional request. Is #275 the only way?

AdrianAtZyte: It is the only *clean* way I can think of. What kind of high-level API are you hoping for?

tests/test_selectors.py
AdrianAtZyte: Renamed the test and reworded the docstring.

tests/test_selectors.py
AdrianAtZyte: Done.

tests/test_selectors.py
AdrianAtZyte: Not really. I split the test in two: `test_reused_declaration` is the one with an actual use case, reusing a declaration for a field with processors and for a field that builds something else out of the raw value.

tests/test_selectors.py
AdrianAtZyte: https://github.com/scrapinghub/web-poet/pull/277/commits/721c7dd79b4937aa398058e452c561ea476480cd? (docs about how to achieve this with the current code)

AdrianAtZyte:
> | Page | Imperative | Declarative | Speedup |
> | --- | --- | --- | --- |
> | product | 355.5 ms | 47.3 ms | 7.5× |
> | job | 92.3 ms | 9.1 ms | 10.1× |
> | article | 25.5 ms | 7.2 ms | 3.5× |
> | minimal | 1.4 ms | 0.8 ms | 1.7× |

AdrianAtZyte:
@kmike b128407 makes it so that, even with `frostwork` installed, you need to opt in. While more verbose, it does feel safer given frostwork is not a perfect parsel match.

